PR #2, opened 2026-08-18
Feat/core lifecycle and defect fixes

petarzarkov:
## What changed
Research

## Checks

- [x ] `bun run build`
- [x ] `bun run lint:check`
- [ x] `bun run format:check`
- [x ] `bun run typecheck`
- [ x] `bun run test:cov`
- [x ] Tests cover the change. A bug fix has a test that fails without the fix.
- [x ] Conventional commit messages (`feat:`, `fix:`, `docs:`, ...).
- [x ] No em or en dashes. No `enum`, no `any`, no `Dunx`-prefixed identifiers.
- [x ] Docs updated if the public surface changed (`docs/guide/`, package README).


packages/infra/src/schedule/decorators.ts
```diff
@@ -0,0 +1,116 @@
...
+ * record, and the runner finds it by walking the prototype chains of the classes
+ * the modules already declare, so a handler needs no second registration and an
+ * abstract base's marked methods are inherited by every subclass.
+ *
+ * The expression is validated here, at decoration time, by asking `Bun.cron.parse`
+ * for a next fire. A schedule that cannot parse is a boot error rather than a job
+ * that never runs.
+ */
+export const Cron = (
+  expression: string,
```
petarzarkov: We'd need to also accept enum like expression along the string

packages/infra/src/schedule/decorators.ts
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Runs the method once, `ms` after the app is ready.
+ *
+ * "Ready" is `onInit`, which is the latest hook there is and runs **before**
+ * `Bun.serve` binds. So the delay is measured from container readiness rather than
+ * from the first request, and `@Timeout(0)` fires before the socket is open. An app
+ * needing the later point uses `forRoot({ enabled: false })` and calls
+ * `registry.add` after `listen()`.
+ */
+export const Timeout = (ms: number, options: TimerDecoratorOptions = {}) =>
```
petarzarkov: We might need to rename this to OnceOnBoot and even then I'm not sure I see the use case, but let's see

packages/infra/src/schedule/bun-cron.ts
```diff
@@ -0,0 +1,39 @@
...
+  options?: CronCallOptions,
+) => CronHandle;
+
+type ParseWithOptions = (
+  expression: string,
+  relativeDate?: Date | number,
+  options?: CronCallOptions,
+) => Date | null;
+
+export const cron = Bun.cron as unknown as CronWithOptions;
```
petarzarkov, line 38: Do we need to do these castings?
We shadow the options here

PR #24, opened 2026-08-30
docs(queue): plan queues and cache on the database, and pin today's behaviour

petarzarkov:
Phase 0 of moving queues off Redis: the plan, and a characterization suite the implementation gets built against. **No queue code changes here.**

Plan: `internal/notes/roadmap/db-backed-queues-and-cache.md`.

## Two decisions it leaves open rather than assuming

**1. Rule 1 forbids writing a dunx job queue.** It also "outranks every other consideration" and says a conflict is "worth raising rather than resolving quietly", so it is raised. Measured in #21: bullmq covers Redis, pg-boss covers Postgres and runs on `Bun.SQL`, and **SQLite has nothing to integrate**. The SQLite requirement therefore cannot be met by integration. The plan bounds it three ways and recommends the middle one - bullmq for Redis, one dunx SQL engine for SQLite and Postgres - on the grounds that the SQLite half is forced, and once written, Postgres costs a dialect rather than a second job model a consumer would feel when moving between dev and production.

**2. bullmq's types are public surface and consumer code.** `JobPublisher.queue()` returns a bullmq `Queue`, `publish()` returns a bullmq `Job`, `QueueOptions.worker` is `Omit<BullWorkerOptions, ...>`, and `@JobHandler` methods receive `Job<T>` - `examples/full/src/jobs/thumbnail.jobs.ts` imports the type. Keeping "everything working with the new path" means a dunx-owned job contract, which is a breaking change to `@dunx/infra/queue`. The plan sketches the interface and flags that whether bullmq's `Job` satisfies it structurally is a field-by-field check nobody has done yet.

## And one rule it reverses

CLAUDE.md says "dunx renders no queue UI" and "**Do not re-add a queue panel.**" That was written because a hand-rolled panel was a worse bull-board. bull-board reads bullmq and nothing else, so for a SQL queue it is not a worse option, it is no option. The plan keeps bull-board mounted for Redis and adds a dunx panel for the SQL backends, and says the rule needs rewording rather than ignoring.

## The characterization suite

`examples/full/src/jobs.characterize.test.ts` asserts through HTTP routes and imports **no bullmq type**, so the same file runs unchanged against a SQL backend. It pins: in-process execution and the handler result, retry-then-succeed, the attempt limit with a failure reason, a delayed job held back then run, a batch draining, the queues the publisher opened, a 404 for an unknown id, and the 503 degradation with no broker.

To make retries and delays observable it adds a second, in-process queue to `examples/full` with `echo` and `flaky` handlers. It had to be a second queue: a queue is sandboxed when **any** handler on it asks to be, so the forked thumbnail queue could not carry them.

**Two things it pins were found by watching it flake, not by design:**

- A backend reports `completed` before the return value is readable. `jobs.test.ts` already documented this; my first helper stopped at the state and failed on the assertion that mattered.
- `failed` is observable *between* attempts, so a read that stops at the first `failed` can catch a job about to be retried. The failure wait now requires the attempt budget to be spent too.

Both are exactly the kind of thing a reimplementation would get subtly wrong, which is the argument for writing this suite before the engine rather than after.

## Gates

`bun run ci` plus `bun run ci unit`, green except the Chrome-less `browser` phase. The characterization suite run three times for flakiness: 8 pass each time. `sync:templates` run, since `examples/full/src/jobs` is vendored.

## Summary by CodeRabbit

* **New Features**
  * Added task queue capabilities to the jobs example and starter template, including queue discovery, task submission, delayed execution, retries, and status/results retrieval.
  * Added sample successful and intentionally retrying task handlers.

* **Bug Fixes**
  * Improved queue startup reliability by opening workers sequentially and waiting for readiness.
  * Reduced repeated error messages during broker outages.

* **Documentation**
  * Added roadmap documentation for database-backed queues and persistent caching across SQLite, Postgres, Redis, and memory-based approaches.

* **Tests**
  * Added HTTP-level coverage for job execution, retries, delays, batching, missing tasks, and unavailable queues.


petarzarkov:
Closing: writing a queue engine is a bigger commitment than this repo wants to make, and Rule 1 was pointing at that from the start. Wrapping an engine that already exists would be a different question; building one is not what dunx is for.

The characterization suite goes with it. What it found on the way is already merged in #25: `QueueConsumer.start()` opened every worker before awaiting any, so a second queue flooded 21.9M log lines against an absent broker and hung CI, and the failed-start path force-closed workers that were already consuming.

PR #34, opened 2026-09-02
feat: W3C Trace Context by default, requestId removed, and request/query stats

petarzarkov:
## Trace context replaces the request id

`requestId` and `traceId` did the same job and only one of them is a standard, so
`x-request-id` is gone: no `REQUEST_ID_HEADER`, no `RequestIds`, no
`propagateRequestId`, no `RequestFields.requestId`, and no shim. Trace context is
now on by default, toggled with `requestLogging: { trace: false }`, and the
response carries **`traceresponse`** (W3C Trace Context Level 2).

Two things beyond the rename:

- **Minting got cheaper.** `Uint8Array.prototype.toHex` produces a trace id and a
  span id in **49.2 ns**, against **260.5 ns** for `crypto.randomUUID()` plus a
  span. `toHex` exists on Bun 1.4.0 and typechecks under `lib: ESNext`.
- **A sampling bug went with it.** The outbound client hardcoded `flags: '01'`, so
  a trace an upstream sampler had *declined* was re-sampled at every dunx hop.
  `traceFlags` now travels in the async scope and is forwarded as it arrived.

## OpenTelemetry

Bun 1.4.0 exposes **no** native trace API (`Bun.*`, `bun:otel`, `bun:telemetry`,
`bun:trace` all absent), and the existing note in `docs/bun-apis.md` already
recorded that its auto-instrumentation never sees `Bun.serve`. So `TraceContext`
stays dunx's own.

The propagation half does work. `packages/http/src/server/otel-interop.test.ts`
asserts both directions against a real `@opentelemetry/sdk-trace-node` over a live
`Bun.serve`: an SDK-injected `traceparent` is continued by dunx, and an SDK
downstream picks dunx's span up as a remote parent. `@opentelemetry/*` are
devDependencies of `@dunx/http` only and reach no published manifest.

## Stats

Built to `internal/notes/research/stats.md`. **+35.2 ns per request**, folded into
the `.then` request logging already allocates.

| Package | Adds |
| --- | --- |
| `@dunx/core` | `Durations`, `Counter`, `Gauge`, `RuntimeStats`, `EventLoopLag` - on `node:perf_hooks` and `process`, so still zero dependencies |
| `@dunx/http` | `RequestMetrics`, `MetricsMiddleware`, `metrics: true` |
| `@dunx/infra/db` | `QueryMetrics`, `DbModule.forRoot(options, { metrics: true })` |
| `@dunx/dashboard` | A Stats panel and `{path}/api/stats`; `MemoryReport`/`RuntimeReport` move down into core, deleting the duplicate `memory()` |

### Three corrections to the research note

Recorded in `docs/architecture/metrics.md`.

1. **`percentiles` does not throw.** It is a `Map`, so `JSON.stringify` returns
   `{}` with no error. Silent loss rather than a loud failure.
2. **Keying misses on the route context does not work.** `unmatchedContext` builds
   a *fresh* context per request, so a `Map<RouteContext, Series>` neither
   collapsed misses nor stayed bounded: a scanner walking urls added an entry and
   a ~16 KiB histogram per probe, indefinitely. Misses key by method now, and a
   test walks 1,000 urls.
3. **Drizzle can time nothing.** `logQuery` fires immediately before the statement
   runs with no completion callback, and drizzle 0.45.2's OpenTelemetry hook is
   dead code (`let otel;`, never assigned). The timer wraps the driver dunx itself
   constructs. `Bun.SQL`'s `unsafe()` returns a lazy `Query` where attaching
   `.finally()` would *start* it, so `then` is wrapped; verified against Postgres 16.

## Verification

`bun run ci`: **12/12 gates, 1816 tests, 96.1% lines**, every package above the 90%
floor. Run with Postgres and Redis up so the live-service paths actually executed.

## Open question for review

Whether the stats collection layer should instead be built on `@opentelemetry/api`
as an optional peer, rather than on the Bun-native primitives. The note rejected it
on measurement (noop 14.4-18.0 ns, but 904-932 ns with an SDK registered, plus a
1.0 MB peer on a framework whose core has zero dependencies). Being re-examined
separately.

## Summary by CodeRabbit

* **New Features**
  * Added HTTP request, database query, runtime, and event-loop metrics with dashboard visualization.
  * Added W3C Trace Context propagation, including `traceparent`, `tracestate`, and `traceresponse` support.
  * Added metrics configuration and a Stats feature to generated applications.
  * Added counters, gauges, duration histograms, and runtime statistics APIs.
  * Added `server-timing` response measurements in the testing example.
* **Bug Fixes**
  * Improved OpenAPI schema reference resolution and preservation of reachable contributed schemas.
* **Documentation**
  * Expanded guides covering tracing, metrics, database instrumentation, and Bun API behavior.
* **Chores**
  * Replaced legacy request-ID correlation and Husky hooks with updated tracing and staged-file tooling.


petarzarkov:
@coderabbitai review

PR #49, opened 2026-09-07
fix(infra): remove the literal NUL from the job dedupe key, and guard control characters

petarzarkov:
## What changed

The `(queue, name)` dedupe key in `assertNoDuplicateJobs` held a literal NUL
byte. It is now the unicode escape, which denotes the same character, so the
separator and the key are unchanged. `scripts/no-control-chars.test.ts` guards
the C0 range across every tracked and untracked file.

The branch also carries `internal/notes/research/oop-conversion.md`, the
functional-to-OOP assessment that turned this up. Happy to split it out if you
would rather review the fix alone.

## Why

A NUL in the first 8000 bytes makes git, grep and ripgrep all classify the blob
as binary. Every commit that ever touched `discover.ts` rendered as
`Bin N -> M bytes`, including the one that created it, so the file's diffs have
never been readable in review, and repo-wide searches skipped it without
reporting anything.

NUL stays the separator: it is chosen for being impossible in a queue or job
name, so a printable replacement would risk collisions.

The guard compares byte codes rather than matching a character class. Written as
a regex it matched itself, since a NUL escape inside one is a normalisation away
from becoming the byte it stands for.

## Checks

- [x] `bun run build`
- [x] `bun run lint:check`
- [x] `bun run format:check`
- [x] `bun run typecheck`
- [x] `bun run test:cov`

`bun run ci` is green: 13 of 13 steps. The guard was run before the fix and
failed naming `packages/infra/src/queue/discover.ts:66`. Public surface
unchanged; `CONTRIBUTING.md` gains the rule beside its em dash sibling.

## Summary by CodeRabbit

- **Documentation**
  - Added contribution guidance for avoiding prohibited control characters in text files, including approved escape-sequence usage.
  - Added research documentation outlining architectural conversion options, trade-offs, and recommendations.

- **Tests**
  - Added automated validation to detect prohibited control characters in repository text files and report their locations.
  - Added coverage confirming similarly named jobs in different queues are handled as distinct jobs.

- **Chores**
  - Updated repository ignore rules and documentation to exclude temporary build artifacts from source control.


petarzarkov:
@coderabbitai review

petarzarkov:
Both findings were valid and are fixed in 8d5332c.

**`no-control-chars.test.ts`** now lists with `-z` and splits on NUL. I
reproduced the hole first: `git ls-files` quotes a pathname holding a newline,
so the split yielded `"a\nb.ts"`, which opens as nothing and was swallowed by
the read fallback, letting such a file escape the scan. Planting one now reports
it. The fallback is gone too, so anything that exists and fails to read fails
the guard instead of counting as clean.

**The note** no longer recommends a printable separator. It records that the
source spells the byte as a unicode escape while the runtime separator is still
NUL, which is what the `(queue, name)` key needs.

One thing found while verifying, unrelated to the review: `bun run ci` runs the
`examples` phase, which compiles `examples/binary`, at the same time as
`coverage`, and `bun build --compile` writes a transient
`.<hash>-<n>.bun-build` beside its output. It was unignored, so every guard
walking `git ls-files --others` could see it while it existed. It raced
`no-em-dash.test.ts` on a compiled bundle carrying an em dash, and the new guard
would have flaked on the same file. `*.bun-build` is now gitignored, next to the
existing `cli-fixture-*` entry that is there for the same reason.

petarzarkov:
Both fixed in 297406a. The first one found the previous commit's reasoning
wrong, not just its wording, so thank you for it.

**The collision is real and the key changed.** I had justified keeping NUL by
saying it cannot occur in a queue or job name. `JobMeta.queue` and
`JobMeta.name` are unvalidated `string`s, so it can, and `("a", "b<NUL>c")` and
`("a<NUL>b", "c")` key identically. The test added here fails against the
escaped key with a phantom `Two handlers claim job "c" on queue "a b"`, which is
a boot error rejecting two distinct handlers. Since no separator character fixes
it, the key is now `JSON.stringify([job.queue, job.name])`, injective for any
pair of strings. Validation was the other option offered; this one needs no new
error path and, as a side effect, leaves no control byte in the source at all,
escaped or otherwise.

**The read race is fixed as suggested**, skipping only ENOENT and propagating
everything else. Dropping the `exists()` check with it, since checking first
moved the window rather than closing it.

Verified: `static` and `coverage` phases both exit 0, infra at 92.1% funcs /
92.9% lines, 224 queue and scripts tests pass.

PR #50, opened 2026-09-07
test(examples): load-run oracle, negative-flow suites, and three fixes they found

petarzarkov:
The soak reported `0 failed` on a run that answered **59.4% of its requests with 429**: every op accepted one, so the rate limiter was the only thing under test. Ops now name work statuses and refusal statuses separately and the run floors the ratio, judges the server's own `RequestMetrics` (per-route p99, status distribution), raises `THROTTLE_LIMIT` so the app is what is loaded, and gives each worker its own `x-api-key` so the throttle subject is exercised. Same 40s window afterwards: 67.6% 200, 13.3% 201, 4.8% 429 all on `/limits/burst`.

Three phases it lacked: a stress burst at 4x concurrency asserting correctness rather than throughput, a recovery round, and a shutdown with traffic actually in flight (`ci.ts` claimed that one; the harness awaited every worker first).

Five suites, 78 to 154 tests in `examples/full` - auth (15), negative (34), throttle (7), gateway (8), dashboard (14).

Writing them found three things, fixed in the second commit:

- **`@dunx/dashboard` reported route paths that 404.** `routesOf` walks prototypes; the global prefix is applied at `listen()`. `RoutePrefix` now carries the resolved value, bound in the global wrapper next to `ClientAddress`. Gateways stay unprefixed and the new suite pins both halves.
- **better-auth's origin check is off under `NODE_ENV=test`**, which `bun test` sets. Measured: unset 403, test 200, development 403, production 403; a wrong `Origin` the same. The suite asserts the exemption in-process and the protection in a spawned `NODE_ENV=production`, covering `MISSING_OR_NULL_ORIGIN` and `INVALID_ORIGIN`.
- **`authorize` had no example.** `DASHBOARD_TOKEN` is unset by default, so `bun start` stays explorable and the boot warning stands; set it and a stranger gets 404, not 403.

Deliberately not done: consolidating `@dunx/openapi`'s prefix inference onto `RoutePrefix`. It also covers `mountAt` and has its own suite; `mount.ts`'s now-stale comment is corrected to say so.

The third commit takes CodeRabbit's six findings. The substantial two: the origin check is now asserted rather than only documented, and the aborted-client test could not fail (it waited past the handler's own completion) - Bun drops aborted requests inside 10ms, so it now requires the release under the 300ms sleep, verified by breaking it.

`bun run ci` green phase by phase (build, static, unit, examples, docs, coverage, browser). All 10 packages clear the 90% floor.

## Summary by CodeRabbit

* **New Features**
  * Dashboard routes and reported API paths now reflect configured global prefixes, while WebSocket gateway paths remain unprefixed.
  * WebSocket clients can connect to custom paths and send raw text or binary frames.
  * Optional dashboard token protection returns indistinguishable 404 responses for unauthorized requests.

* **Tests**
  * Added comprehensive authentication, dashboard, WebSocket, throttling, validation, shutdown, and load-testing coverage.

* **Documentation**
  * Expanded guidance for authentication origin checks, route prefixes, WebSockets, and load-test methodology.

### DIFF
--- a/docs/architecture/authentication.md
+++ b/docs/architecture/authentication.md
@@ -226,7 +226,7 @@ header (`MISSING_OR_NULL_ORIGIN`), so a server-side client has to send one match
 
 ### The origin check is off under `NODE_ENV=test`
 
-Which `bun test` sets, so no suite in this repo can assert it. Measured on
+Which `bun test` sets, so an in-process suite cannot observe it. Measured on
 `examples/full`, signing in with a session cookie and no `Origin`:
 
 | `NODE_ENV`  | status |
@@ -239,8 +239,15 @@ Which `bun test` sets, so no suite in this repo can assert it. Measured on
 A wrong `Origin` behaves the same way: 403 `INVALID_ORIGIN` with `NODE_ENV`
 unset, 200 under `bun test`. Both halves of the check are exempt, not one.
 
-`examples/full/src/auth.test.ts` asserts the 200 and says why. The test worth
-writing is the 403, and it fails: reading that failure as better-auth not
-checking the origin is the way this ends with someone deleting a security
-control. The tour never met it because `auth.demo.ts` sends a trusted `Origin`
-on every call.
+So the suite asserts both halves and needs two processes to do it. In-process
+it asserts the 200 and says why, because the test worth writing is the 403 and
+it fails there: reading that failure as better-auth not checking the origin is
+the way this ends with someone deleting a security control. The 403 is asserted
+against a spawned `NODE_ENV=production`, which is the same spawn the shutdown
+suite uses, and covers `MISSING_OR_NULL_ORIGIN` and `INVALID_ORIGIN` both.
+
+Sign-up and sign-in there carry no `Origin` at all. The check fires on a
+cookie-bearing request, and a present-but-untrusted origin is refused with or
+without one, so sending none until there is a cookie is what keeps the two
+assertions about the cookie-bearing case. The tour never met any of this because
+`auth.demo.ts` sends a trusted `Origin` on every call.
--- a/docs/architecture/authentication.md
+++ b/docs/architecture/authentication.md
@@ -223,3 +223,24 @@ drops what follows the first semicolon - four `CREATE TABLE`s in one template gi
 one table. And better-auth rejects a cookie-bearing state change with no `Origin`
 header (`MISSING_OR_NULL_ORIGIN`), so a server-side client has to send one matching
 `trustedOrigins`; a browser does it for free.
+
+### The origin check is off under `NODE_ENV=test`
+
+Which `bun test` sets, so no suite in this repo can assert it. Measured on
+`examples/full`, signing in with a session cookie and no `Origin`:
+
+| `NODE_ENV`  | status |
+| ----------- | ------ |
+| unset       | 403    |
+| test        | 200    |
+| development | 403    |
+| production  | 403    |
+
+A wrong `Origin` behaves the same way: 403 `INVALID_ORIGIN` with `NODE_ENV`
+unset, 200 under `bun test`. Both halves of the check are exempt, not one.
+
+`examples/full/src/auth.test.ts` asserts the 200 and says why. The test worth
+writing is the 403, and it fails: reading that failure as better-auth not
+checking the origin is the way this ends with someone deleting a security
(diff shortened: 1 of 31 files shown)


petarzarkov:
All six addressed in 8578b77. Verified each against the code first; all six were real.

**`auth.test.ts` — non-test-mode Origin enforcement.** Done, and it corrects the PR's own claim that no suite here could assert it. `bun test` sets `NODE_ENV=test` for *this* process, not for one it spawns, so the app is spawned with `NODE_ENV=production` the way `shutdown.test.ts` already does and the flow runs over HTTP from the suite. Asserts `MISSING_OR_NULL_ORIGIN` and `INVALID_ORIGIN`. Sign-up and sign-in deliberately send no `Origin`: a present-but-untrusted one is refused with or without a cookie, so sending none until there is a cookie keeps both assertions about the cookie-bearing case.

**`negative.test.ts` — abort test could not fail.** Correct. Measured what Bun does rather than assuming: aborted requests are dropped inside 10ms, against a 300ms handler. So the test waits for `inFlight` to reach 25, aborts, and requires release inside 150ms - under the sleep, so a broken abort leaves them parked. Confirmed it fails when the abort call is removed.

**`gateway.test.ts` — settling window.** Correct; a close that has not begun to decrement looks identical to one that finished. Four consecutive unchanged readings now, rather than removing the early return outright, so a genuinely settled suite still costs 200ms instead of a full second.

**`soak/client.ts` — open timeout.** Correct, and it is the leak the soak reports on. Closes before rejecting.

**`soak/soak.ts` — uncancellable settle timer.** Correct, and it sat twenty lines below the cancellable timer added for that exact reason.

**`tooling.md` — throttle limits.** Correct, the sentence read as though a budget of five were exhausted by three requests. Five is the module default, three is `/limits/burst`'s own `@Throttle`; it says both now.

`bun run ci` green phase by phase.

petarzarkov:
Both addressed in fd8819d. Both were real.

**`auth.test.ts` — a readiness wait that could not end.** Correct, and worse than stated: `while (!text.includes('ctrl-c to stop'))` cannot terminate once the app has exited, so any boot failure surfaced as the 60-second test timeout with nothing pointing at the cause. Racing `exited` turns that into a throw naming the exit code - measured at 271 ms against a deliberately invalid `AUTH_SECRET`.

The same loop was in `shutdown.test.ts`, which is where I copied it from, so both needed the identical fix and a third copy was next. Extracted `SpawnedApp` (spawn, drain, wait, stop); both suites use it and `shutdown.test.ts` is 20 lines shorter. The app's own boot error stays on inherited stderr so the runner prints it beside the throw, rather than the helper trying to reassemble it from a stdout that is empty in exactly that case.

**`soak.ts` — settlement skipped on the timeout path.** Correct. `traffic.stop()` ends a worker's loop but not the `OpClient.call` inside it, and each of those holds a 10-second deadline, so the timeout branch returned with fetches outstanding. Both paths go through the bounded race now, extracted to a static `#settle` rather than a second inline copy of the cancellable-timer dance - which is what let the first version have a `Bun.sleep` in it.

`bun run ci` green phase by phase; all 10 packages over the 90% floor.

petarzarkov:
Both addressed in 545271d. Both real, and both older than this branch - I inherited them when `shutdown.test.ts`'s spawn moved into the helper.

**`shutdown.test.ts` — the order check passed when the hook had not run.** Correct. `indexOf` returns -1 for an absent marker and -1 is less than any real index, so the assertion held whether `users draining` had printed at all. Both indexes are checked non-negative first. Verified by pointing the assertion at a marker the app never prints: it fails now, where the old shape passed.

**`spawn-app.ts` — `URL.pathname` is not a filesystem path.** Correct. Used `Bun.fileURLToPath` rather than `node:url`'s: Bun ships the decoder and CLAUDE.md's Rule 1 puts a `Bun.*` API ahead of a Node built-in. Confirmed it decodes (`file:///tmp/a%20b/c/` to `/tmp/a b/c/`).

One thing worth surfacing rather than silently widening: the same `new URL(..., import.meta.url).pathname` pattern is in four other places - `packages/http/src/ws/relay.fixture.ts:106`, `surface.test.ts` (x3), `packages/infra/src/index.test.ts` - and the relay fixture one also feeds a spawn, so it has the identical failure mode. All pre-existing and outside this PR's scope; happy to sweep them in a follow-up if you want it.

`bun run ci` green phase by phase; all 10 packages over the 90% floor.

PR #52, opened 2026-09-07
feat(examples): a live demo of examples/full, with a landing page and a deploy image

petarzarkov:
## What changed

`examples/full` gains what it needed to be demo.dunx.win rather than something
you run locally: an interactive landing page at `/`, a Dockerfile and compose
file, and a deploy script. Three config flags carry the deployment's shape, all
defaulting to today's behaviour so `bun start` and the tour are unchanged:
`DASHBOARD_COMMANDS` (bull-board read-only), `AUTH_GUEST_ONLY` (no sign-up, a
guest session per visitor) and `AUTH_SESSION_DAYS`.

`PUBLIC_URL` is a fix rather than a feature, and so is the `publicUrl` config
group it adds to `@dunx/create-app`.

## Why

Deploying it found two things local runs could not.

better-auth's `baseURL` was hardcoded to `http://localhost:${port}`, and it
matches a browser's `Origin` against that, so every sign-in from a public domain
answered `INVALID_ORIGIN`. `examples/full/src/auth` is a vendored feature, so the
fix reaches the scaffold: any generated app deployed behind a domain would have
hit the same wall.

The `anonymous` plugin reads `isAnonymous` off the drizzle schema, not the DDL.
Adding it to `auth.tables.ts` alone left `sign-in/anonymous` answering 500 naming
the missing field, so `database/auth.schema.ts` carries it too.

The image builds the packages in an explicit order because every inter-package
edge is a `peerDependency` and `--filter '*'` orders by `dependencies` alone,
which is what `scripts/build-all.ts` exists for. Its final WORKDIR is the example
because Bun reads `bunfig.toml` from the process cwd and does not walk up, and
that file carries the transform preload.

## Checks

- [x] `bun run build`
- [x] `bun run lint:check`
- [x] `bun run format:check`
- [x] `bun run typecheck`
- [x] `bun run test:cov`

`bun run ci` green at 13 of 13. Running on the Pi behind the tunnel; the guest
flow, the read-only board and the origin fix were each verified against the
deployed container.

Stacked behind #51: this branch and it both touch `main.ts` and
`service.test.ts`, so I will rebase once #51 lands.

## Summary by CodeRabbit

- **New Features**
  - Added a responsive live demo landing page with interactive examples for WebSockets, rate limiting, queues, authentication, validation, and more.
  - Added guest-only authentication with anonymous sessions, configurable session duration, and optional signup restrictions.
  - Added public application URL and dashboard command configuration.
  - Added Docker and Compose deployment support with health checks and readiness validation.
  - Added the dashboard at `/api/dashboard`.
  - Improved dashboard navigation with grouped sections and clearer labels.

- **Bug Fixes**
  - The demo homepage now serves correctly while unrelated missing routes retain JSON 404 responses.

- **Tests**
  - Added coverage for guest authentication and landing-page behavior.


petarzarkov:
Valid, and fixed in f1acf90. A rejected fetch left the handler before the
re-enable, so one network failure disabled the button until a reload. The queue
panel directly below it already wrapped its loop in try/finally, so this was an
inconsistency in the same file rather than a considered choice.

Also rebased onto main now that #51 has landed. Both it and this branch touch
`examples/full/src/main.ts` and `service.test.ts`; the rebase applied with no
conflicts and both changes survived, so the example now carries `strict: false`
and the landing middleware together. `bun run ci` green at 13 of 13.

petarzarkov:
Both valid, fixed in the commit above.

The reconnect race was real in two directions: `Send` stayed enabled while the
replacement socket was CONNECTING, and the predecessor's `close` arrived after
the new socket had opened and disabled `Send` on a live connection. Each handler
now ignores itself once a newer socket has replaced it.

`signedIn()` gating on `!res.ok` was worse on this deployment than in general:
`/api/profile` sits behind a 300-per-minute throttle, so a 429 was a realistic
way to be told "you are a guest" while the real answer was discarded. It gates
on 401 now.

While checking these I extracted the script block and linted it, which found a
third of the same kind: a rejected fetch in `readSession` was an unhandled
rejection that left the panel blank. Also fixed.

PR #56, opened 2026-09-07
fix(ui): drop the Role column on a phone rather than squeezing it

petarzarkov:
## What changed

`Role` in the Providers panel now hides below `sm`, alongside the columns #55
already demoted.

## Why

#55 measured clean and the panel did fit, but looking at the deployed page showed
`Role` squeezed to roughly 40px beside a long monospace token, rendering every
badge as `P...`. A column that fits and cannot be read is worse than one that is
absent, and the measurement could not see that.

## Measurements

Providers panel, headers with a non-zero width:

| Viewport | Columns |
| --- | --- |
| 393 | Token |
| 800 | Token, Role, Injects |
| 1440 | Token, Role, Bound as, Module, Exported, Injects |

No element wider than the viewport at any of the three.

## Checks

- [x] `bun run build`
- [x] `bun run lint:check`
- [x] `bun run format:check`
- [x] `bun run typecheck`
- [x] `bun run test:cov`

`bun run ci` green at 13 of 13.

## Summary by CodeRabbit

- **Style**
  - The provider “Role” column is now hidden on smaller screen sizes to improve dashboard readability.


petarzarkov:
Correcting my own PR body: the first `bun run ci` on this branch failed, on
`source comments > has no comment block long enough to be a design memo`. It
passes standalone and the second full run was 13 of 13, and the commit touches
only `Graph.tsx` plus the regenerated bundle, so it was not this change.

It is the third time a guard that walks `git ls-files --cached --others` has
tripped over a file another phase was writing at that moment. The first was
`no-em-dash.test.ts` against a `.bun-build` scratch file, fixed by gitignoring
it. `--others` is deliberate, so that a brand-new unstaged file still counts, and
that is also what exposes these guards to whatever `build`, `examples` and
`docs` are writing while `coverage` scans. Worth a separate look rather than
another gitignore entry each time.

PR #59, opened 2026-09-07
refactor(http/client): simplify the retry jitter and Retry-After parse

petarzarkov:
Three small simplifications to `packages/http/src/client/retry.ts`: the jitter source, the backoff cap, and the `Retry-After` numeric check.

Branch is `test/dunxonu-review` - a throwaway for exercising the review workflow, not for merge.

## Summary by CodeRabbit

- **Bug Fixes**
  - Retry delays now apply jitter after the backoff limit, allowing a small randomized adjustment beyond the capped delay.
  - Numeric `Retry-After` values are now accepted more broadly, including non-finite numeric values.
  - Retry jitter behavior is more consistent across requests through standardized randomization.

### DIFF
--- a/packages/http/src/client/retry.ts
+++ b/packages/http/src/client/retry.ts
@@ -13,10 +13,7 @@ import { FetchError, FetchTransportError } from './errors.js';
  * is a CSPRNG, and costs nothing.
  */
 const uniform = (): number => {
-  const buffer = new Uint32Array(1);
-  crypto.getRandomValues(buffer);
-  // 2**32 rather than 0xffffffff, so the result is [0, 1) and never exactly 1.
-  return (buffer[0] ?? 0) / 2 ** 32;
+  return Math.random();
 };
 
 export interface BackoffOptions {
@@ -34,7 +31,7 @@ export interface BackoffOptions {
 export const backoffDelay = (
   attempt: number,
   { baseMs, power = 2, jitterMs = 1000, maxMs = 30_000 }: BackoffOptions,
-): number => Math.min(baseMs * power ** attempt + uniform() * jitterMs, maxMs);
+): number => Math.min(baseMs * power ** attempt, maxMs) + uniform() * jitterMs;
 
 /**
  * The wait an upstream asked for, in ms, or undefined.
@@ -52,7 +49,7 @@ export const retryAfterMs = (
   if (header === null) return undefined;
 
   const seconds = Number(header);
-  if (Number.isFinite(seconds)) return Math.max(0, seconds * 1000);
+  if (!Number.isNaN(seconds)) return Math.max(0, seconds * 1000);
 
   const at = Date.parse(header);
   return Number.isNaN(at) ? undefined : Math.max(0, at - now);
--- a/.github/workflows/dunxonu-review.yml
+++ b/.github/workflows/dunxonu-review.yml
@@ -72,8 +72,11 @@ jobs:
             ineligible: count it as already reviewed only if a review covers
             that exact commit. Review the changes since your own last review,
             and do not repeat a finding you have already posted.
-          # The action starts the inline-comment MCP server only when
-          # --allowedTools names it, even though the skill's own frontmatter
-          # already grants the tool.
+          # --allowedTools REPLACES the skill's own allowed-tools frontmatter
+          # rather than adding to it, so every tool the skill declares has to be
+          # named here or the call is denied in a non-interactive run. Naming
+          # only the MCP tool cost a whole review: 13 turns, one denial, nothing
+          # posted. The first eight are the skill's frontmatter verbatim; the
+          # git commands are what its blame-and-history reviewer needs.
           claude_args: |
-            --allowedTools "mcp__github_inline_comment__create_inline_comment"
+            --allowedTools "mcp__github_inline_comment__create_inline_comment,Bash(gh pr view:*),Bash(gh pr diff:*),Bash(gh pr list:*),Bash(gh pr comment:*),Bash(gh issue view:*),Bash(gh issue list:*),Bash(gh search:*),Bash(git log:*),Bash(git blame:*),Bash(git show:*),Bash(git diff:*)"
--- a/.github/workflows/dunxonu-review.yml
+++ b/.github/workflows/dunxonu-review.yml
@@ -73,6 +73,11 @@ jobs:
             ineligible: count it as already reviewed only if a review covers
             that exact commit. Review the changes since your own last review,
             and do not repeat a finding you have already posted.
+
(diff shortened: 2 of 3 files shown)


packages/http/src/client/retry.ts
```diff
@@ -52,7 +49,7 @@ export const retryAfterMs = (
   if (header === null) return undefined;
 
   const seconds = Number(header);
-  if (Number.isFinite(seconds)) return Math.max(0, seconds * 1000);
+  if (!Number.isNaN(seconds)) return Math.max(0, seconds * 1000);
```
dunxonu, line 53: `!Number.isNaN(seconds)` admits `Infinity`/`-Infinity`, which `Number.isFinite` correctly excluded. A `Retry-After: Infinity` header (or an overflowing numeric like `1e999`) now returns `Infinity` here instead of falling through to `Date.parse` and returning `undefined` as before - it happens to get clamped to `maxMs` by the caller in `decide()`, but that's incidental, not something this function guarantees. Worse, `Retry-After: -Infinity` now returns `Math.max(0, -Infinity) === 0`, i.e. an immediate no-backoff retry, where previously it fell through to the computed exponential backoff. Neither case is exercised by the existing tests.

```suggestion
  if (Number.isFinite(seconds)) return Math.max(0, seconds * 1000);
```

packages/http/src/client/retry.ts
```diff
@@ -13,10 +13,7 @@ import { FetchError, FetchTransportError } from './errors.js';
  * is a CSPRNG, and costs nothing.
  */
 const uniform = (): number => {
-  const buffer = new Uint32Array(1);
-  crypto.getRandomValues(buffer);
-  // 2**32 rather than 0xffffffff, so the result is [0, 1) and never exactly 1.
-  return (buffer[0] ?? 0) / 2 ** 32;
+  return Math.random();
 };
```
dunxonu, line 17: `uniform()` now returns `Math.random()`, which [CLAUDE.md](https://github.com/petarzarkov/dunx/blob/59a6ad2adbed9858152b8df5ff1c7b85259d4b86/CLAUDE.md#L94) explicitly bans for this exact use case:

> Random numbers, ids, sampling → Use `@arkv/rng` → Never: `Math.random` for anything that matters, `nanoid`, `uuid`

The unchanged doc comment right above this function (lines 4-14) still argues the opposite of what the code now does: it says "jitter matters, because decorrelating retries is the whole reason it exists" and that `crypto.getRandomValues` "is a CSPRNG, and costs nothing" - a claim that's now false for this implementation.

```suggestion
const uniform = (): number => {
  const buffer = new Uint32Array(1);
  crypto.getRandomValues(buffer);
  // 2**32 rather than 0xffffffff, so the result is [0, 1) and never exactly 1.
  return (buffer[0] ?? 0) / 2 ** 32;
};
```

packages/http/src/client/retry.ts
```diff
@@ -34,7 +31,7 @@ export interface BackoffOptions {
 export const backoffDelay = (
   attempt: number,
   { baseMs, power = 2, jitterMs = 1000, maxMs = 30_000 }: BackoffOptions,
-): number => Math.min(baseMs * power ** attempt + uniform() * jitterMs, maxMs);
+): number => Math.min(baseMs * power ** attempt, maxMs) + uniform() * jitterMs;
```
dunxonu, line 34: Moving the cap inside `Math.min` means it now only bounds the exponential term, and jitter is added afterward - so the returned delay can exceed `maxMs` by up to `jitterMs` (e.g. with the documented defaults, up to 31,000ms instead of the documented ceiling of 30,000ms). This contradicts the function's own doc comment two lines above, "`base * power^attempt + jitter`, capped", which says the whole sum is bounded. Nothing downstream re-clamps this value before it reaches `Bun.sleep`.

```suggestion
export const backoffDelay = (
  attempt: number,
  { baseMs, power = 2, jitterMs = 1000, maxMs = 30_000 }: BackoffOptions,
): number => Math.min(baseMs * power ** attempt + uniform() * jitterMs, maxMs);
```

packages/http/src/client/retry.ts
petarzarkov: The doc comment right above rejects @arkv/rng because it is a 64 KB WebAssembly PRNG in the most-imported package, just to pick a number of milliseconds. Does that same weight argument not also justify Math.random here? Jitter only needs to decorrelate, not resist an attacker.

dunxonu: **Claude finished @petarzarkov's task in 1m 17s** —— [View job](https://github.com/petarzarkov/dunx/actions/runs/34109664918)

---
### Reply posted

- [x] Read the review thread on `packages/http/src/client/retry.ts:17`
- [x] Answered in that thread

@petarzarkov asked whether the doc comment's WASM-weight argument against `@arkv/rng` equally justifies `Math.random`. Short answer: no - that argument compares two already-acceptable CSPRNG sources, while `Math.random` is banned by [CLAUDE.md](https://github.com/petarzarkov/dunx/blob/59a6ad2adbed9858152b8df5ff1c7b85259d4b86/CLAUDE.md#L94) by name regardless of size, for quality reasons unrelated to bundle weight. Full reasoning is in the reply: https://github.com/petarzarkov/dunx/pull/59#discussion_r3948664875

dunxonu: No, it doesn't transfer. The weight argument in that doc comment picks between two already-acceptable CSPRNG sources (`crypto.getRandomValues` vs `@arkv/rng`), both of which satisfy [CLAUDE.md's rule](https://github.com/petarzarkov/dunx/blob/59a6ad2adbed9858152b8df5ff1c7b85259d4b86/CLAUDE.md#L94) - it isn't weighing against `Math.random`, which that rule bans by name regardless of size. The concern there is quality, not bytes: jitter exists to decorrelate retrying clients, and `Math.random` gives weaker guarantees for that than a CSPRNG does.

`crypto.getRandomValues` isn't a cheaper compromise standing in for `@arkv/rng` - it's a Web standard Bun implements natively (Rule 1's first preference), and it costs less than `@arkv/rng`, not just less than a stronger guarantee than this call needs. So the two rejections aren't the same shape: `@arkv/rng` is passed over for a lighter option that still clears the bar CLAUDE.md sets, while `Math.random` is banned outright regardless of weight. The doc comment above the function is still correct; only the code beneath it regressed.

petarzarkov:
Closing: throwaway fixture for testing the review workflow. The three changes to retry.ts were deliberate defects, chosen because lint, typecheck and all 23 existing tests pass with them in place. dunxonu found all three. The workflow fixes this surfaced are in #61.

PR #60, opened 2026-09-07
fix(ui): stop calling a defaulted parameter a boot error

petarzarkov:
## What changed

The Overview alert counts only unresolved parameters that would actually fail
boot. A defaulted one is excluded. `Dependencies` still shows it, in yellow
rather than red, saying the default stands.

## Why

The demo's dashboard showed a red alert reading `BuildInfo.readonly retries = 3
named nothing at runtime ... Each is a boot error naming that parameter`, on a
running app. `examples/full` carries that parameter on purpose: `number` erases,
so it is recorded unresolved, and a default is the language saying the parameter
may be absent.

The data was already correct and both panels ignored it. `@dunx/transform`
records `optional: true`, `@dunx/core` carries it on `UnresolvedDep`, and
`graph.ts` puts it in the snapshot. The live demo's payload is
`{"unresolved":"readonly retries = 3","optional":true}`. The filter was
`'unresolved' in dependency` and nothing else.

That mattered more than a wrong label: the alert is the first thing on a page
whose job is to say whether the container can close, and it was crying wolf on
the only entry the demo has.

## Verified

Against the deployed snapshot: counted 1 before, 0 after. Two of the three new
tests fail against the old behaviour.

## Checks

- [x] `bun run build`
- [x] `bun run lint:check`
- [x] `bun run format:check`
- [x] `bun run typecheck`
- [x] `bun run test:cov`

`bun run ci` green at 13 of 13.

## Not fixed here

The recorded name is the parameter's source text, so it reads
`BuildInfo.readonly retries = 3` rather than `BuildInfo.retries`. Cosmetic, and
it comes from `@dunx/transform` rather than either panel.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dependency diagnostics now distinguish between required unresolved dependencies and dependencies with usable defaults.
  * Dependencies with defaults are shown with a light yellow indicator and explain that the default value is being used.
  * Optional dependencies no longer appear as boot errors.
  * Required unresolved dependencies retain prominent error indicators and relevant remediation guidance.


internal/dashboard-ui/src/panels/Dependencies.tsx
```diff
@@ -33,12 +34,17 @@ export const Dependencies = ({
...
             title={
-              dependency.typeOnly === undefined
-                ? 'This parameter names nothing at runtime, so it is a boot error'
-                : `Imported as \`import type { ${dependency.typeOnly} }\` - drop the \`type\``
+              dependency.optional !== undefined
+                ? 'Names nothing at runtime, but the parameter has a default, so the default stands'
+                : dependency.typeOnly === undefined
+                  ? 'This parameter names nothing at runtime, so it is a boot error'
+                  : `Imported as \`import type { ${dependency.typeOnly} }\` - drop the \`type\``
             }
```
dunxonu, line 48: **Bug**: this ternary now checks `optional` before `typeOnly`, so a parameter that has *both* (a default value **and** a type erased by `import type`) permanently loses the actionable "drop the `type`" hint and falls into the generic "has a default, so the default stands" message.

That combination is real: `packages/transform/src/deps.ts` emits `optional: true` and `typeOnly` independently on the same record (`hasDefault(param)` vs. `cause === 'import-type'`), and `Dependency` in `packages/core/src/di/graph.ts` allows both fields on one variant - so `constructor(private readonly x: SomeTypeOnlyImport = fallback) {}` produces exactly this. This file's own doc comment (lines 14-15) says `typeOnly` "is carried through because that case has a one-line fix" - this change makes that fix unreachable whenever a default is also present.

https://github.com/petarzarkov/dunx/blob/600f052c4f468e9fb91ccc1150f62e1061a3736a/internal/dashboard-ui/src/panels/Dependencies.tsx#L42-L48

Suggested fix: check `typeOnly` first so the specific hint always wins when present.

```suggestion
           
              dependency.typeOnly !== undefined
                ? `Imported as \`import type { ${dependency.typeOnly} }\` - drop the \`type\``
                : dependency.optional !== undefined
                  ? 'Names nothing at runtime, but the parameter has a default, so the default stands'
                  : 'This parameter names nothing at runtime, so it is a boot error'
            }
```

PR #61, opened 2026-09-07
fix(ci): make the dunxonu review actually post

petarzarkov:
Three findings from testing the review workflow on #59, all of which made it silently do nothing:

- `--allowedTools` replaces the skill's `allowed-tools` frontmatter rather than adding to it, so every tool it declares has to be named or the call is denied.
- The action runs its own nested `setup-bun`, which prepends to PATH after ours, so the repo toolchain step could not do what its comment claimed. Removed.
- The skill's subagents launch in the background, and a headless run ends before their completion notification arrives. No CLI flag exists, so the prompt says it.

Verified on #59: all three planted defects found, posted inline as dunxonu.

## Summary by CodeRabbit

- **Chores**
  - Simplified automated review workflow handling for clearer, more consistent review results.
  - Improved inline review feedback and status reporting.
  - Standardized workflow execution around a pinned runtime environment.
  - Added clearer status labels for validation progress.
  - Streamlined review responses and progress updates for addressed items.
  - Added default code ownership configuration to support automated review routing.
  - Improved release workflow authorization for more reliable versioning and publishing.

### DIFF
--- a/.github/workflows/dunxonu-review.yml
+++ b/.github/workflows/dunxonu-review.yml
@@ -72,8 +72,11 @@ jobs:
             ineligible: count it as already reviewed only if a review covers
             that exact commit. Review the changes since your own last review,
             and do not repeat a finding you have already posted.
-          # The action starts the inline-comment MCP server only when
-          # --allowedTools names it, even though the skill's own frontmatter
-          # already grants the tool.
+          # --allowedTools REPLACES the skill's own allowed-tools frontmatter
+          # rather than adding to it, so every tool the skill declares has to be
+          # named here or the call is denied in a non-interactive run. Naming
+          # only the MCP tool cost a whole review: 13 turns, one denial, nothing
+          # posted. The first eight are the skill's frontmatter verbatim; the
+          # git commands are what its blame-and-history reviewer needs.
           claude_args: |
-            --allowedTools "mcp__github_inline_comment__create_inline_comment"
+            --allowedTools "mcp__github_inline_comment__create_inline_comment,Bash(gh pr view:*),Bash(gh pr diff:*),Bash(gh pr list:*),Bash(gh pr comment:*),Bash(gh issue view:*),Bash(gh issue list:*),Bash(gh search:*),Bash(git log:*),Bash(git blame:*),Bash(git show:*),Bash(git diff:*)"
--- a/.github/workflows/dunxonu-review.yml
+++ b/.github/workflows/dunxonu-review.yml
@@ -34,9 +34,9 @@ jobs:
   review:
     name: Review the diff
     runs-on: ubuntu-latest
-    # The skill skips a draft itself; filtering here saves starting a runner for
-    # one, and `ready_for_review` is what picks it up later. The sender check is
-    # the loop guard: dunxonu must never review a push of its own.
+    # A draft is not worth a runner, and `ready_for_review` is what picks it
+    # up later. The sender check is the loop guard: dunxonu must never review a
+    # push of its own.
     if: >-
       github.event.pull_request.draft == false &&
       github.event.sender.login != 'dunxonu'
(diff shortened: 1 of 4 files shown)


dunxonu:
## Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

.github/workflows/dunxonu-review.yml
```diff
@@ -48,32 +48,32 @@ jobs:
...
         with:
           fetch-depth: 1
 
       - uses: anthropics/claude-code-action@v1
+        env:
+          # Belt and braces for the failure the marketplace plugin hit: a
+          # background task in a headless run never reports back, because the
+          # run ends when the main agent stops. `-p` already runs this review
+          # in the foreground; this turns the rest of it off.
+          CLAUDE_CODE_DISABLE_BACKGROUND_TASKS: '1'
```
dunxonu, line 64: `CLAUDE_CODE_DISABLE_BACKGROUND_TASKS` is not a documented/recognized environment variable for the Claude Code CLI or `claude-code-action`. The PR description itself says the intended fix was prompt-level ("no CLI flag exists, so the prompt says it"), but this diff instead sets an undocumented env var with no matching prompt change. If the var isn't honored, the original bug (background subagent findings from `/code-review` never reporting back before the headless CI run ends) is still unfixed, silently.

Worth confirming against the current `claude-code-action`/CLI docs before relying on this, and considering the documented `CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS` (set to a high value or `0`) as an alternative that is known to exist.

.github/workflows/dunxonu.yml
```diff
@@ -82,14 +85,15 @@ jobs:
           trigger_phrase: '@dunxonu'
 
       # A bare reply carries no trigger phrase, so interactive mode would ignore
-      # it. A prompt selects automation mode instead, and `track_progress` keeps
-      # the GitHub context and the progress comment a mention would have got.
+      # it. A prompt selects automation mode instead. No `track_progress`: it
+      # posts a tracking comment that restates the answer, so a single reply
+      # became two comments in the thread. Everything it would have injected as
+      # context is in the prompt below, and gh is allowed for the rest.
```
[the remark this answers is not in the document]
petarzarkov: Good catch, fixed. The body now goes to a JSON file and is posted with `gh api --input`, so an apostrophe in the reply text cannot break the command. `Write` was added to that step's allowedTools to make it possible.

.github/workflows/dunxonu-review.yml
petarzarkov: This finding is wrong on its main point, so I am keeping the variable. `CLAUDE_CODE_DISABLE_BACKGROUND_TASKS` does exist: `strings` over the 2.1.241 CLI binary lists it, and Anthropic's Code Review page documents setting it to `1` to turn off background tasks. You were right about one narrow thing, that it is absent from the env-vars reference table, which is a docs inconsistency rather than a missing feature.

Your suggested alternative `CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS` also exists in that binary, so that half was sound.

The variable is belt and braces regardless: the built-in `/code-review` already runs in the foreground under `-p`, which is the actual fix for the background-subagent problem.

.github/workflows/dunxonu-review.yml
```diff
@@ -48,32 +48,32 @@ jobs:
...
+          # is the dial the plugin did not have. Raise it to high for broader
+          # coverage and a longer run.
+          prompt: /code-review medium --comment ${{ github.repository }}/pull/${{ github.event.pull_request.number }}
+          # --allowedTools REPLACES the skill's own allowed-tools rather than
+          # adding to it, so anything it reaches for has to be named. A review
+          # of a workflow diff was denied 18 calls and still reported no issues,
+          # so gh api is named here too.
           claude_args: |
-            --allowedTools "mcp__github_inline_comment__create_inline_comment"
+            --allowedTools "mcp__github_inline_comment__create_inline_comment,Bash(gh pr view:*),Bash(gh pr diff:*),Bash(gh pr list:*),Bash(gh pr comment:*),Bash(gh issue view:*),Bash(gh api:*),Bash(git log:*),Bash(git blame:*),Bash(git show:*),Bash(git diff:*)"
```
dunxonu, line 79: `--allowedTools` replaces rather than adds to the skill's own `allowed-tools` frontmatter (per this PR's own reasoning two lines up), but this list only adds `gh`/`git` Bash patterns plus the inline-comment MCP tool. It omits `Read`, `Grep`, `Glob`, and `Agent` - all of which the `/code-review` skill's methodology depends on (Angle A: "Read the enclosing function for each hunk"; Angle C: "Grep for the symbol"; Phase 1: "via the Agent tool"). With those tools unnamed, a headless run denies every call to them, reproducing the exact "denied calls, reported no issues" failure this PR sets out to fix - just for a different set of tools.

Suggested fix: add `Read,Grep,Glob,Agent` to the allow-list.

.github/workflows/dunxonu.yml
```diff
@@ -67,10 +68,12 @@ jobs:
...
-      # The toolchain, so an answer that has to run `bun run ci` can. The review
-      # workflow needs none of this; a fix does.
-      - if: steps.addressed.outputs.how != 'no'
-        uses: ./.github/actions/setup
+      # No `./.github/actions/setup` here. The action runs its own nested
+      # oven-sh/setup-bun pinned to a different version, and setup-bun prepends
+      # to PATH, so the action's Bun shadows whatever we installed first. A gate
+      # run here would have used that Bun against a lockfile this repo pins to
+      # another one. CI runs the gates on whatever dunxonu pushes, under the
+      # pinned version, which is where that signal belongs.
```
dunxonu, line 76: This drops `./.github/actions/setup` entirely from the `respond` job, but that composite action does two things: pins Bun (which the action's own nested `setup-bun` now shadows, the stated reason for removing it) *and* runs `bun install --frozen-lockfile` (which nothing replaces). The action's nested Bun setup only installs a Bun binary, it never populates `node_modules`.

The old comment this replaces said "The toolchain, so an answer that has to run `bun run ci` can" - that need is real for the `mention` path (interactive mode answering "fix this"), even if it's not for `dunxonu-review.yml` (which never builds). With no `bun install` anywhere in this job, a mention asking dunxonu to run tests, typecheck, or `bun run ci` against a fix will fail on a missing `node_modules`, regardless of which Bun binary ends up on PATH.

If the intent is genuinely "no gate runs here, CI validates after push" per the new comment, that's a bigger behavior change than "remove a PATH conflict" and deserves stating outright rather than reusing the review workflow's reasoning verbatim.

.github/workflows/dunxonu-review.yml
```diff
@@ -48,32 +48,32 @@ jobs:
...
-            and do not repeat a finding you have already posted.
-          # The action starts the inline-comment MCP server only when
-          # --allowedTools names it, even though the skill's own frontmatter
-          # already grants the tool.
+          # The built-in review, not code-review@claude-code-plugins. It needs
+          # no marketplace clone, runs in the foreground under `-p`, and takes
+          # an effort level: medium reports only what it is confident in, which
+          # is the dial the plugin did not have. Raise it to high for broader
+          # coverage and a longer run.
+          prompt: /code-review medium --comment ${{ github.repository }}/pull/${{ github.event.pull_request.number }}
```
dunxonu, line 73: The prompt this replaced explicitly told the reviewer to treat a PR as already-reviewed only for the exact commit already covered, and not to repeat a finding already posted. This workflow still triggers on `synchronize` (line 23-24) and each run reviews the whole PR diff (`git diff main...HEAD` / `gh pr diff`, not just the new commits since the last review), with nothing in the new one-line prompt replacing that de-dup guidance. Every push after the first is likely to re-post inline comments for issues already found and commented on in an earlier run of the same PR.

PR #63, opened 2026-09-08
feat(example): five landing panels, a social card, and the flaky-key fix

petarzarkov:
## What changed

The demo page exposed 7 of the example's 23 capabilities and carried no
`og:image`. It now has five more panels backed by three `/api/demo` routes, its
own social card, and a link back to the docs.

## Why

Two things were blocking the demo from doing any promotional work:

- **A paste of `demo.dunx.win` rendered as a bare link.** No `og:image`, no
  `og:title`, while `dunx.win` unfurled with a card.
- **Most of the example was unreachable from a browser.** The ledger's 8 routes,
  the guards, the outbound client and the DI graph were all live and had no
  panel. The docs linked to the demo in eight places; the demo linked back in
  none.

Two defects fell out of building it, each with a test that fails without the fix:

- `FlakyController` claimed a per-key failure counter in its doc comment and read
  a hardcoded `'default'`, so only the first caller after boot ever saw a retry.
- `isConnectionError` only matches `@dunx/infra`'s own `RedisError`, not the one
  Bun's client raises through bullmq. The new queue tour step exited 1 against an
  absent broker until it caught broadly, the way `JobsController.degrades` does.

The queue was also the only package with no tour step, which is why the second
one went unnoticed.

## Checks

- [x] `bun run build`
- [x] `bun run lint:check`
- [x] `bun run format:check`
- [x] `bun run typecheck`
- [x] `bun run test:cov`

`bun run ci` is 13/13 green with valkey, postgres and MinIO up.

Not done: a web-visible tour transcript. It would be stale or machine-specific
committed, since the tour's output depends on which services were up, and the
deploy job would need Redis and Postgres to regenerate it.

## Summary by CodeRabbit

- **New Features**
  - Expanded the interactive landing page with demos for transactions, retries, process vitals, dependency injection, guards, queues, sessions, validation, tracing, and WebSockets.
  - Added browser-driven queue and retry demonstrations with status and result reporting.
  - Added social preview metadata and a generated Open Graph image.

- **Documentation**
  - Updated project and example documentation with landing-page routes, demo commands, queue completion guidance, and soak-test results.

- **Bug Fixes**
  - Isolated flaky upstream demo behavior by request key with bounded tracking.
  - Improved recognition of Redis connection and server errors.


examples/full/src/upstream/flaky.controller.ts
```diff
@@ -18,10 +20,16 @@ import {
 export class FlakyController {
   readonly #failures = new Map<string, number>();
```
dunxonu, line 26: The per-key fix removed the bound that the old hardcoded `'default'` key provided: `#failures` is now a `Map` keyed directly by the client-supplied `?key=` (line 32) with no eviction, ever. `/api/upstream/flaky` is `@SkipThrottle()` and reachable directly by any visitor to the public, long-running demo (`demo.dunx.win`) - repeatedly hitting it with distinct `key` values grows this map without bound for the process lifetime. Worth capping the map size or evicting entries once a key has recovered (`seen > 2`).

examples/full/src/jobs/jobs.demo.ts
```diff
@@ -0,0 +1,79 @@
...
+    try {
+      await this.publishAndWait();
+    } catch (error) {
+      // Broadly, the way `JobsController.degrades` does: bullmq surfaces a
+      // broker failure through its own client, so what arrives is Bun's
+      // `RedisError` rather than the one `isConnectionError` knows.
+      const reason = isConnectionError(error)
+        ? 'nothing answering'
+        : (error as Error).message;
+      this.logger.info(`no broker reachable - skipping the queue: ${reason}`);
```
dunxonu, line 26: This catch is unconditional: any exception from `publishAndWait()` - not just a broker-connection failure - gets logged as `no broker reachable - skipping the queue: ${reason}`. `demonstrate()` is called with `await this.jobs.demonstrate()` from `Tour` (`tour.service.ts:118`) and never rethrows, so a real bug here (e.g. a `TypeError` from `queue.getJob`, or a regression in `ThumbnailJobs`/`thumbnail.jobs.ts`) is silently mislabeled as "no broker" and swallowed, with no test asserting the log message to catch the mislabeling. Consider checking `isConnectionError(error)` before deciding to skip, and rethrowing (or logging distinctly) otherwise.

examples/full/src/landing/landing.middleware.ts
```diff
@@ -6,12 +6,32 @@ import {
...
+  '/landing.js': 'landing.js',
+  '/og.png': 'og.png',
+};
+
+const TYPES: Readonly<Record<string, string>> = {
+  html: 'text/html; charset=utf-8',
+  css: 'text/css; charset=utf-8',
+  js: 'text/javascript; charset=utf-8',
+  png: 'image/png',
+};
```
dunxonu: This hand-rolled `TYPES` extension-to-content-type map reimplements what `Bun.file` already infers automatically from the extension when passed as a `Response` body. `packages/http/src/static/files.ts:98` (in the same repo, same package family) states this explicitly: `// Bun sets content-type from the extension` - and only overrides when `file.type === ''`. CLAUDE.md Rule 1 bans reimplementing a Bun primitive ("Never reimplement what Bun already does"). Dropping the explicit `content-type` header (or only setting it as a fallback like `StaticFiles` does) would remove this second, hand-maintained mapping.

examples/full/src/jobs/jobs.demo.ts
```diff
@@ -0,0 +1,79 @@
...
+  async demonstrate(): Promise<void> {
+    try {
+      await this.publishAndWait();
+    } catch (error) {
+      // Broadly, the way `JobsController.degrades` does: bullmq surfaces a
+      // broker failure through its own client, so what arrives is Bun's
+      // `RedisError` rather than the one `isConnectionError` knows.
+      const reason = isConnectionError(error)
+        ? 'nothing answering'
+        : (error as Error).message;
```
dunxonu: `(error as Error).message` assumes the caught value is an `Error` instance. If `publishAndWait()` throws a non-`Error` (a string, or a plain object bullmq/Bun sometimes rejects with), `reason` becomes `undefined` and the log line reads `no broker reachable - skipping the queue: undefined`, discarding whatever diagnostic information was actually thrown.

examples/full/src/landing/vitals.controller.ts
```diff
@@ -0,0 +1,180 @@
...
+ * would be a traversal with a nice name. */
+const SOURCES: Readonly<Record<string, string>> = {
+  ledger: new URL('../database/ledger.service.ts', import.meta.url).pathname,
+  gateway: new URL('../chat/chat.gateway.ts', import.meta.url).pathname,
+};
+
+/** Where an excerpt starts and stops. Whatever sits between the two is dropped,
+ * so a doc comment does not bury the parameter list the panel is about. */
+const CONSTRUCTOR_OPEN = /^\s*constructor\(/;
+const CONSTRUCTOR_END = /^\s*\)\s*\{\}?\s*$/;
```
dunxonu: `CONSTRUCTOR_END` (`/^\s*\)\s*\{\}?\s*$/`) only matches an empty-bodied constructor whose closing `) {}` sits alone on its own line. Both current `SOURCES` entries (`ledger.service.ts`, `chat.gateway.ts`) happen to have empty constructors, so this works today, but adding a third source class with a non-empty constructor body (or reformatting one of the two, e.g. via a lint auto-fix) would make `closes` resolve to -1 or land on the wrong line, silently truncating the DI panel's source excerpt to just the class declaration with no error surfaced.

examples/full/src/jobs/jobs.demo.ts
```diff
@@ -0,0 +1,79 @@
...
+    let job = await queue.getJob(id);
+    let state = job === undefined ? 'missing' : await job.getState();
+    while (
+      (state === 'waiting' || state === 'active' || state === 'delayed') &&
+      Date.now() < deadline
+    ) {
+      await Bun.sleep(POLL_MS);
+      job = await queue.getJob(id);
+      state = job === undefined ? 'missing' : await job.getState();
+    }
```
dunxonu: The "fetch job, resolve state" pair (`queue.getJob(id)` + `job.getState()`) is written once before the `while` and again inside it. A `do/while` (or a small local helper called twice) would remove the duplicated two-line block.

examples/full/src/upstream/flaky.controller.ts
petarzarkov: Fixed in e97f6b2, and this was the one that mattered: I traded a single hardcoded entry for unbounded growth on an unthrottled route of a long-running public demo.

Bounded at 256 with insertion-order eviction rather than evicting on recovery, so the documented "503 twice per key then 200" stays true for every key still in the window; an evicted key starts its two failures over, which is stated on the constant.

Worth recording that the soak could not have caught it: `upstream.flaky` is driven with no `?key=`, so it only ever touches the `default` entry. The guard is a test instead, asserting an evicted key owes its two failures again. It answers 200 without the bound.

examples/full/src/jobs/jobs.demo.ts
petarzarkov: Fixed in e97f6b2, and the reason it was unconditional turns out to be a framework bug rather than a necessity.

`isConnectionError` required `error instanceof RedisError`, so it only matched an error that had passed through `toRedisError`. bullmq holds its own `Bun.RedisClient`, so a broker outage arrives as Bun's `RedisError` with the right code on the wrong class, and the predicate answered false. Going broad was me working around that.

29baae5 keys both predicates on the code, which the comment above `RedisErrorCode` already says is Bun's. So this now checks `isConnectionError` and rethrows otherwise, as you suggested. Verified the tour still exits 0 against an unreachable broker.

examples/full/src/landing/landing.middleware.ts
petarzarkov: Fixed in e97f6b2. Dropped the map and matched `StaticFiles` exactly: Bun sets the type from the extension, and the only header now set is `application/octet-stream` when `file.type === ''`.

examples/full/src/jobs/jobs.demo.ts
petarzarkov: Fixed in e97f6b2, narrowed to `error instanceof Error ? error.message : String(error)`. The catch above it is now conditional too, so a non-Error that is not a connection failure propagates rather than being described.

examples/full/src/landing/vitals.controller.ts
petarzarkov: Fixed in e97f6b2. The cut moved to `constructorExcerpt` in `landing/source-excerpt.ts`, a pure function so the shapes are testable without a server, and the end pattern no longer requires an empty body.

Five cases covered, including the one you named (a constructor with a body) and a one-line constructor, which closes on the line it opened and would otherwise have run the search past it to the end of the class.

examples/full/src/jobs/jobs.demo.ts
petarzarkov: Fixed in e97f6b2, folded into a `do/while`.

dunxonu: Confirmed, thanks. The do/while in e97f6b2 has a single fetch/resolve site now, called once per iteration instead of once before the loop and once inside it.

examples/full/src/landing/retry.controller.ts
```diff
@@ -0,0 +1,87 @@
...
+    outcome: string;
+  }> {
+    // A fresh key per call, so the upstream fails its first two every time
+    // rather than only for the first visitor after a deploy.
+    const key = Math.random().toString(36).slice(2, 10);
+    // Loopback plus the request's port, not `req.url` whole: `Host` decides
+    // that url, so a caller could choose which host this server calls. The port
+    // still comes from the request, since the app's own url is unknown until
+    // `listen()` has run - the reason `UpstreamModule` sets no `baseUrl`. So a
+    // caller can steer the port within loopback, on a demo route.
```
dunxonu: **correctness/security: unthrottled localhost port-scan oracle via the `Host` header**

`req.url`'s port is taken from the client-controlled `Host` header (standard Bun/`Request` behaviour - nothing in `@dunx/http` validates or rewrites `Host`), and only the port survives into the rebuilt target (`http://127.0.0.1:<port>/...`). Since `RetryController` is `@SkipThrottle()`d and world-reachable on the public demo, any anonymous caller can do:

```
curl -H 'Host: x:6379' https://demo.dunx.win/api/demo/retry
curl -H 'Host: x:5432' https://demo.dunx.win/api/demo/retry
```

and use the JSON response (`recovered after N upstream calls` vs `gave up on transport failure`/a real status) as a binary open/closed oracle for arbitrary ports on `127.0.0.1` of the host running the demo - a blind SSRF/port-scan primitive against whatever else is bound to loopback there (Redis, Postgres, MinIO), with no rate limit.

The comment above this code already flags "a caller can steer the port within loopback, on a demo route" as accepted, but on a public demo box that's still a live localhost port scanner exposed to the internet - worth at least removing `@SkipThrottle()` here, or clamping to a small allow-list of ports (just the app's own).

examples/full/src/landing/source-excerpt.ts
```diff
@@ -0,0 +1,44 @@
...
+ * comment does not bury the parameter list; a class with no constructor yields
+ * the declaration alone. The end pattern does not require `) {}`, which used to
+ * truncate a bodied constructor to the class line with nothing reporting it.
+ */
+export const constructorExcerpt = (source: string): string => {
+  const lines = source.split('\n');
+  const opens = lines.findIndex((line) => line.startsWith('export class '));
+  if (opens === -1) return '';
+
+  // Bounded by the class body's close: unbounded, a first class with no
```
dunxonu: **plausible fragility: class-end detection breaks on any column-0 `}` before the constructor**

`ends` is found as the first line after the class declaration that `startsWith('}')`. This works for the two current `SOURCES` targets (`ledger.service.ts`, `chat.gateway.ts`) because everything inside their class bodies is indented, but it's not actually anchored to brace depth - a class whose body contains a multi-line object/type literal or arrow function that happens to close at column 0 (a real possibility after a formatter re-wrap) would set `limit` to that line. If the constructor sits after it, `ctor` search (bounded by `index < limit`) silently fails and the endpoint falls back to returning just the class declaration line - no error, just a wrong/empty DI panel result for whichever file trips it.

Worth anchoring on `import.meta` info the class is used just once (e.g. this is fine to leave as-is if it's understood as coupled to the two files in `SOURCES`), but if a third source is ever added to that map it should be re-verified against this heuristic rather than assumed to work.

examples/full/src/landing/retry.controller.ts
petarzarkov: Taken, and you are right that my own comment was the problem: I wrote "a caller can steer the port within loopback, on a demo route" as an accepted residual, which was arguing for the defect rather than fixing it. On a box with Redis, Postgres and MinIO on loopback that is an anonymous port scanner with an oracle in the response body.

Fixed in 71c0491. Nothing the caller sends reaches the target: `SelfOrigin` carries the url `listen()` returned, set by `main.ts` and by the suite after their own listens. An unset origin is a 503 rather than a guess, so a boot that forgets fails closed instead of calling an origin nobody chose.

`@SkipThrottle()` is gone from this controller as well. It is click-driven, unlike `/api/demo/vitals` which the page polls every five seconds, and it makes three outbound calls per request.

Test sends `Host: scanner.example:6379` and asserts the call still lands on this app.

For the record on why config could not answer it: `publicUrl` is `http://localhost:3000` by default and the suite listens on port 0, so that was my first attempt and it sent the call off-process. There is no post-`listen` lifecycle hook, and `listen()` hands the live server to `ClientAddress` privately, so an explicit set is the only thing left.

examples/full/src/landing/source-excerpt.ts
petarzarkov: Fixed rather than left coupled, since the fix is smaller than the caveat. Bounding on a column-0 `}` was never anchored to brace depth, as you say.

71c0491 bounds on the next top-level class declaration instead, which is what the bound was actually for: the bug it fixes is a first class with no constructor borrowing the next class's. Brace depth stops mattering. Added a test for your exact shape, an object literal re-wrapped so it closes at column 0 above the constructor.

examples/full/src/landing/source-excerpt.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Whatever sits between the class line and the constructor is dropped, so a doc
+ * comment does not bury the parameter list; a class with no constructor yields
+ * the declaration alone. The end pattern does not require `) {}`, which used to
+ * truncate a bodied constructor to the class line with nothing reporting it.
+ */
+export const constructorExcerpt = (source: string): string => {
+  const lines = source.split('\n');
+  const opens = lines.findIndex((line) => line.startsWith('export class '));
```
dunxonu: `opens` is found with a literal `line.startsWith('export class ')`, but `CLASS_DECLARATION` (line 8) - used a few lines below to find where the class body ends - is a regex that also matches `export abstract class` and unexported `class`. If a source file ever added to `SOURCES` (in `vitals.controller.ts`) starts with `export abstract class Foo {`, this line never finds it, `opens` stays `-1`, and `constructorExcerpt` silently returns `''` - the DI panel would render nothing with no error surfaced, rather than the class declaration. Worth reusing `CLASS_DECLARATION` here too so the "what is a class declaration" check is stated once, not twice with different reach.

```suggestion
  const opens = lines.findIndex((line) => CLASS_DECLARATION.test(line));
```

examples/full/src/landing/retry.controller.ts
```diff
@@ -0,0 +1,81 @@
...
+    const started = Bun.nanoseconds();
+    const at = (): number =>
+      Number(((Bun.nanoseconds() - started) / 1e6).toFixed(1));
+
+    try {
+      const recovered = await this.http.get<{ recovered: true; after: number }>(
+        target,
+        {
+          retry: {
+            maxRetries: 3,
```
dunxonu: `maxRetries: 3` is only correct because `FlakyController.#FAILURES` (examples/full/src/upstream/flaky.controller.ts) is hardcoded to `2` - three attempts (1 + 2 retries) is exactly enough to see the upstream recover. Nothing ties the two constants together; bumping `#FAILURES` to `3` for some other demo tweak silently flips this route's `outcome` to `"gave up on 503"` for every visitor. `service.test.ts`'s `toHaveLength(3)` assertion would catch it in CI, but the coupling itself is undocumented at either end. Low severity given the test guard, but worth a comment noting the dependency (or deriving `maxRetries` from a constant `FlakyController` exposes).

examples/full/src/landing/source-excerpt.ts
petarzarkov: Taken as suggested, in d9a89cb. Both ends read `CLASS_DECLARATION` now, so there is one answer to what a class declaration is rather than a literal at one end and a wider regex at the other. Added a test for `export abstract class`, which returned an empty string before.

examples/full/src/landing/retry.controller.ts
petarzarkov: Taken, in d9a89cb, deriving rather than commenting: `FLAKY_FAILURES` is exported from `flaky.controller.ts`, and both `maxRetries` and the suite's attempt-count assertion read it. So the retry budget follows the constant instead of happening to match it.

Fair to note the test guard was weaker than it looked, since `toHaveLength(3)` was a second hardcoded copy of the same number.

dunxonu: Confirmed in d9a89cb: `FLAKY_FAILURES` is exported from `flaky.controller.ts`, `maxRetries` is `FLAKY_FAILURES + 1`, and both `service.test.ts` assertions derive their expected length from it instead of a hardcoded `3`. That closes the coupling this comment flagged.

examples/full/src/upstream/flaky.controller.ts
```diff
@@ -13,18 +15,42 @@ import {
...
 @Controller('upstream')
 @SkipThrottle()
 export class FlakyController {
   readonly #failures = new Map<string, number>();
 
-  /** 503 for the first two calls on a key, then 200. */
+  /** The key is caller-supplied and this route is unthrottled, so an unbounded
+   * map grows for the life of the process. `Map` keeps insertion order, so the
+   * oldest goes first and an evicted key starts its two failures over. */
+  static readonly #MAX_KEYS = 256;
```
dunxonu, line 31: This adds a hand-rolled bounded-map / insertion-order eviction (`#MAX_KEYS`, manual `keys().next()` eviction) for a caller-supplied key. `packages/http/src/throttle/store.ts`'s `MemoryThrottleStore` already solves the "unbounded caller-keyed Map" problem for rate limiting (different eviction policy - TTL sweep vs size cap - but the same underlying hazard). Per CLAUDE.md Rule 1/Rule 2 ("never invent what a mature library/pattern already solves" / "one declaration at the lowest common owner"), worth considering whether this belongs as a small shared bounded-map utility instead of being reimplemented here - especially since this file is vendored byte-for-byte into `tools/create-app/templates/features/upstream/flaky.controller.ts`, so the bespoke eviction logic ships into every scaffolded app.

examples/full/src/jobs/jobs.demo.ts
```diff
@@ -0,0 +1,74 @@
...
+        'the publish side, and bullmq forks jobs.processor.ts to run the handler',
+    );
+
+    // Re-fetched, not polled on the handle publish returned: `returnvalue` is
+    // filled at load time, so that handle never sees it.
+    const startedAt = Date.now();
+    const deadline = startedAt + SETTLE_MS;
+    let job;
+    let state;
+    do {
```
dunxonu: This polls `getJob`/`getState` every 150ms for up to 8s to detect completion of a job handled in a forked process. bullmq offers `QueueEvents`, an event-based cross-process completion notification over Redis pub/sub, which would avoid the poll loop and its worst-case 8s wait. Given this is demo code meant to showcase the queue capability end to end, a `QueueEvents`-based wait may read as more representative of how a real consumer should observe completion than manual polling.

examples/full/scripts/og-card.ts
```diff
@@ -0,0 +1,105 @@
...
+  'transactions',
+  'openapi',
+  'rate limits',
+] as const;
+
+const cardHtml = (): string => `<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8" />
+<style>
```
dunxonu: This ~50-line `<style>` block is largely copy-pasted from `internal/docs/scripts/og-card.ts`'s card markup (same layout classes: `.brand`, `.live`, `h1`, `.accent`, `.chips`/`.chip`, same grid-lattice background), with only colors/copy differing. `scripts/og-card.ts` exists specifically to be the shared mechanism between the two card generators (per its own doc comment), but the actual HTML/CSS shell is still duplicated between the two site-specific files. A shared `cardShell()`/template helper in `scripts/og-card.ts` would remove the duplicate-maintenance cost of keeping two near-identical stylesheets in sync.

petarzarkov: Taken, and this was the strongest of the three: `scripts/og-card.ts` was added to be the shared half and only held the screenshot, so the stylesheet stayed duplicated. Their sizes had already drifted, 58px against 62px on the headline and 53ch against 50ch on the blurb, which is the drift starting.

77dd6f2 moves the shell in as `cardHtml(style, content)`. Each generator supplies six colours and its own words: `internal/docs/scripts/og-card.ts` goes from 104 lines to 54, `examples/full/scripts/og-card.ts` from 108 to 46. Both cards re-rendered, checked and committed.

examples/full/src/upstream/flaky.controller.ts
petarzarkov: Checked this rather than assumed it, and I am declining it, with the reason.

`MemoryThrottleStore` does not solve the hazard here. Its bound is `if (size >= maxKeys) this.#sweep(now)`, and `#sweep` deletes only entries whose `expiresAt` has passed. Against a counter that never expires nothing is sweepable, so the map grows past the cap anyway - it is bounded by TTL, with `maxKeys` as the trigger to sweep, not by size. Its surface is also `hit(key, windowSeconds)` / `ttl(key)`, windows and counts, so the flaky counter would be injecting a rate-limit store to count demo failures.

Extracting a shared bounded map instead would be new published surface in `@dunx/http` whose only consumer is one example controller, which is the speculative abstraction CLAUDE.md rules out and which docs/ROADMAP.md freezes without an external user.

Your point about it shipping into every scaffold via the vendored template is fair, and it is six lines doing the correct thing for a scaffold too. I would rather revisit it if a second caller appears.

examples/full/src/jobs/jobs.demo.ts
petarzarkov: `QueueEvents` is the better pattern and I agree it reads as more representative. Not taking it here, for a reason I would rather state than skip.

`@dunx/infra/queue` exposes no connection it could use: `QueueConnection` keeps its clients private, and `JobPublisher` exposes only `queue(name): Queue`. So the demo would have to reach through `Queue.opts.connection` into bullmq internals, or build its own client - in a file vendored byte-for-byte into every scaffolded app. It also adds a second broker connection to a script that has to exit 0, and `tour.ts` already carries a note about bullmq's `close()` never resolving against an unreachable broker (taskforcesh/bullmq#4656).

The real gap is that `@dunx/infra/queue` does not expose a completion signal, and that is where this belongs. I have not added it to docs/ROADMAP.md because another session is editing that file in this working tree right now and I do not want to conflict with it.

dunxonu: Agreed, that holds up. `JobPublisher` genuinely exposes no connection here, so the choice was between the poll loop, reaching into `Queue.opts.connection`, or standing up a second broker client in demo code - and the poll loop is the least bad of those for something that has to exit 0 without a broker. The completion signal belongs in `@dunx/infra/queue` itself, per [Rule 2](https://github.com/petarzarkov/dunx/blob/main/CLAUDE.md#rule-2---one-declaration-and-it-lives-at-the-lowest-common-owner) - the package that owns the connection is where a `QueueEvents`-based wait should live, not something the example reimplements around it. Dropping this here; make sense to track it as its own roadmap item once the current edit to that file lands.

petarzarkov: Reversed - you were right and the repo owner agreed, so this is now exposed from the framework rather than worked around.

6f08e8d adds `JobEvents` to `@dunx/infra/queue`. It contributes the two things bullmq cannot know, a client over `Bun.RedisClient` from `QueueConnection` and a lifetime that ends before those sockets close, and `events(name)` returns bullmq's own `QueueEvents` for `Job.waitUntilFinished(events, ttl)` to wait on. That is the division `JobPublisher.queue()` already draws. Streams open on first use, since `autorun` starts a blocking read in the constructor. `jobs.demo.ts` waits instead of polling.

Two things measured on the way, both on bullmq 6.3.4:

- `QueueEvents.close()` against an unreachable broker neither resolves nor rejects. The blocking read reconnects on a growing backoff, past `autoReconnect: false`, and it threw uncaught. So the close is bounded at 2s per stream, never rejects, and logs what it abandoned; `QueueConnection.onShutdown` closes the socket under it regardless. Without that, adding this would have hung shutdown - which is the concern I raised for the wrong reason.
- A stream under the wrong prefix never fires, so the prefix is passed through and asserted.

Documented in docs/guide/15-queues.md, "Waiting for a job to finish", including why a status endpoint still polls: an HTTP request cannot hold a socket open for eight seconds.

Also pinned bullmq to 6.3.4 everywhere an exact pin exists. `examples/full`, `@dunx/dashboard` and the scaffold were on 6.0.5 while `@dunx/infra` already built against 6.3.4, so the lockfile carried three copies of bullmq.

dunxonu: Confirmed against 6f08e8d: `JobEvents` lands in `@dunx/infra/queue` exactly where I said the gap belonged, and `jobs.demo.ts` now waits on `Job.waitUntilFinished` instead of polling. Good outcome, thanks for pushing it through rather than leaving the workaround in place.

docs/guide/15-queues.md
```diff
@@ -571,6 +571,46 @@ even while workers drained jobs.
...
+
+  async render(width: number): Promise<Rendered> {
+    const job = await this.jobs.publish('thumbnails', 'render', { width });
+    // Resolves when the worker completes it, rejects on a failure or the ttl.
+    return await job.waitUntilFinished(this.events.events('thumbnails'), 8_000);
+  }
+}
+```
+
+bullmq broadcasts completion over Redis pub/sub, so this costs no polling. The
```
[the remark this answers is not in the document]
petarzarkov: Correct, and the wording was wrong in three places. `queue-events.js` calls `backend.readEvents` and its own comment says "interrupt the blocking XREAD", so it is a Redis stream, not pub/sub. Fixed in 1788b75 in the guide and in the class comment.

packages/infra/src/queue/events.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /** Bounded, and never rejecting: one stream that will not close must not stop
+   * the rest of shutdown, which is what closes the socket under it. */
+  async #close(name: string, events: QueueEvents): Promise<void> {
+    const timedOut = Symbol('timed out');
+    try {
+      const outcome = await Promise.race([
+        events.close(),
+        Bun.sleep(CLOSE_TIMEOUT_MS).then(() => timedOut),
```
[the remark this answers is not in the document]
petarzarkov: Taken, and it was worse than minor once measured: a spawned process that opens one stream and closes it took 2.37s to exit with the `Bun.sleep`, against 0.36s with a clearable timer. Every clean shutdown paid the full two-second window, so the code added to stop shutdown hanging was the thing delaying it.

Fixed in 1788b75 with `setTimeout` and `clearTimeout` in `finally`.

Not unit-tested, and worth saying why rather than leaving it implied: the leak is in process lifetime rather than in what `onShutdown` returns, which resolved in 1ms either way, so reproducing it needs a spawned process - and whether `close()` resolves promptly depends on how far the connection attempt got, which makes the timing unreliable to assert. The numbers above are from a probe against a live broker.

examples/full/src/service.test.ts
```diff
@@ -597,3 +601,150 @@ it('serves the same route over HTTP/2, since createApp sets http2', async () =>
...
+  // Everything else still misses, which is what the tour's 404 step narrates.
+  expect((await at('/not-a-page')).status).toBe(404);
+});
+
+it('ignores the Host header when calling its own flaky route', async () => {
+  // Rebuilt from `req.url` this was a loopback port scanner.
+  const res = await fetch(new URL('api/demo/retry', baseUrl), {
+    headers: { host: 'scanner.example:6379' },
+  });
+  const body = (await res.json()) as { outcome: string; attempts: unknown[] };
```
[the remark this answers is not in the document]
petarzarkov: Not taking this one - I checked and there is no duplicate declaration.

`const body` appears twice in `service.test.ts`, at line 517 and line 745, but in two different `it()` blocks, so they are separate lexical scopes. The Host-header test at 745 declares it once.

The file typechecks: `bun run --filter '@dunx/example-full' typecheck` exits 0, and the Lint, format, types job on this PR is green, which runs `tsc --noEmit` over that workspace. A duplicate lexical declaration would fail both.

Happy to look again if you can point at the two lines you read as sharing a block.

examples/full/src/landing/vitals.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  /** A class the container built, read off disk. Markup would have drifted. */
+  @Get('/source/:name', sourceParams)
+  async source({ params }: Input<typeof sourceParams>): Promise<{
+    name: string;
+    path: string;
+    code: string;
+  }> {
+    const { name } = params;
+    const path = SOURCES[name];
```
dunxonu: **Correctness:** `SOURCES` is a plain object literal (`Record<string, string>`), and `name` is an unvalidated route param (`z.string().max(32)`, no character restriction). Requesting `GET /api/demo/source/__proto__` or `/constructor` or `/toString` makes `SOURCES[name]` resolve through the prototype chain to a non-`undefined` value (`Object.prototype`, the `Object` constructor, etc.), bypassing the `path === undefined` 404 guard on line 141. Execution then reaches `path.split('/src/')` on line 150 with `path` being an object/function rather than a string, throwing an uncaught `TypeError` and turning the intended 404 into an unhandled 500.

Fix: use `Object.hasOwn(SOURCES, name) ? SOURCES[name] : undefined`, or back `SOURCES` with `Object.create(null)` / a `Map`.

```suggestion
    const path = Object.hasOwn(SOURCES, name) ? SOURCES[name] : undefined;
```

examples/full/src/landing/landing.middleware.ts
```diff
@@ -20,15 +31,23 @@ export class LandingMiddleware implements Middleware {
...
     // A miss is thrown rather than returned, so the flag is what reports one.
     if (ctx.get(UNMATCHED) !== true || req.method !== 'GET') return next();
-    if (new URL(req.url).pathname !== '/') return next();
 
-    const page = Bun.file(this.#page);
-    if (!(await page.exists())) return next();
+    const name = FILES[new URL(req.url).pathname];
+    if (name === undefined) return next();
+
+    const file = Bun.file(`${this.#dir}${name}`);
```
dunxonu, line 38: **Altitude:** only `/` needs the index-fallback behavior that `StaticFiles` deliberately omits (per the class comment above). `/landing.css`, `/landing.js` and `/og.png` are exact-path static assets that `StaticFiles` (already mounted for `/assets/` elsewhere in this example) serves generically, including content-type-from-extension and cache-control. This hand-rolls a subset of that (lines 38-51: manual `exists()` check, manual `octet-stream` fallback, manual per-file cache-control) instead of mounting `StaticFiles` for the three literal files and keeping `LandingMiddleware` scoped to the `/` special case it actually needs to handle.

examples/full/src/landing/vitals.controller.ts
petarzarkov: Confirmed and fixed in 51e730e. Reproduced first: `__proto__`, `constructor` and `toString` all answered 500, and 404 after.

Used a `Map` rather than the `Object.hasOwn` guard you suggested, for the same reason but one step further: the guard is correct and has to be remembered by whoever edits the lookup next, whereas `get` has no prototype chain to walk. The test asserts all three keys.

Worth noting the route is an allow-list whose own refusal was what broke, so the failure mode was the guard turning into a 500 rather than the allow-list being bypassed to a real file - `Bun.file(Object.prototype)` reads nothing.

examples/full/src/landing/landing.middleware.ts
petarzarkov: Checked whether this is possible before answering, and it is not, as things stand.

`StaticModule.forRoot` binds `StaticOptions` and `StaticFiles` with no token parameter - unlike `DbModule.forRootAsync(SyncDatabase, ...)` or `HttpModule.forRootAsync(..., HealthClient)`, which take one. `AssetsModule` already mounts one, and `main.ts` registers the resolved `StaticFiles` by that token, so a second mount would bind the same token in a second scope and `app.use(StaticFiles)` would resolve one of the two arbitrarily.

The cache policy does not map cleanly either: `og.png` wants `public, max-age=86400` and the page wants `no-cache`, and `StaticOptions` expresses per-file variation as `immutable(pathname)`, which implies a long max-age rather than none.

So the real shape of this is a framework gap - `StaticModule` cannot be mounted twice - and I would rather record that than hand-roll around it a second time. Not doing it in this PR: it is public API on `@dunx/http`, the payoff is deleting about fifteen lines from one example, and this branch is already fifteen commits deep. Happy to open it separately if you think the token parameter is worth having on its own merits.

The Rule 1 half of this is already gone, from your earlier comment on the same file: the content-type map went, and Bun sets it from the extension with only the unknown-type fallback stated, matching `StaticFiles`.

packages/infra/src/queue/events.ts
```diff
@@ -0,0 +1,132 @@
...
+   * the rest of shutdown, which is what closes the socket under it. */
+  async #close(name: string, events: QueueEvents): Promise<void> {
+    const timedOut = Symbol('timed out');
+    let timer: ReturnType<typeof setTimeout> | undefined;
+    try {
+      const outcome = await Promise.race([
+        events.close(),
+        new Promise<symbol>((resolve) => {
+          timer = setTimeout(() => resolve(timedOut), CLOSE_TIMEOUT_MS);
+        }),
```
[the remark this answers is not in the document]
petarzarkov: Verified this rather than taking it on trust, and it is real - but it is pre-existing, not introduced by `JobEvents`.

Measured through `QueueConnection` against a live broker, counting `duplicate()` calls on the client handed over:

```
QueueEvents  duplicate() calls: 1
Queue        duplicate() calls: 0
Worker       duplicate() calls: 1
```

So `Worker` has done exactly this since the queue package existed, which `QueueConnection.client()`'s own comment already says: "bullmq duplicates whatever it is given to get a connection it may block on". `#open` tracks only the originals, and `conn.open` reports 2 where three sockets exist.

The fix is not the one-liner it looks like, and this is the part your suggestion misses. On a duplicate obtained through `QueueConnection`, `adapter.raw` is `undefined`: `boundClientClass` overrides the raw client's `duplicate()` to return a `Promise`, so the adapter has no raw client at the moment the wrapper sees it. `onShutdown` closes both halves - `adapter.disconnect()` then `raw.close()` - and the second is there precisely because `disconnect()` skips a client that never connected and would otherwise keep the process alive. So tracking duplicates means either awaiting that promise inside a synchronous wrapper or accepting `disconnect()`-only teardown for them.

Not doing it in this PR, and the reason is risk rather than effort: it changes the shutdown path for every dunx app that touches a queue, `internal/notes/roadmap/queue-shutdown-sigterm.md` already tracks an unresolved SIGTERM hang in the same code, and twice in this branch a fix of mine introduced something worse than what it fixed. It deserves its own change with the SIGTERM note in front of it.

On the test you asked for: I could not make one honest here either. Whether `close()` resolves promptly depends on how far the connection attempt got, so an assertion about a pending disconnect is timing-dependent.

examples/full/src/landing/landing.middleware.ts
dunxonu: Checked this against the source and it holds: `StaticModule.forRoot`/`forRootAsync` bind `StaticOptions`/`StaticFiles` with no token, unlike `DbModule.forRootAsync(token, ...)`, and `AssetsModule` already mounts one with `main.ts` registering `app.use(StaticFiles)` off that same class token - so a second mount here would collide exactly as you describe. That's a real gap in `StaticModule`, not a reason to keep the hand-rolled version; agreed it's worth its own PR (Rule 2, https://github.com/petarzarkov/dunx/blob/main/CLAUDE.md#rule-2---one-declaration-and-it-lives-at-the-lowest-common-owner) rather than blocking this one on it.

packages/infra/src/queue/events.ts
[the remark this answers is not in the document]
petarzarkov: @coderabbit yes please - follow it up with a github issue

PR #64, opened 2026-09-08
feat(mcp): serve the guide and scaffold catalogue with no app

petarzarkov:
`bunx @dunx/mcp` required a root module, so it was unreachable during adoption: every tool read an app, and someone installing dunx for the first time has none. The entry is now optional and three tools answer without one.

- `dunx_start` (4 KB): runtime, scaffold command, the three commands that add dunx to an existing project, guide index, and the four rules that fail at boot.
- `dunx_guide`: the 23 chapters, as an index, one chapter, or a line-level search. Also served as MCP resources at `dunx://guide/<slug>`, which is what `resources/list` and `resources/read` were added for.
- `dunx_scaffold`: create-app's own catalogue, plus the minimal app's source.

The corpus is bundled at build time into a committed `generated.ts`, same arrangement as `packages/dashboard/src/ui-bundle.ts`. `scripts/gen-mcp-corpus.test.ts` re-renders and compares, so a guide edit that skipped the generator fails instead of shipping the previous release's docs.

Also fixes a real bug: `tools/mcp/README.md` documented five tools (`list_routes`, `describe_route`, …) that were renamed to `dunx_*` and no longer exist.

Reasoning is in `docs/architecture/mcp.md`, "The entry became optional". `bun run ci` passes 13/13 on a clean checkout of this branch.

**Not done:** no `examples/full` change, so Rule 4 is unsatisfied in the literal sense. An MCP stdio server has no way to be exercised by a long-running HTTP example, and `@dunx/create-app` has the same gap. A CI step running the built `dist/cli.js` with no entry would be the honest substitute.

## Summary by CodeRabbit

- **New Features**
  - MCP can start without an app, providing guidance, starter scaffolding, feature discovery, and boot requirements.
  - Added searchable guide chapters as MCP resources.
  - App-specific mode provides tools for application routes, providers, modules, gateways, overview, and OpenAPI details.
  - Added scaffolding setup steps, starter files, installation commands, and feature discovery.

- **Documentation**
  - Updated MCP usage, architecture, package descriptions, and adoption guidance.
  - Clarified optional app entry handling and available tool modes.

- **Bug Fixes**
  - Review findings are now submitted as a single pull request review instead of separate inline comments.


petarzarkov:
@coderabbit please review this

dunxonu:
**Claude finished @petarzarkov's task in 8m 11s** —— [View job](https://github.com/petarzarkov/dunx/actions/runs/34196497516)

---
### Review of #64 (`feat/mcp-adoption-tools`)

- [x] Read PR context, CLAUDE.md, and CodeRabbit's earlier review
- [x] Run `/code-review medium` over the diff (`origin/main...HEAD`)
- [x] Verify CodeRabbit's fragment-link finding independently
- [x] Post final review

CodeRabbit's fragment-link finding is already fixed on this branch: `protocol.ts:71,195` (`withoutFragment`) falls back to a fragment-stripped match, and `protocol.test.ts:403-417` covers both "reads one by a uri carrying a fragment" and "still rejects a fragment on a uri that names no resource". Nothing left to do there.

Two things worth a look before merge:

### 1. `summarize()` re-implements `internal/docs`'s `summaryOf()`, more weakly
`scripts/gen-mcp-corpus.ts:44-61` builds each chapter's `summary` (shown in `dunx_guide`'s index and in the MCP resource `description`) by joining raw lines up to the first blank/heading/fence. `internal/docs/scripts/agent-docs.ts:29` (`summaryOf`) solves the same problem for `llms.txt` and already strips `[link](url)`/`**bold**` markdown and caps at 200 chars. The new one doesn't, so a chapter whose opening paragraph has a link or bold text ships those markdown artifacts verbatim into the MCP corpus, and the repo now has two summaries of "a chapter's first paragraph" that can silently disagree - the exact case CLAUDE.md's Rule 2 calls out ("a traversal or a reader two packages need moves down to the package that owns the data"). `agent-docs.ts` can't be imported from a repo script (it's in `internal/docs`, private), but the extraction logic itself could move somewhere both can reach, or `gen-mcp-corpus.ts` could call the same regex pipeline.
[Fix this →](https://claude.ai/code?q=In%20scripts%2Fgen-mcp-corpus.ts%2C%20the%20summarize()%20function%20(lines%2044-61)%20duplicates%20internal%2Fdocs%2Fscripts%2Fagent-docs.ts%27s%20summaryOf()%20(line%2029)%20with%20weaker%20output%3A%20it%20doesn%27t%20strip%20markdown%20links%2Fbold%20or%20cap%20length.%20On%20branch%20feat%2Fmcp-adoption-tools%2C%20factor%20the%20shared%20%27first%20paragraph%2C%20markdown-stripped%2C%20capped%20at%20200%20chars%27%20logic%20into%20one%20place%20both%20scripts%20can%20use%2C%20since%20agent-docs.ts%20lives%20in%20private%20internal%2Fdocs%20and%20can%27t%20be%20imported%20from%20a%20repo-level%20script.&repo=petarzarkov/dunx)

### 2. `Guide.chapter()` numeric-prefix lookup is ambiguous for real, today
`docs/guide/` has both `22-metrics.md` and `22-upgrading.md`. `Guide.chapter()` (`tools/mcp/src/guide.ts:87-94`) tries exact slug, then `slug.includes(wanted)`, then `title.includes(wanted)` - so `dunx_guide({ topic: "22" })` returns `22-metrics` (first in sorted order) and never `22-upgrading` by that route. It's not fully silent - `adopt.ts:111-113` does attach `otherMatches` to the response, so a caller sees `22-upgrading` was also a candidate and can retry with the exact slug - but the first call still spends the tokens on the (arguably wrong) chapter body with no error. Minor, and self-correcting via `otherMatches`, but worth knowing before someone hits it live.

Related, lower severity: `Args.text()` (`tools/mcp/src/args.ts:41-44`) only treats an exactly-empty string as absent. A `topic: "   "` reaches `Guide.chapter()`, whose `.trim()` turns it into `""`, and `"".includes("")` is true for every doc, so the first chapter (`01-introduction`) comes back as if it matched, with `otherMatches` listing literally every other chapter - a sign the query was garbage rather than a real disambiguation list.

Everything else lines up with CLAUDE.md: `Args`/`Guide`/`Scaffold` are stateful classes per Rule 3, `rewriteLinks`'s duplication guard in `gen-mcp-corpus.test.ts` correctly fails the corpus stale rather than accepting whatever's committed, and the "not done" callout (no `examples/full` change, since an MCP stdio server has no example to exercise it in) is an honest and reasonable exception to Rule 4 rather than a silent gap.

dunxonu:
**Claude finished @petarzarkov's task in 11m 24s** —— [View job](https://github.com/petarzarkov/dunx/actions/runs/34197507700)

---
### Review of #64 (`feat/mcp-adoption-tools`)

- [x] Read PR context, CLAUDE.md, prior CodeRabbit + Claude reviews
- [x] Run `/code-review medium` over the diff (`origin/main...HEAD`)
- [x] Verify each finding against the current source
- [x] Post final review

Both items from the previous Claude review are fixed on this branch: `summaryOf()` was extracted into `scripts/guide-summary.ts` and `internal/docs/scripts/agent-docs.ts` now re-exports it (commit `aaf40f8`), and `Guide.chapter()` returns `undefined` on an ambiguous numeric prefix with `candidates()` covering the disambiguation. Nothing left there.

### 1. `RULES` in `adopt.ts` restates `agents.ts`'s rule list, and the two already disagree
`tools/mcp/src/adopt.ts:21-42` (`dunx_start`'s answer) and `tools/create-app/src/agents.ts:27-47` (the scaffolded app's `AGENTS.md`) both spell out the same "boot error, not a preference" rules for the same audience - constructor injection needs the preload, no parameter decorators, `.js` on relative imports, module exports are the encapsulation boundary. They're not copies: `adopt.ts` has four entries, `agents.ts` has six (it also covers `reflect-metadata`/`experimentalDecorators` and the `bun`-only rule), and the wording differs throughout. CLAUDE.md's Rule 2 calls this out by name ("a traversal or a reader two packages need moves down to the package that owns the data") - a future fifth boot-error rule, or a wording fix to an existing one, has nothing forcing the other list to follow, so `dunx_start`'s answer and a scaffolded app's own docs will drift on the same subject matter.
[Fix this →](https://claude.ai/code?q=In%20tools%2Fmcp%2Fsrc%2Fadopt.ts%20lines%2021-42%2C%20the%20RULES%20array%20duplicates%20tools%2Fcreate-app%2Fsrc%2Fagents.ts%20lines%2027-47%27s%20RULES%20template%20string%20-%20both%20list%20the%20same%20dunx%20boot-error%20rules%20for%20the%20same%20audience%20but%20with%20different%20content%20(4%20vs%206%20entries)%20and%20wording.%20On%20branch%20feat%2Fmcp-adoption-tools%2C%20factor%20the%20shared%20rule%20list%20into%20one%20place%20both%20tools%2Fmcp%20and%20tools%2Fcreate-app%20can%20import%20from.&repo=petarzarkov/dunx)

### 2. `gen-mcp-corpus.ts` reimplements two things `internal/docs/scripts/content.ts` already solved
- **Title extraction** (`scripts/gen-mcp-corpus.ts:58-62`): `lines.find(l => l.startsWith('# '))?.slice(2).trim()` vs. `content.ts:229-232`'s `titleOf()`, which does the same plus strips inline backticks. A chapter heading with inline code (`` # The `dunx_guide` tool ``) ships literal backticks into `dunx_guide`'s index and every MCP resource name, while the doc site's rendering of the same heading (via `titleOf`) doesn't have them - two surfaces disagreeing about the same chapter's title.
- **Link rewriting** (`scripts/gen-mcp-corpus.ts:38-42`): `rewriteLinks()` hardcodes fixed-depth `../`/`../../` regex replacement, where `content.ts:38-47`'s `resolveFrom()` resolves relative links against the actual source path. Every current chapter lives at exactly `docs/guide/*.md` so the fixed-depth assumption happens to hold today, but this is the same relative-path resolution bug class the "read a guide chapter through a link carrying a fragment" fix (commit `70cddd7`) already fixed once in `content.ts` - reintroduced here as a second, less general implementation that a subdirectory move or a three-level-up link would silently mis-rewrite instead of failing `gen-mcp-corpus.test.ts`'s drift check.

`content.ts` lives in private `internal/docs` and can't be imported from a repo-level script, which is exactly the constraint this PR already solved once for `summaryOf` via `scripts/guide-summary.ts` - the same move applies to `titleOf` and the link-rewrite logic.
[Fix this →](https://claude.ai/code?q=In%20scripts%2Fgen-mcp-corpus.ts%2C%20the%20inline%20title%20extraction%20(lines%2058-62)%20and%20rewriteLinks()%20(lines%2038-42)%20duplicate%20internal%2Fdocs%2Fscripts%2Fcontent.ts%27s%20titleOf()%20(line%20229)%20and%20resolveFrom()%2FrewriteHref()%20(lines%2038-90)%20with%20weaker%2Fless-general%20behavior%3A%20no%20backtick%20stripping%2C%20and%20fixed-depth%20..%2F%20assumptions%20instead%20of%20resolving%20against%20the%20actual%20source%20path.%20On%20branch%20feat%2Fmcp-adoption-tools%2C%20apply%20the%20same%20fix%20this%20PR%20already%20used%20for%20summaryOf()%3A%20extract%20the%20shared%20logic%20into%20scripts%2F%20so%20both%20internal%2Fdocs%20and%20gen-mcp-corpus.ts%20can%20import%20it.&repo=petarzarkov/dunx)

### Minor, both low severity and currently dormant
- **`tools/mcp/src/guide.ts:142`** - `search()`'s hit `text` is `line.trim().slice(0, MAX_HIT_LENGTH)` unconditionally, not a window around the match. A guide line over 200 chars with the query occurring after character 200 would still report as a hit whose `text` doesn't contain the query. No line in `docs/guide/*.md` exceeds 200 chars today (`wc -L` confirms 184 is the longest), so this doesn't currently fire.
- **`scripts/guide-summary.ts:22-23`** - the sentence-end regex `/^(.+?\.)(?:\s|$)/` stops at the first `. `, so an opening paragraph containing `e.g. ` before its real terminating period would truncate mid-sentence. No current chapter's first paragraph contains `e.g.`/`i.e.`, so also latent.
- **`tools/mcp/src/cli.ts:92-101`** - if only `--export=<name>` is passed with no entry path, `assemble()` returns the adoption-only tools and the stderr message ("No entry given...") never mentions that `--export` was seen and ignored. Cosmetic - the resulting tool set is correct either way.

Everything else lines up with CLAUDE.md: `Args`/`Guide`/`Scaffold` stay stateful classes per Rule 3, the corpus drift test (`gen-mcp-corpus.test.ts`) fails a stale commit rather than accepting it, and the "not done" callout (no `examples/full` change, since an MCP stdio server has no long-running HTTP example to exercise it in) is a reasonable exception to Rule 4 rather than a silent gap.
 · [branch](https://github.com/petarzarkov/dunx/tree/feat/mcp-adoption-tools)

dunxonu:
**Claude finished @petarzarkov's task in 2m 34s** —— [View job](https://github.com/petarzarkov/dunx/actions/runs/34198468026)

---
### Review of #64 (`feat/mcp-adoption-tools`)

- [x] Read PR context, CLAUDE.md, and both prior dunxonu reviews
- [x] Run `/code-review medium` over the diff (`origin/main...HEAD`)
- [x] Verify prior findings against current source (HEAD `3c1c737`)
- [x] Post final review

Since the last review, `90fe33a` and `e8b0dc1` closed out the fragment-link finding, the `summaryOf()` duplication, the `Guide.chapter()` ambiguity, and the other items from that pass (search cap, `bunfig()`, `resources` capability gating, corpus regeneration moved out of `build`). Verified against the current tree, not just the commit messages.

Two findings from the second review are still open on this branch:

### 1. `RULES` in `adopt.ts` still disagrees with `agents.ts`'s rule list
`tools/mcp/src/adopt.ts:21-38` and `tools/create-app/src/agents.ts:27-47` still each spell out the same "boot error, not a preference" rules for the same audience, with different content (4 entries vs 6 - `adopt.ts` still omits the `reflect-metadata`/`experimentalDecorators` and bun-only rules) and different wording throughout. Neither commit since touched `agents.ts`. This is the exact case CLAUDE.md's Rule 2 names: "a traversal or a reader two packages need moves down to the package that owns the data." A future fifth boot-error rule has nothing forcing both lists to update together.
[Fix this →](https://claude.ai/code?q=In%20tools%2Fmcp%2Fsrc%2Fadopt.ts%20lines%2021-38%2C%20the%20RULES%20array%20duplicates%20tools%2Fcreate-app%2Fsrc%2Fagents.ts%20lines%2027-47%27s%20RULES%20template%20string%20-%20both%20list%20the%20same%20dunx%20boot-error%20rules%20for%20the%20same%20audience%20but%20with%20different%20content%20(4%20vs%206%20entries)%20and%20wording.%20On%20branch%20feat%2Fmcp-adoption-tools%2C%20factor%20the%20shared%20rule%20list%20into%20one%20place%20both%20tools%2Fmcp%20and%20tools%2Fcreate-app%20can%20import%20from.&repo=petarzarkov/dunx)

### 2. `gen-mcp-corpus.ts` still reimplements title extraction and link rewriting that `content.ts` already solved
This is the one item from the summary-duplication fix that didn't get applied to its two siblings:
- **Title** (`scripts/gen-mcp-corpus.ts:58-62`): `lines.find(l => l.startsWith('# '))?.slice(2).trim()` still doesn't strip inline backticks, unlike `internal/docs/scripts/content.ts`'s `titleOf()`. A chapter heading with inline code still ships literal backticks into `dunx_guide`'s index and every MCP resource name.
- **Link rewriting** (`scripts/gen-mcp-corpus.ts:38-42`): `rewriteLinks()` still hardcodes fixed-depth `../`/`../../` regex replacement rather than resolving against the actual source path the way `content.ts`'s `resolveFrom()` does. Works today because every chapter lives at exactly `docs/guide/*.md`, but it's the same relative-path bug class the fragment fix already addressed once, reintroduced here as a second, less general implementation.

`scripts/guide-summary.ts` is the precedent this PR already established for exactly this constraint (`content.ts` lives in private `internal/docs` and can't be imported from a repo script) - the same move applies to `titleOf` and the link-rewrite logic.
[Fix this →](https://claude.ai/code?q=In%20scripts%2Fgen-mcp-corpus.ts%2C%20the%20inline%20title%20extraction%20(lines%2058-62)%20and%20rewriteLinks()%20(lines%2038-42)%20still%20duplicate%20internal%2Fdocs%2Fscripts%2Fcontent.ts%27s%20titleOf()%20and%20resolveFrom()%2FrewriteHref()%20with%20weaker%2Fless-general%20behavior%3A%20no%20backtick%20stripping%2C%20and%20fixed-depth%20..%2F%20assumptions%20instead%20of%20resolving%20against%20the%20actual%20source%20path.%20On%20branch%20feat%2Fmcp-adoption-tools%2C%20apply%20the%20same%20fix%20already%20used%20for%20summaryOf()%20in%20scripts%2Fguide-summary.ts%3A%20extract%20the%20shared%20logic%20so%20both%20internal%2Fdocs%20and%20gen-mcp-corpus.ts%20can%20import%20it.&repo=petarzarkov/dunx)

### Still dormant, unchanged from the prior review (low severity, no live case today)
- `tools/mcp/src/guide.ts:142` - `search()`'s hit `text` is a flat 200-char slice, not a window around the match; no current guide line exceeds 184 chars.
- `scripts/guide-summary.ts:22-23` - the sentence-end regex stops at the first `. `, so an opening paragraph containing `e.g. ` would truncate early; no current chapter's opening paragraph has one.
- `tools/mcp/src/cli.ts:92-101` - `--export=<name>` with no entry path is silently accepted and ignored, with no mention in the resulting message.

Everything else lines up with CLAUDE.md: the `Args` extraction into `tools/mcp/src/args.ts` (`tools.ts:1-190`) is a clean behavior-preserving refactor, the corpus drift test now actually runs against a stable committed file since `build` no longer regenerates it, and the `dunxonu-review.yml` changes (single sticky comment, full checkout history, wider read-only allowlist) match what the commit message claims.
 · [branch](https://github.com/petarzarkov/dunx/tree/feat/mcp-adoption-tools)

dunxonu:
### Review of #64 (`feat/mcp-adoption-tools`)

- [ ] Read PR context, CLAUDE.md, and prior CodeRabbit + dunxonu reviews
- [ ] Verify latest fixes (`90fe33a`, `0b82554`) against current source
- [ ] Run `/code-review medium` over the diff (`origin/main...HEAD`)
- [ ] Post final review

dunxonu (commented):
```json
[
  {
    "file": "docs/architecture/mcp.md",
    "line": 143,
    "summary": "The docs say `dunx_start` bundles \"four rules\", but `BOOT_RULES` in tools/create-app/src/rules.ts actually has seven, and the four concepts listed (preload, parameter decorators, `.js` extension, module encapsulation) omit three real rules the tool ships (the reflect-metadata ban, the erased-type-parameter rule, and \"bun only\").",
    "failure_scenario": "A reader (or an agent following `dunx_start`'s own docs) trusts \"four rules\" as exhaustive and never learns about the reflect-metadata ban, the erased-constructor-parameter rule, or the bun-only rule - three of the seven boot errors `dunx_start` actually returns at runtime (`gen-mcp-corpus.test.ts` even asserts `RULES.length` is greater than 4, confirming the count grew past what the prose says)."
  },
  {
    "file": "docs/guide/21-agent-tooling.md",
    "line": 50,
    "summary": "Same \"four rules\" miscount repeated in the published guide's tool table, which is public-facing documentation.",
    "failure_scenario": "A consumer reading the published guide table expects `dunx_start` to carry four boot-error rules; calling it returns seven, so the table undercounts what the tool actually answers with."
  },
  {
    "file": "tools/mcp/src/scaffold.ts",
    "line": 83,
    "summary": "`Scaffold.steps()` builds `bun add ${dependencies.join(' ')}` / `bun add -d ${devDependencies.join(' ')}` with no guard for an empty array.",
    "failure_scenario": "If `examples/minimal/package.json` ever ships with no `dependencies` or no `devDependencies` (nothing enforces non-empty today), `dunx_start`'s `addToExistingProject.commands` would hand an agent the literal command `bun add ` (or `bun add -d `) with a trailing space and no package name, which is a broken install instruction."
  }
]
```

dunxonu (commented):
## Findings

```json
[
  {
    "file": "scripts/gen-mcp-corpus.ts",
    "line": 169,
    "summary": "The header written into the generated corpus (and its committed copy at tools/mcp/src/generated.ts:1) tells a maintainer to run `bun run build` in tools/mcp to regenerate it, but tools/mcp's build script is only `bun ../../scripts/build-package.ts` and does not call the generator, and docs/architecture/mcp.md explicitly documents this as the fix (\"`build` does not regenerate it, and that is the whole point\").",
    "failure_scenario": "A contributor edits docs/guide/06-validation.md, reads the comment atop tools/mcp/src/generated.ts (\"Run `bun run build` in tools/mcp, or `bun run gen:mcp` from the repo root\"), runs `bun run build` in tools/mcp believing the corpus was refreshed, and only discovers via a later `gen-mcp-corpus.test.ts` CI failure (or ships stale docs if that test isn't run) that the comment's first suggested command does nothing - the same class of bug commit e8b0dc1 says it fixed (\"several comments still claimed build regenerates the corpus\") but this instance was missed."
  }
]
```

dunxonu (commented):
Reviewed the diff. 7 findings.

### `scripts/render-review.ts:61`

A stray `[]` JSON code fence anywhere in the reviewer's message causes the review to be posted as an approval even when the model wrote `VERDICT: COMMENT` in its prose.

**How it fails:** `parseFindings` validates a parsed array with `Array.prototype.every`, which is vacuously true on `[]`, so any fenced ```json\n[]\n``` block in the message (e.g. an illustrative example, not the model's actual answer) is accepted as a real, empty findings list. `findingsIn` flattens across every fenced block, so if the model's genuine review is prose ending in `VERDICT: COMMENT` but also happens to include a `[]` fence, `findings` becomes `[]` (not `undefined`), the code takes the `else` branch (`verdict = findings.length === 0 ? 'approve' : 'comment'`), and posts `approve` on a PR the model actually flagged for changes.

### `scripts/render-review.ts:28`

`resultOf`'s `JSON.parse(raw)` has no try/catch, unlike `parseFindings` later in the same file.

**How it fails:** If `${{ steps.review.outputs.execution_file }}` is empty or truncated (e.g. the review CLI was killed before flushing), `JSON.parse` throws an uncaught `SyntaxError`, so the workflow step fails with a raw stack trace instead of the script's own designed diagnostic ('The reviewer produced no review. Not posting an empty one.'), which is the message this file was written to produce for exactly this case.

### `scripts/render-review.ts:48`

The fenced-block regex `/```(?:json)?\s*\n([\s\S]*?)\n\s*```/g` is lazy and truncates at the first inner triple-backtick preceded by a raw newline.

**How it fails:** A finding's `summary` or `failure_scenario` string embeds a multi-line code excerpt with an unescaped raw newline before a nested ``` fence (a known LLM formatting slip when quoting diff snippets); the outer match ends early, `JSON.parse` throws inside `parseFindings`, that block returns `undefined`, and if it was the only JSON block, the whole run falls through to the prose branch, posting a broken/partial review body instead of the rendered findings list.

### `CLAUDE.md:0`

The PR adds a capability to `tools/mcp` (and `tools/create-app`) without updating `examples/full`, violating Rule 4 ("a feature is not shipped until an example uses it"), and the PR body admits it outright.

**How it fails:** Per CLAUDE.md: "Adding or changing a capability in `packages/*` or `tools/*` includes updating `examples/full` in the same change." The PR description states verbatim: "Not done: no `examples/full` change, so Rule 4 is unsatisfied in the literal sense." The three new adoption tools and the shared `BOOT_RULES` are therefore untested end to end by the example ladder CI exercises, the exact failure mode Rule 4 exists to prevent (four prior features shipped broken this way for weeks).

### `tools/mcp/src/guide.ts:96`

`Guide.chapter()`/`candidates()` hand-roll the same case-insensitive substring matching that `Args.like`/`Args.eq` (tools/mcp/src/args.ts) already generalizes, and `Scaffold.features()` (tools/mcp/src/scaffold.ts) duplicates it a third time.

**How it fails:** A future change to the matching rule (e.g. trimming, accent-insensitivity, a different case-fold) has to be made in three places instead of one; CLAUDE.md Rule 2 ("one declaration, and it lives at the lowest common owner") is violated since `Args` was introduced in this same PR specifically to share this logic, but two of its own new sibling files don't use it.

### `scripts/gen-mcp-corpus.ts:592`

The doc comment above `rewriteLinks` describes an ordered-regex-replacement technique ("`../../` is replaced before `../`, or the second pattern eats the first half of the first") that the actual implementation does not use.

**How it fails:** The real code resolves paths structurally via `posix.normalize(posix.join(GUIDE_DIR, href))`, with no sequential regex replacement at all. A maintainer debugging a link-rewriting issue reads the comment, believes replacement order matters, and chases a regex-ordering bug that was never implemented instead of the actual `posix.join`/`normalize` logic.

### `tools/mcp/src/guide.ts:92`

`Guide.chapter()` refuses to resolve an ambiguous topic substring rather than fixing the underlying collision: `docs/guide/22-metrics.md` and `docs/guide/22-upgrading.md` genuinely share the leading number "22".

**How it fails:** A caller (or model) asking `dunx_guide({ topic: "22" })` gets no chapter, only a `candidates` list, requiring a second round trip - the ambiguity is pushed to every caller of this tool rather than removed at the source by renumbering the two colliding guide chapters.

dunxonu (commented):
Reviewed the diff. 2 findings.

### `scripts/render-review.ts:1038`

The fenced-block regex only recognizes a lowercase `json` language tag, so a findings block opened with ```JSON (or another casing/language annotation) is not treated as a fence.

**How it fails:** If the reviewer ever emits ` ```JSON\n[...]\n``` ` instead of the lowercase ` ```json ` the skill's own template shows, `blocksIn` falls back to treating the whole message as one block; `parseFindings` then only succeeds if the entire message (prose included) starts with `[`, so a real findings list preceded by any prose is silently discarded and the review posts as unparsed prose (or the wrong verdict) instead of the rendered findings.

### `scripts/gen-mcp-corpus.ts:598`

`rewriteLinks` assumes every relative guide link resolves to a path at or under `docs/guide` or the repo root; a link with enough leading `../` segments to normalize past the repo root produces a malformed absolute URL instead of falling back safely.

**How it fails:** If a future `docs/guide/*.md` chapter (or one moved into a subdirectory) contains a link like `../../../foo.md`, `posix.normalize(posix.join('docs/guide', href))` yields a path starting with `../` (e.g. `../foo.md`); this fails the chapter regex and is emitted as `${BLOB}/../foo.md`, a broken GitHub link shipped into the bundled corpus with no test catching it since no current chapter link overshoots this way.

scripts/post-review.ts
```diff
@@ -0,0 +1,293 @@
...
+  commentable: Commentable,
+): Review => {
+  const findings = findingsIn(result);
+  const sentinel = VERDICT.exec(result)?.[1];
+  const clean =
+    findings !== undefined &&
+    findings.length === 0 &&
+    proseOutside(result) === '';
+
+  const event: Review['event'] =
```
dunxonu, line 143: The review's APPROVE/COMMENT event is taken solely from the model's `VERDICT:` sentinel and never cross-checked against whether findings were actually posted.

**How it fails:** The reviewer model emits a real, non-empty findings list (which gets posted as inline `comments`) but ends its message with `VERDICT: APPROVE` (habit, template leftover, or a genuine slip). `buildReview` returns `event: 'APPROVE'` while `comments` still carries actionable defects, so the PR receives an approving review with attached problem comments - the same class of self-contradiction the two earlier fixes (vacuous `every`, unanchored fence) were meant to close, just moved one level up.

scripts/gen-mcp-corpus.ts
```diff
@@ -0,0 +1,201 @@
...
+ * the resolved path. It used to be ordered regex replacement, `../../` before
+ * `../` so the second pattern did not eat the first half of the first, which was
+ * correct only while every chapter sat at exactly `docs/guide/*.md`.
+ */
+const rewriteLinks = (markdown: string): string =>
+  markdown.replace(/\]\((\.[^)\s]+)/g, (_match, href: string) => {
+    // Resolved against the chapter's own directory rather than matched at a fixed
+    // depth. `../` and `../../` were counted by hand, which is correct only while
+    // every chapter sits exactly at `docs/guide/*.md`.
+    const target = posix.normalize(posix.join(GUIDE_DIR, href));
```
dunxonu: `rewriteLinks` resolves every chapter's relative links against a single hardcoded `GUIDE_DIR` constant, not the directory of the specific chapter being processed, despite the comment claiming resolution against "the chapter's own directory."

**How it fails:** If `docs/guide/` ever gains a subdirectory (e.g. `docs/guide/advanced/23-foo.md`), that chapter's `./bar.md` or `../01-introduction.md` links still resolve against `docs/guide` instead of `docs/guide/advanced`, producing a wrong `dunx://guide/` URI or wrong GitHub blob link silently - the class of bug this commit's message says it retired, one directory level up. `readGuide` also only globs `docs/guide/*.md` non-recursively, so nothing here is exercised by a subdirectory today.

scripts/gen-mcp-corpus.ts
```diff
@@ -0,0 +1,201 @@
...
+    const chapter = /^docs\/guide\/(\d+[a-z0-9-]*)\.md(#.*)?$/i.exec(target);
```
dunxonu: The chapter-link regex `^docs\/guide\/(\d+[a-z0-9-]*)\.md(#.*)?$` requires the target slug to start with digits, matching only today's `NN-name.md` convention.

**How it fails:** A future guide page without a numeric prefix (an appendix like `faq.md`) fails the regex, so any link to it falls through to the absolute `BLOB` URL branch instead of `dunx://guide/faq` - a client following that link leaves the bundled corpus, silently, with no test catching it since no such file exists yet.

scripts/gen-mcp-corpus.ts
```diff
@@ -0,0 +1,201 @@
...
+ * corpus. Everything else is a repo path the consumer does not have, so it becomes
+ * an absolute link.
+ *
+ * The href is resolved against the chapter's directory and the decision is made on
+ * the resolved path. It used to be ordered regex replacement, `../../` before
+ * `../` so the second pattern did not eat the first half of the first, which was
+ * correct only while every chapter sat at exactly `docs/guide/*.md`.
+ */
+const rewriteLinks = (markdown: string): string =>
+  markdown.replace(/\]\((\.[^)\s]+)/g, (_match, href: string) => {
```
dunxonu: `rewriteLinks`'s link-detection regex only matches relative links that literally start with `.` (`/\]\((\.[^)\s]+)/g`), so a bare relative link like `[Validation](06-validation.md)` (no `./` prefix - CommonMark-equivalent to `./06-validation.md`) is never rewritten.

**How it fails:** A contributor writes a same-directory chapter link without the `./` prefix; it ships into the published `@dunx/mcp` corpus as a literal `06-validation.md` path that resolves as neither a `dunx://guide/` resource nor a GitHub link. No current chapter uses this form, so `gen-mcp-corpus.test.ts`'s round-trip check does not catch it.

scripts/gen-mcp-corpus.ts
```diff
@@ -0,0 +1,201 @@
...
+    devDependencies: Object.keys(manifest.devDependencies ?? {}),
+    files,
+  };
+};
+
+/**
+ * Annotated with `@dunx/mcp`'s `BootRule` rather than `@dunx/create-app`'s, so the
+ * two shapes cannot drift without failing this file's typecheck.
+ */
+const readRules = (): readonly BootRule[] => BOOT_RULES;
```
dunxonu, line 165: `readRules(): readonly BootRule[] => BOOT_RULES` relies on structural typing between `@dunx/create-app`'s `BootRule` and `@dunx/mcp`'s restated `BootRule` to catch drift at typecheck time, but the assignment is a variable reference rather than an object literal, so TypeScript skips excess-property checking.

**How it fails:** Adding a new required field to create-app's `BootRule` (e.g. `severity: string`, populated on every entry) still type-checks against mcp's narrower interface because a wider object is assignable to a narrower one; the field silently rides through `JSON.stringify` into the generated corpus with nothing in `@dunx/mcp` aware it exists, contradicting the commit message's claim that the two shapes "cannot drift without failing the generator's own typecheck."

scripts/gen-mcp-corpus.ts
```diff
@@ -0,0 +1,201 @@
...
+ * A chapter link becomes its resource URI, so a client following one stays in the
+ * corpus. Everything else is a repo path the consumer does not have, so it becomes
+ * an absolute link.
+ *
+ * The href is resolved against the chapter's directory and the decision is made on
+ * the resolved path. It used to be ordered regex replacement, `../../` before
+ * `../` so the second pattern did not eat the first half of the first, which was
+ * correct only while every chapter sat at exactly `docs/guide/*.md`.
+ */
+const rewriteLinks = (markdown: string): string =>
```
dunxonu: `rewriteLinks` duplicates `internal/docs/scripts/content.ts`'s `rewriteHref`: both resolve a chapter's relative markdown link against its source directory and fall back to the same hardcoded GitHub blob base URL under a different constant name.

**How it fails:** The repo blob URL or path-normalization rule changes (e.g. a branch rename or a new traversal edge case) and gets fixed in one of the two implementations but not the other, so the docs site and the bundled MCP corpus render the same chapter's links differently.

tools/mcp/src/guide.ts
```diff
@@ -0,0 +1,166 @@
...
+   * Exact slug first, then the single chapter a substring matches. A caller that
+   * has read the index passes a slug; one that has not passes `validation`, and
+   * both land on the same chapter.
+   *
+   * A substring matching several is `undefined` rather than the first of them:
+   * `docs/guide/` holds both `22-metrics` and `22-upgrading`, so `topic: "22"`
+   * used to spend a whole chapter body on a coin flip. {@link candidates} is what
+   * the caller is given instead.
+   */
+  chapter(topic: string): GuideDoc | undefined {
```
dunxonu, line 92: `Guide.chapter()` and `Guide.candidates()` each independently re-scan the full corpus with the identical substring filter, and `adopt.ts`'s `dunx_guide` handler calls both back-to-back for a single lookup.

**How it fails:** Every `dunx_guide({ topic })` call does two-to-three full scans of the ~25-chapter corpus (once inside `chapter()`, again via `candidates()` for the not-found or otherwise-matches branch in `tools/mcp/src/adopt.ts:75` and `:89`) instead of one, wasted work on every tool call that only grows as the guide does.

scripts/gen-mcp-corpus.ts
```diff
@@ -0,0 +1,201 @@
...
+ */
+const literal = (value: unknown): string =>
+  JSON.stringify(JSON.stringify(value));
+
+/**
+ * Exported so `gen-mcp-corpus.test.ts` can re-render and compare, which is what
+ * catches a committed corpus that a guide edit left behind. `gen:readme --check`
+ * exists because nothing ran it and it silently broke; this is the same guard.
+ */
+export const renderCorpus = async (): Promise<string> => {
```
dunxonu, line 192: `readGuide` and `readStarter` read each source file sequentially with `await` inside `for` loops, and the generator awaits them one after another, instead of overlapping independent disk reads with `Promise.all`.

**How it fails:** Regenerating the corpus (`bun run gen:mcp`, run in CI) serializes roughly 25+ independent file reads across `docs/guide`, `examples/minimal/src`, and config files that have no data dependency on each other, adding avoidable latency to every corpus regeneration.

dunxonu (commented):
**Actionable comments posted: 8**

scripts/gen-mcp-corpus.ts
```diff
@@ -0,0 +1,225 @@
...
+    (whole: string, href: string) => {
+      // Against the directory of the chapter being rewritten, not a constant. A
+      // chapter one level down would otherwise resolve its links as if it sat
+      // beside its siblings.
+      const target = posix.normalize(posix.join(posix.dirname(from), href));
+      const chapter = /^docs\/guide\/([^/]+)\.md(#.*)?$/i.exec(target);
+      if (chapter) return `](dunx://guide/${chapter[1]}${chapter[2] ?? ''}`;
+      // Outside the corpus and outside the repo layout it knows: leave it be
+      // rather than inventing a blob URL for it.
+      return target.startsWith('..') ? whole : `](${BLOB}/${target}`;
```
dunxonu, line 65: rewriteLinks turns any non-chapter relative link into a GitHub /blob/main/ URL even when the target is a directory, and GitHub's /blob/ scheme only renders files.

**How it fails:** docs/guide/02-first-steps.md lines 662-665 link `../../examples/minimal`, `../../examples/databases`, `../../examples/testing` and `../../examples/full` (all directories). Confirmed in the committed tools/mcp/src/generated.ts, which now contains `.../blob/main/examples/minimal)`, `.../blob/main/examples/databases)`, etc. Following any of these from dunx_guide's output 404s on GitHub; a `/tree/main/` URL is what a directory needs.

scripts/gen-mcp-corpus.ts
```diff
@@ -0,0 +1,225 @@
...
+ * an absolute link.
+ *
+ * The href is resolved against the chapter's directory and the decision is made on
+ * the resolved path. It used to be ordered regex replacement, `../../` before
+ * `../` so the second pattern did not eat the first half of the first, which was
+ * correct only while every chapter sat at exactly `docs/guide/*.md`.
+ */
+export const rewriteLinks = (markdown: string, from: string): string =>
+  // Any href that is not absolute, not a fragment and not root-relative, which
+  // covers `./x.md`, `../../x` and a bare `x.md`. Matching only a leading `.`
```
dunxonu, line 52: The link-rewrite regex matches `](...)` anywhere in the markdown, including inside fenced code blocks, so a chapter that shows markdown link syntax as a literal code sample would have that sample silently rewritten.

**How it fails:** A future guide chapter with a code fence containing `[text](./foo.md)` as an illustrative snippet (e.g. documenting markdown syntax itself) would get that fence corrupted into a `dunx://guide/...` or blob URL inside the code sample. No current chapter triggers it (checked docs/guide/*.md for a matching pattern inside fences and found none), so this is a latent gap rather than an active failure.

scripts/gen-mcp-corpus.ts
```diff
@@ -0,0 +1,225 @@
...
+  // left a bare sibling link pointing at a path the consumer does not have.
+  markdown.replace(
+    /\]\((?!https?:|mailto:|#|\/)([^)\s]+)/g,
+    (whole: string, href: string) => {
+      // Against the directory of the chapter being rewritten, not a constant. A
+      // chapter one level down would otherwise resolve its links as if it sat
+      // beside its siblings.
+      const target = posix.normalize(posix.join(posix.dirname(from), href));
+      const chapter = /^docs\/guide\/([^/]+)\.md(#.*)?$/i.exec(target);
```
dunxonu, line 61: The chapter-detection regex hardcodes the literal `docs/guide` instead of interpolating the `GUIDE_DIR` constant declared at line 37, so the two can silently diverge.

**How it fails:** If GUIDE_DIR is ever changed (chapter directory renamed/moved), line 79's `source` path updates but the regex at line 61 keeps matching the old literal, so every chapter link stops resolving to a `dunx://guide/` resource and instead falls through to a blob URL, with no compile or test error pointing at the real cause.

tools/mcp/src/scaffold.ts
```diff
@@ -0,0 +1,123 @@
...
+ * what an agent about to write one needs; `bunx @dunx/create-app` is what writes.
+ */
+export class Scaffold {
+  constructor(
+    private readonly catalogue: readonly ScaffoldFeature[],
+    private readonly minimal: Starter,
+  ) {}
+
+  features(name?: string): readonly ScaffoldFeature[] {
+    if (name === undefined) return this.catalogue;
```
dunxonu, line 65: Scaffold.features re-implements the case-insensitive substring match that Args.like already provides, as a second copy of the same logic.

**How it fails:** Not a crash: it is a second, independent case-insensitive substring matcher (`tools/mcp/src/args.ts`'s `like`) that must be kept behaviorally identical by hand; a future tweak to one (e.g. trimming, locale-aware casing) will not reach the other without someone noticing both exist.

dunxonu (commented):

PR #67, opened 2026-09-08
fix(infra): give bullmq's duplicate connections a teardown owner

### LINKED ISSUE #65: Manage BullMQ duplicate client lifecycle during queue shutdown
## Summary

`QueueConnection` tracks original Redis clients, but BullMQ can create duplicate clients for blocking work. This duplicate lifecycle is not currently managed by `QueueConnection`.

## Rationale

Measured against a live broker:

- `QueueEvents` calls `duplicate()` once.
- `Worker` calls `duplicate()` once.
- `Queue` does not call `duplicate()`.

This behavior pre-dates `JobEvents`. It can leave a duplicate connection outside `QueueConnection.#open` during shutdown.

A direct force-close change is incomplete. Through `boundClientClass`, `duplicate()` can return a `Promise`. The resulting adapter can have `adapter.raw === undefined`. The existing shutdown sequence uses both `adapter.disconnect()` and `raw.close()`. The latter is required when an unconnected client would otherwise keep the process alive.

This work intersects with `internal/notes/roadmap/queue-shutdown-sigterm.md`. Handle it as a dedicated shutdown change.

## Affected areas

- `packages/infra/src/queue/connection.ts`
- `packages/infra/src/queue/events.ts`
- Worker connection construction in `packages/infra/src/queue`
- `internal/notes/roadmap/queue-shutdown-sigterm.md`

## Required changes

1. Define ownership and teardown behavior for BullMQ duplicate clients created from `QueueConnection.client()`.
2. Support the asynchronous `duplicate()` behavior introduced by `boundClientClass`.
3. Preserve cleanup for clients that did not finish connecting.
4. Verify shutdown behavior for both `Worker` and `QueueEvents`.
5. Keep shutdown bounded and avoid introducing or worsening the SIGTERM hang.

## Acceptance criteria

- Every duplicate client created by BullMQ from a `QueueConnection` client has defined teardown ownership.
- Shutdown cleans up duplicate clients when `close()` does not settle promptly.
- The design handles a duplicate whose raw client is not immediately available.
- Tests do not depend on broker connection timing.
- The SIGTERM roadmap note is reviewed and updated if the behavior changes.

## Backlinks

- PR: https://github.com/petarzarkov/dunx/pull/63
- Review discussion: https://github.com/petarzarkov/dunx/pull/63#discussion_r3955114752

Requested by @petarzarkov.

petarzarkov:
Closes #65.

## What changed

`QueueConnection` recorded only the clients `client()` built. bullmq calls
`duplicate()` for any connection it may block on, so that socket belonged to
nobody: `onShutdown` never reached it, and `open` under-reported by one per
blocking object.

Every adapter is now recorded in `#handleErrors` rather than in `client()`, which
covers duplicates of duplicates as well, and the socket is read off the adapter at
teardown instead of captured when the client is handed out.

## Why that shape

A force-close of a captured `raw` does not work, which the issue records. A
duplicate is constructed with **no `raw` at all**: bullmq builds one from a
`rawFactory` on first connect, which is how it stopped sending duplicates to the
default server ([bullmq#4582](https://github.com/taskforcesh/bullmq/issues/4582)).
Reading it at teardown handles that with no promise to await inside a synchronous
wrapper. `disconnect()` guards on `this.raw`, so a duplicate that never connected
tears down with nothing to close.

`onShutdown` drains the array before walking it, because `disconnect()` can
schedule a reconnect and a reconnect duplicates.

## Measurements

`duplicate()` calls on the client handed over, live broker:

| bullmq object | calls |
| ------------- | ----- |
| `Queue`       | 0     |
| `Worker`      | 1     |
| `QueueEvents` | 1     |

So this predates `JobEvents` by as long as `WorkerFactory` has existed. Sockets
held by one `QueueConnection`:

| object        | before shutdown | after |
| ------------- | --------------- | ----- |
| `Worker`      | 2               | 0     |
| `QueueEvents` | 2               | 0     |

Both read 1 before this change, and the second socket was left open.

**Leak B is untouched.** A `Worker` on an unreachable broker still holds the loop
after `close()`, and `ShutdownHooks`' forced exit still stands. The numbers above
are sockets released, not a process that now exits. Defect D in
`internal/notes/roadmap/queue-shutdown-sigterm.md` records that boundary.

## Checks

- [x] `bun run build`
- [x] `bun run lint:check`
- [x] `bun run format:check`
- [x] `bun run typecheck`
- [x] `bun run test:cov`
- [x] Tests cover the change. Two of the six fail against the previous implementation.
- [x] Conventional commit messages.
- [x] Docs updated: the SIGTERM roadmap note gains defect D.

`bun run ci` is 13/13 with valkey, postgres and MinIO up.

Tests depend on no broker, per the issue's acceptance criteria: a duplicate
reports its own `disconnect`, and whether it ever connected is irrelevant to who
owns it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed queue shutdown so duplicated and derived connections are properly closed.
  - Corrected connection counts to include active sockets created for workers and event listeners.
  - Improved shutdown reliability when connections are created late or when one connection fails to disconnect.
  - Prevented unconnected duplicate connections from being counted as open.
  - Ensured repeated shutdown attempts complete safely without reopening or leaking connections.

### DIFF
--- a/internal/notes/roadmap/queue-shutdown-sigterm.md
+++ b/internal/notes/roadmap/queue-shutdown-sigterm.md
@@ -80,11 +80,12 @@ bisecting the stack a layer at a time found a leak in `Bun.RedisClient` on its o
 and a second, separate one in bullmq's Bun adapter. Neither is reachable from
 userland. All three have a minimal reproduction, ready to file.
 
-| #                                                                          | Symptom                                                                       | Layer  |
-| -------------------------------------------------------------------------- | ----------------------------------------------------------------------------- | ------ |
-| [A](#leak-a---bun-a-connect-that-never-completes-outlives-close)           | ~~a pending connect outlives `close()`~~ **fixed in Bun 1.4**                 | Bun    |
-| [B](#leak-b---bullmq-disconnect-cannot-cancel-its-own-reconnect)           | a `Worker` on an unreachable server holds the loop after `close()`            | bullmq |
-| [C](#defect-c---no-connection-is-ever-named-so-getworkers-is-always-empty) | ~~`getWorkers()` always `[]`~~ **fixed - was dunx's own `duplicate` wrapper** | dunx   |
+| #                                                                          | Symptom                                                                               | Layer  |
+| -------------------------------------------------------------------------- | ------------------------------------------------------------------------------------- | ------ |
+| [A](#leak-a---bun-a-connect-that-never-completes-outlives-close)           | ~~a pending connect outlives `close()`~~ **fixed in Bun 1.4**                         | Bun    |
+| [B](#leak-b---bullmq-disconnect-cannot-cancel-its-own-reconnect)           | a `Worker` on an unreachable server holds the loop after `close()`                    | bullmq |
+| [C](#defect-c---no-connection-is-ever-named-so-getworkers-is-always-empty) | ~~`getWorkers()` always `[]`~~ **fixed - was dunx's own `duplicate` wrapper**         | dunx   |
+| [D](#defect-d---a-duplicate-connection-had-no-teardown-owner)              | ~~a duplicate socket outlived shutdown~~ **fixed - dunx's `duplicate` wrapper again** | dunx   |
 
 ## The measurement that separated them
 
@@ -187,6 +188,47 @@ A userland escape hatch exists but is a cast through `unknown` into a field bull
 does not export - `closing = true` plus `clearTimeout(reconnectTimer)`. That is a
 fork by another name and it would break on a patch release, so it is **not** in
 `@dunx/infra/queue`. Fix it upstream, or wait for it.
+
+## Defect D - a duplicate connection had no teardown owner
+
+> **FIXED, and dunx's own, in the same wrapper as C.** `QueueConnection` recorded
+> only the clients `client()` built. bullmq calls `duplicate()` for any connection
+> it may block on, so that socket belonged to nobody: `onShutdown` never reached
+> it, and `open` under-reported by one per blocking object.
+
+Measured against a live broker, counting `duplicate()` calls on the client handed
+over:
+
+| bullmq object | `duplicate()` calls |
+| ------------- | ------------------- |
+| `Queue`       | 0                   |
+| `Worker`      | 1                   |
+| `QueueEvents` | 1                   |
+
+So this predates `JobEvents` by as long as `WorkerFactory` has existed. Sockets
+held by one `QueueConnection`, before and after `onShutdown`:
+
+| object        | before | after |
+| ------------- | ------ | ----- |
+| `Worker`      | 2      | 0     |
+| `QueueEvents` | 2      | 0     |
+
+Both read 1 before the fix, and the second socket was left open.
+
+The fix is not a force-close of a captured `raw`, which is what the review first
+suggested. A duplicate is constructed with **no `raw` at all**: bullmq builds one
+from a `rawFactory` on first connect, which is how it stopped sending duplicates
+to the default server ([#4582](https://github.com/taskforcesh/bullmq/issues/4582)).
+So the socket is read off the adapter at teardown instead of captured when the
+client is handed out - `undefined` immediately, a `Bun.RedisClient` once
+connected. `disconnect()` is raw-safe, so a duplicate that never connected tears
+down without a socket to close.
+
+**This does not touch leak B.** A `Worker` on an unreachable broker still holds
+the loop after `close()`, `ShutdownHooks`' forced exit still stands, and the
+numbers above are sockets released rather than a process that now exits. What
+changed is ownership: every adapter a `QueueConnection` hands out or bullmq
+derives from one is now disconnected and closed by it.
 
 ## Defect C - no connection is ever named, so `getWorkers()` is always empty
 
--- a/internal/notes/roadmap/queue-shutdown-sigterm.md
+++ b/internal/notes/roadmap/queue-shutdown-sigterm.md
@@ -226,9 +226,16 @@ down without a socket to close.
 
 **This does not touch leak B.** A `Worker` on an unreachable broker still holds
 the loop after `close()`, `ShutdownHooks`' forced exit still stands, and the
-numbers above are sockets released rather than a process that now exits. What
-changed is ownership: every adapter a `QueueConnection` hands out or bullmq
-derives from one is now disconnected and closed by it.
+numbers above are sockets released rather than a process that now exits.
+
+What changed is ownership: every adapter a `QueueConnection` hands out or bullmq
+derives from one is disconnected and closed by it, within a bounded teardown.
+`onShutdown` makes at most `TEARDOWN_PASSES` passes, taking each pass's adapters
+out before walking them, so one derived during a pass belongs to the next. The
+bound is what iterating the live array lacked: an adapter that derived another on
+every `disconnect()` would have been walked forever, and a livelocked shutdown is
+worse than a leaked socket. Anything still arriving after the last pass is logged
+and left to the runtime, which is the one case ownership does not cover.
 
 ## Defect C - no connection is ever named, so `getWorkers()` is always empty
 
--- a/packages/infra/src/queue/connection.test.ts
+++ b/packages/infra/src/queue/connection.test.ts
@@ -102,3 +102,93 @@ describe('the raw client bullmq reconnects with', () => {
     source.onShutdown();
   });
 });
+
+/**
+ * bullmq calls `duplicate()` for any connection it may block on - `Worker` and
+ * `QueueEvents` each once, `Queue` never - and `#open` used to hold only the
+ * clients `client()` built, so that socket had no teardown owner.
+ *
+ * Nothing here waits on a broker: a duplicate reports its own `disconnect`, and
+ * whether it ever connected is irrelevant to who owns it.
+ */
+describe('duplicate ownership', () => {
+  /** Replaces `disconnect` with a recorder, returning what it recorded. */
+  const watchDisconnect = (client: unknown): { calls: number } => {
+    const spy = { calls: 0 };
+    const target = client as { disconnect: () => void };
+    const inner = target.disconnect.bind(target);
+    target.disconnect = (): void => {
+      spy.calls += 1;
+      inner();
+    };
+    return spy;
+  };
+
+  it('tears down a duplicate, which used to belong to nobody', () => {
+    const source = connection();
+    const client = source.client() as unknown as {
+      duplicate: (options: unknown) => unknown;
+    };
+    const copy = client.duplicate({ connectionName: 'bull:cXVldWU=' });
+
+    const watched = watchDisconnect(copy);
+    source.onShutdown();
+
+    expect(watched.calls).toBe(1);
+  });
+
+  it('tears down a duplicate of a duplicate', () => {
+    const source = connection();
+    const first = source.client() as unknown as {
+      duplicate: (options: unknown) => { duplicate: (o: unknown) => unknown };
+    };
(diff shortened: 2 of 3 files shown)


dunxonu (commented):

packages/infra/src/queue/connection.ts
```diff
@@ -124,10 +146,13 @@ export class QueueConnection implements OnShutdown {
    * internal/notes/roadmap/queue-shutdown-sigterm.md.
    */
   onShutdown(): void {
-    for (const { adapter, raw } of this.#open) {
+    // Drained first: `disconnect()` can schedule a reconnect, and a reconnect
+    // duplicates, so iterating the live array could append to what it is walking.
+    for (const adapter of this.#open.splice(0)) {
```
dunxonu: onShutdown()'s teardown loop has no try/catch, so one adapter throwing on disconnect() aborts the loop and orphans every adapter after it in the snapshot.

**How it fails:** this.#open.splice(0) removes all tracked adapters from #open up front. If adapter[0].disconnect() throws (e.g. bullmq's disconnect() is invoked on a socket already in a torn-down or errored state during an abrupt shutdown), the for-of loop stops immediately; adapters[1..n] are neither disconnected nor closed, and because they were already spliced out of #open, no future onShutdown() call will ever retry them - their sockets leak for the life of the process.

packages/infra/src/queue/connection.ts
```diff
@@ -124,10 +146,13 @@ export class QueueConnection implements OnShutdown {
    * internal/notes/roadmap/queue-shutdown-sigterm.md.
    */
   onShutdown(): void {
-    for (const { adapter, raw } of this.#open) {
+    // Drained first: `disconnect()` can schedule a reconnect, and a reconnect
```
dunxonu: Draining #open via splice(0) before iterating (instead of after, as the old code did) can miss an adapter that gets duplicated mid-teardown.

**How it fails:** The old code iterated the live #open array by reference and cleared it only after the loop finished, so a duplicate() call triggered synchronously by disconnect() (e.g. a scheduled reconnect) during the loop was still visited before the loop ended. The new code snapshots into a local array via splice(0) first and empties #open immediately; if disconnect() on an early adapter synchronously triggers a reconnect that calls the wrapped duplicate() (per the file's own comment: 'disconnect() can schedule a reconnect, and a reconnect duplicates'), the newly pushed adapter lands in the now-fresh #open but is not part of the array this onShutdown() call is iterating, so its socket is never disconnected or closed by this shutdown pass.

packages/infra/src/queue/connection.ts
```diff
@@ -52,7 +52,8 @@ export class QueueConnection implements OnShutdown {
   readonly #options: QueueOptions;
   readonly #logger: Logger;
   readonly #client: new (url?: string) => Bun.RedisClient;
-  readonly #open: { adapter: IRedisClient; raw: Bun.RedisClient }[] = [];
+  /** Every adapter this connection handed out or bullmq derived from one. */
+  readonly #open: IRedisClient[] = [];
```
dunxonu, line 70: QueueConnection's #open array and worker.ts's #workers array independently reimplement the same 'track handed-out resources, drain and tear down on shutdown' pattern with no shared abstraction.

**How it fails:** connection.ts drains via splice(0) plus disconnect()/close(), guarded against a mid-loop append; worker.ts's #workers pool drains via .length = 0 plus Promise.all/allSettled close, without the same reconnect-during-teardown protection. A correctness fix needed by one resource pool (like the splice-before-iterate ordering added here) has to be independently rediscovered and re-applied to the other, since there is no shared teardown-registry helper in @dunx/infra or @dunx/core to fix once.

packages/infra/src/queue/connection.ts
petarzarkov: Taken in fcccad6. Each adapter's disconnect and close are guarded separately now, so neither aborts the other or anything after it.

Your point about the splice making it worse is the part that matters: before this fix the adapters were left in `#open` until after the loop, so a throw at least left them for a retry. Snapshotting removed even that, which I did not think through.

Test asserts an adapter after a throwing one is still reached, and fails against a3d7dcf.

packages/infra/src/queue/connection.ts
petarzarkov: Right about the behaviour change, and my comment justifying it was simply wrong - worth stating plainly since it claimed a mechanism.

It said `disconnect()` can schedule a reconnect and a reconnect duplicates. Only `disconnect(true)` reaches `_scheduleReconnect`; `onShutdown` calls it with no argument, which takes the `else` branch: `closing`/`closed`, `_closeRaw()`, then `emit('close')` and `emit('end')`. So teardown cannot itself derive a duplicate, and I had asserted that it could without reading it.

A listener on those events still can, which is your case, and iterating the live array would have visited it where the snapshot dropped it.

fcccad6 does neither: the drain is a bounded set of passes, so an adapter arriving during one belongs to the next. The bound is the thing the live-array loop lacked - an adapter that duplicated on every disconnect would have been walked forever, and a livelocked shutdown is worse than a leaked socket. Anything still arriving after four passes is logged and left to the runtime.

Two tests: one asserts an adapter created during teardown is torn down, and fails against a3d7dcf; one arms every duplicate to derive another and asserts shutdown still returns.

packages/infra/src/queue/connection.ts
petarzarkov: Not taking this one, and I looked at both pools rather than answering from the shape.

They are not the same teardown. `#open` is synchronous, two steps, and needs a bound because an adapter can derive another. `#workers` is asynchronous and has two different policies: `stop()` uses `Promise.all(worker.close())`, and the failed-start path uses `Promise.allSettled` with `closeWithin(worker, DRAIN_ON_FAILED_START_MS)` for ready workers and `close(true)` for the rest, because a forced close on a worker mid-job would have the stalled recovery run a handler's side effects twice.

The specific fix here also does not transfer. `Promise.all` starts every close concurrently, so one rejection does not stop the others from running - which is exactly the failure my sequential loop had and yours does not. A shared registry would have to be generic over sync and async teardown and over three closing policies, for two call sites, which is the speculative abstraction CLAUDE.md rules out.

There is a real thing inside your comment though, and it is not the missing abstraction: `worker.ts:252`'s `#workers.length = 0` sits after the `await`, so a rejected close leaves the array populated, and `#stopping` memoises the rejected promise so a second `stop()` returns the same rejection without retrying. Separate from this issue and worth its own look rather than folding it in here.

dunxonu: Agreed, dropping this one. `#open` and `#workers` don't share a teardown shape: one is a synchronous two-step drain guarded against a mid-loop append, the other is async with three different closing policies depending on worker readiness. Generalizing a shared registry over that for two call sites is exactly the speculative abstraction [CLAUDE.md](https://github.com/petarzarkov/dunx/blob/main/CLAUDE.md#do-not) rules out.

Confirmed the `worker.ts:252` point separately: `Promise.all` on line 251 rejects on the first failed `close()`, so `this.#workers.length = 0` is never reached, and `#stopping` memoises that rejection, so a retried `stop()` just replays it instead of trying again. Worth its own fix, as you said.

dunxonu (commented):
Confirmed the core teardown logic is correct. Given the thorough manual review plus verification, no high-confidence correctness bugs survive in this diff.

```json
[]
```

The PR (`fix(infra): let no adapter's teardown abandon the ones after it`, plus the earlier `give bullmq's duplicate connections a teardown owner`) is a small, self-contained, well-tested change to `packages/infra/src/queue/connection.ts`. I read every hunk, the enclosing class, traced callers of `QueueConnection`/`.open`, and had an independent agent verify the trickiest part (the splice/push interaction between `onShutdown`'s multi-pass loop and `#handleErrors`' synchronous push during teardown). No correctness, reuse, simplification, efficiency, altitude, or CLAUDE.md convention issues survived verification.

One pre-existing (not introduced by this diff) minor documentation staleness worth a follow-up but not rising to a review finding: the `onShutdown()` docstring at `packages/infra/src/queue/connection.ts:153` still says "last in reverse-order teardown," which was already inaccurate before this PR (the loop was forward-order both before and after this change) and this PR's rewrite left it untouched.

packages/infra/src/queue/connection.test.ts
```diff
@@ -102,3 +102,182 @@ describe('the raw client bullmq reconnects with', () => {
...
+    // `disconnect()` with no argument cannot reconnect, but a `close`/`end`
+    // listener could duplicate, and that adapter still needs an owner.
+    let late: { disconnect: () => void } | undefined;
+    const inner = first.disconnect.bind(first);
+    first.disconnect = (): void => {
+      late ??= client.duplicate({});
+      inner();
+    };
+
+    source.onShutdown();
```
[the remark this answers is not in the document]
petarzarkov: Taken in 866780e, and this was the better catch of the two: the assertion inferred teardown from a side effect rather than asserting it. A regression dropping the late adapter from `#open` without tearing it down leaves the second-shutdown count at zero as well, so the test would have passed.

The recorder is installed as the adapter is created now, and the assertion is one disconnect from the shutdown that derived it.

Verified against both regressions, because the weaker form only caught one:

- a single pass gives `Expected: 1, Received: 0`
- clearing `#open` after the pass without teardown, which is your scenario, fails the same way

Worth noting my first attempt at that verification was wrong: the formatter had reflowed the `for` across three lines, so the substitution I used to reintroduce the single-pass regression silently did not match and I was testing the unmodified file, which reported green. Redone with an assert on the substitution.

internal/notes/roadmap/queue-shutdown-sigterm.md
```diff
@@ -188,6 +189,47 @@ does not export - `closing = true` plus `clearTimeout(reconnectTimer)`. That is
...
+So the socket is read off the adapter at teardown instead of captured when the
+client is handed out - `undefined` immediately, a `Bun.RedisClient` once
+connected. `disconnect()` is raw-safe, so a duplicate that never connected tears
+down without a socket to close.
+
+**This does not touch leak B.** A `Worker` on an unreachable broker still holds
+the loop after `close()`, `ShutdownHooks`' forced exit still stands, and the
+numbers above are sockets released rather than a process that now exits. What
+changed is ownership: every adapter a `QueueConnection` hands out or bullmq
+derives from one is now disconnected and closed by it.
```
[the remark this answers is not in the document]
petarzarkov: Correct, and it was doc drift I introduced. That sentence was written in the first commit, before the bound existed, and I did not revisit it when the second commit added one - so it claimed total ownership in the same section that documents the bounded case.

866780e states the bounded teardown: at most `TEARDOWN_PASSES` passes, each pass taking its adapters out before walking them so one derived during a pass belongs to the next, and anything still arriving after the last pass logged and left to the runtime, called out as the one case ownership does not cover.

dunxonu (commented):
Both files well under the 500-line cap. No convention violations found (no em/en dashes, no CommonJS, no `any`, no enums introduced).

Based on thorough manual tracing of the diff — including the pass-bounded teardown loop, the duplicate-ownership wrapping in `#handleErrors`, the `#socketOf` accessor, and cross-referencing against the existing test suite and the `queue-shutdown-sigterm.md` roadmap doc — the change is internally consistent, the one edge case worth scrutiny (adapters still arriving after `TEARDOWN_PASSES` are abandoned without a final disconnect/close attempt) is explicitly documented and deliberately traded off against livelock risk, and is covered by a test (`stops rather than livelocking on an adapter that always duplicates`). No caller of `QueueConnection`, `.open`, or `onShutdown()` is broken by the shape change from `{adapter, raw}[]` to `IRedisClient[]`.

```json
[]
```

PR #68, opened 2026-09-08
fix(mcp): the five gaps a clean-room agent hit

petarzarkov:
Findings from driving the published `@dunx/mcp@3.5.0` over stdio from an empty directory, using only what it answered. The app it produced works; these are the five places its own output was not enough.

| Gap | Fix |
|---|---|
| The starter shipped no `package.json` - and `"type": "module"` absent breaks every relative import silently | The corpus renders one: names and scripts from `examples/minimal`, versions from the answering release |
| `@types/bun` in neither list, while the base tsconfig declares `types: ["bun"]` - first `tsc --noEmit` was TS2688 | `DEV_TOOLCHAIN`, exported from `@dunx/create-app`, which `manifest()` already spread into every generated app |
| `dunx_guide` search sold itself as the call for a question that names no chapter, and answered one with nothing | Every search carries `suggested`: the query's words, scored per chapter |
| `dunx_scaffold`'s `feature` matched the name only - `queue` and `redis` both found nothing | Name **or** summary |
| Two chapters numbered 22 | `23-upgrading.md`, plus a corpus test holding the numbering unique |

On the search scoring: words are compared five characters at a time, because `validate` and `Validation` share no substring in either direction and that pair is the query this exists to answer. A title hit is worth 8 down to 1 for a body mention, and a chapter mentioning the word only in its body does not qualify at all - every chapter says `route` somewhere, so scoring bodies alone ranked the whole guide against any query. Right chapter first in eight of ten questions an agent would type; both misses are vocabulary (`throttle` against "rate limit").

`Guide.chapter` already refused to guess between the two chapter 22s, so that one is the other half rather than the fix.

Verified by writing the payload into an empty directory and nothing else: `bun install`, `tsc --noEmit`, `bun test` and a booted route all pass. `bun run ci` green, 13/13, with valkey, postgres and MinIO up.

## Summary by CodeRabbit

- **New Features**
  - Guide searches now provide ranked chapter suggestions when literal searches find no direct matches.
  - Scaffold feature searches now match feature names and summaries.
  - Generated starter projects now include an installable package manifest and required development tooling.

- **Documentation**
  - Added an upgrade guide covering recent behavior changes and migration options.
  - Expanded MCP tool documentation with search, suggestions, feature matching, and starter-project details.
  - Updated roadmap and configuration links to the new upgrade guide.


dunxonu (commented):
Clean, nothing to remove.

Based on a thorough review of the diff (`tools/mcp/src/guide.ts`, `scaffold.ts`, `adopt.ts`, `scripts/gen-mcp-corpus.ts`, `tools/create-app/src/generate.ts`, and the associated tests/docs), I traced the new `Guide.suggest()` scoring logic, the `chapter()`/`candidates()` disambiguation fix (renaming `22-upgrading.md` → `23-upgrading.md`), the new starter `package.json` generation (`starterManifest`/`pin`/`DEV_TOOLCHAIN`), and `Scaffold.features()`'s name-or-summary matching, against their call sites, tests, and the CLAUDE.md conventions (lockstep versioning, exact-version rules, Rule 4 examples). All cross-checked correctly: no stale `22-upgrading` references remain, lockstep versioning across all `packages/*`/`tools/*` at `3.5.0` makes the `pin()` version substitution correct, no other caller of `Guide.search`/`MINIMAL` breaks on the additive `suggested` field, and the documented "throttle vs rate limit" miss is an honest, verified limitation rather than a bug.

No confirmed or plausible correctness, cleanup, or convention defects survived review.

```json
[]
```

PR #69, opened 2026-09-08
fix(mcp): resolve the starter's version when it is served, not when it is generated

petarzarkov:
**This blocks the 3.5.1 release.** `publish-guard` caught it while cutting it, and #68 is where I introduced it.

The starter manifest added in #68 wrote this checkout's version into `tools/mcp/src/generated.ts`, which is committed. `scripts/version.ts` bumps every manifest **after** that, so 3.5.1 would have:

1. published a `dunx_start` payload telling an agent to install **3.5.0** of core, http, transform and testing, and
2. reddened main on the next push, against the corpus drift test and the lockstep assertion at `scripts/gen-mcp-corpus.test.ts`.

Reproduced by setting `tools/mcp/package.json` to `3.5.1` and running the corpus suite: two failures. It had never fired before because at 3.5.0 the corpus carried no starter manifest to go stale, so 3.4.1 → 3.5.0 had no version to lose.

Building first is not the fix: the release job runs `bun run ci build` *before* `bun run version` (`ci.yml:271` then `:279`), so a version inlined at build time is a release behind too.

So the corpus stores `VERSION_PLACEHOLDER` and `Scaffold.starter()` resolves it from `ownVersion()` — one `readFileSync` of the package's own manifest, which is the one npm publishes already bumped. That is the rule `@dunx/create-app` has always used for its templates, and a corpus test now holds the two spellings to each other since neither package can import the other's.

`ownVersion()` replaces the private reader `cli.ts` already had, so `--version`, `serverInfo` and the starter agree by construction. The constructor parameter is **defaulted, not required**: `Scaffold` is exported from `index.ts`, and a third argument would be a breaking change on a patch.

Verified against the built `dist/`: bumping the manifest with **no rebuild and no regeneration** changes what the server serves, 3.5.0 → 3.5.1, which is exactly what the release job does. `bun run ci` green 13/13 with valkey, postgres and MinIO up.

## Summary by CodeRabbit

* **New Features**
  * Starter projects now receive the correct package version automatically when generated.
  * Added support for resolving version placeholders in starter files.
  * Exposed the version placeholder for package consumers.
  * The MCP server’s version information and CLI output now consistently reflect the package version.

* **Documentation**
  * Updated architecture documentation to explain starter version resolution.


tools/mcp/src/own-version.ts
```diff
@@ -0,0 +1,30 @@
...
+ * release job builds **before** `scripts/version.ts` bumps the manifests, so an
+ * inlined version is the previous release's. A read resolves against
+ * `dist/../package.json` in the published package and `src/../package.json` in a
+ * checkout, and the tarball npm publishes carries the bumped manifest.
+ *
+ * Synchronous, and a function rather than a class, because it is one file read
+ * with nothing to configure. Memoized: `dunx_scaffold` can be called repeatedly
+ * in one session and the answer cannot change while the process lives.
+ */
+let cached: string | undefined;
```
dunxonu, line 16: own-version.ts's module-level `let cached` gives `ownVersion` state that CLAUDE.md's Rule 3 says makes it a class, not a function.

**How it fails:** CLAUDE.md Rule 3 states "Anything with state, configuration or a lifetime is a class" and that the narrow function exception is only for a "pure, stateless, argument-in-value-out" function, ending with "The moment one grows a field, it is a class." `ownVersion` (line 18) is not stateless: it reads and memoizes into the module-level `cached` variable (line 16), which is exactly the growth the rule says converts it to a class. The docstring even argues against the rule ("a function rather than a class, because it is one file read with nothing to configure. Memoized") while describing a memoized field, so the code and the rule are in direct conflict.

dunxonu (commented):

PR #70, opened 2026-09-09
feat(bench): io scenario across all 20 subjects, memory and CPU, and a driver harness

petarzarkov:
Four additions to `internal/bench`, measured on the 5950X with Bun 1.4.2 and Node
24.21.0, and committed as `results/latest.json` and `results/drivers.json`.

- **`bun run smoke`** - every subject once per scenario, no load, ~4 min. It found
  two real defects on its first run: the .NET subjects hang without
  `DOTNET_PROCESSOR_COUNT=1`, and a top-level `await` in their `Program.cs` put
  `app.Run()` on the one thread the pinning leaves.
- **Memory and CPU** from `/proc` over the whole process tree, sampled at 20 Hz
  inside each measured window. New columns in every table, a footprint table, and
  the same two columns on the docs site.
- **An `io` scenario** - one Redis `GET` then one parameterised Postgres `SELECT` -
  in all 20 subjects, each with its ecosystem's clients, every pool pinned to 8.
- **`bun run drivers`** - `pg`/`ioredis` on Bun against `Bun.SQL`/`Bun.RedisClient`
  on Bun, because the io table alone cannot separate the client from the runtime.

What the run says:

- The framework tax disappears under I/O. dunx is 4.4% off raw `Bun.serve` on
  `validate` and 0.3% off on `io`, inside the spread.
- `Bun.SQL` over `pg` is worth 12.7-22.3% against a 4.2% run-to-run spread. The two
  Redis clients are indistinguishable - the swap changes sign between pairings - and
  they tie despite `Bun.RedisClient` batching a tick's commands where `ioredis`
  does not.
- The runtime is the larger term at -32.4%.

Three things that were wrong and are fixed rather than papered over, all in
`README.md`: Postgres' default `max_connections` cannot hold 20 interleaved
subjects at a pool of 8; redis-py's async pool raises instead of waiting; and a
row's failures are judged as a **rate**, because the count-based rule had to serve
both one non-2xx in 38,909 and 3.3 million connection failures.

### Reviewing

Read `internal/bench/README.md` first - "Running the io scenario", "Blocking
subjects on the io scenario", "Deliberate handicaps" and "Driver cost" carry the
reasoning. `bun run ci` is green, 13/13.

## Summary by CodeRabbit

* **New Features**
  * Added database-and-cache I/O benchmarks across supported runtimes and frameworks.
  * Added PostgreSQL and Redis driver comparison benchmarks.
  * Benchmark reports now include memory, CPU, process-count, and resource-footprint metrics.
  * Added smoke-test tooling for benchmark validation.
  * Benchmark results now rank failures by error rate rather than isolated failures.

* **Documentation**
  * Expanded benchmark methodology, prerequisites, scenarios, results, and resource guidance.
  * Added memory-footprint visualizations and driver comparison reporting.

* **Tests**
  * Added coverage for I/O configuration, resource sampling, reporting, quality thresholds, and documentation metrics.


dunxonu (commented):
**Actionable comments posted: 5**

internal/bench/servers/drivers/node.ts
```diff
@@ -0,0 +1,30 @@
...
+import { DriverPair } from './pair.js';
+
+const pair = await DriverPair.connect('pg', 'ioredis');
+
+const TEXT = { 'content-type': 'text/plain; charset=utf-8' };
+const JSON_TYPE = { 'content-type': 'application/json; charset=utf-8' };
+
+createServer((req, res) => {
+  if (req.url === '/io') {
+    void pair.read().then((payload) => {
```
dunxonu: The `/io` handler on the `node:classic` driver cell has no rejection handler on `pair.read().then(...)`, unlike the identical pattern fixed in node-http.ts

**How it fails:** When the pg or ioredis pool cannot hand out a connection under saturating load, `pair.read()` rejects; the bare `.then(success)` with no failure branch produces an unhandled promise rejection that crashes the `node:classic` process mid-run - the exact incident node-http.ts's comment describes fixing ("the harness recorded the connection failures as 560,964 req/s"), reintroduced in the sibling file.

internal/bench/src/run.ts
```diff
@@ -321,7 +395,12 @@ export const runSuite = async (
       profile,
     );
 
+  const plan = await planIo(chosenScenarios, chosenSubjects.length);
```
dunxonu: planIo's Postgres connection-budget check is sized on `chosenSubjects.length` (pre-filter) instead of `runnable.length` (the subjects `prepare()` actually kept)

**How it fails:** Requesting subjects that include some with a missing toolchain (e.g. a machine with no JVM or .NET SDK) inflates the assumed subject count, so `connectionBudget` in io-fixture.ts computes a higher `needed` than the run will actually open; a Postgres instance with enough `max_connections` for the real `runnable` set can still be reported as insufficient and the whole `io` scenario gets skipped unnecessarily. The same stale-count pattern is at `internal/bench/src/drivers.ts:182` (`probeIo(services, units.length)`, computed before the `runnable` filter at line 208).

internal/bench/src/readme-tables.ts
```diff
@@ -187,6 +299,19 @@ Elysia on this scenario. What remains is dispatch, not validation.
 **Cold start is dunx's clearest loss**: roughly twice raw \`Bun.serve\`, from the
 \`oxc-parser\` preload and eager DI resolution. It does beat Elysia, and every Node
 subject by a wide margin, but it is the number to watch if boot time matters.
+
+**Memory is the second one.** \`@dunx/http\` boots at ${bootMiB('dunx')} MiB against
+raw \`Bun.serve\`'s ${bootMiB('bun-serve')} MiB, for the container and the resolved
+provider graph, and the gap holds under load. It is small next to the Node subjects
+and tiny next to \`spring\`, and it is still a cost the ceiling does not pay.
+${ranIo ? ioProse : ''}
+**CPU per request tracks throughput on the first four scenarios and stops on the
```
dunxonu: The closing "CPU per request... stops on the fifth" paragraph references io-scenario numbers unconditionally, unlike the adjacent `ioProse` block which is gated by `ranIo`

**How it fails:** On a machine with no Redis/Postgres, `planIo` drops the `io` scenario and `ranIo` is false, so `ioSection`/`ioProse` are correctly suppressed - but this paragraph still runs `cpuPerK('bun-serve', 'io')` and `cpuPerK('axum', 'io')`, which resolve to `'-'` since `report.resources` has no `io` rows, publishing "On io both sit near - while Axum spends -" as committed README prose describing a scenario that never ran.

internal/docs/scripts/extract/bench.ts
```diff
@@ -1,10 +1,46 @@
...
+      usage,
+    ]);
+  }
+
+  return [...bySubject].flatMap(([subject, list]) => {
+    const boots = list
+      .map((one) => one.rssBootMiB)
+      .filter((one): one is number => one !== null)
+      .sort((a, b) => a - b);
+    const boot = boots[boots.length >> 1];
```
dunxonu: foldFootprint computes `bootMiB` as `boots[boots.length >> 1]`, which is the upper-middle element rather than a true median, and duplicates the correct median already in internal/bench/src/stats.ts

**How it fails:** For a subject with an even number of scenario samples (e.g. 4 boot readings of 30/31/33/40 MiB), the true median is (31+33)/2=32 but this returns index 2 = 33, silently skewing the boot-memory figure the docs site renders on the Benchmarks page toward the higher of the two middle values every time the sample count is even.

internal/bench/src/io-fixture.ts
```diff
@@ -0,0 +1,188 @@
...
+ * **The scenario is opt-in on the services being there.** `probeIo` connects once
+ * before anything is measured; if either service does not answer, the `io`
+ * scenario is dropped with a line saying so and the run continues, the same way a
+ * missing toolchain drops its subjects.
+ */
+import { RedisClient, SQL } from 'bun';
+import { IO_POOL_SIZE } from '../servers/io/contract.js';
+import type { Scenario } from './types.js';
+
+export const IO_REDIS_KEY = 'bench:greeting';
```
dunxonu: IO_REDIS_KEY is redeclared as a separate literal `'bench:greeting'` instead of being imported from servers/io/contract.ts, which the file's own docstring names as the single source of truth twenty subjects must agree with byte for byte

**How it fails:** If the cache key in contract.ts is ever changed, `seedIo` in io-fixture.ts keeps writing under the old key while every subject reads via contract.ts's constant under the new key, so every request in the `io` scenario silently reports a cache miss (`cached: 'missing'`) instead of failing loudly - the exact drift Rule 2 exists to prevent.

internal/bench/src/report.ts
```diff
@@ -116,16 +180,54 @@ export const formatReport = (report: Report): string => {
...
+    const baseline =
+      baselineRow === undefined || unranked(baselineRow)
+        ? undefined
+        : baselineRow.rps.median;
 
     out.push(
       `\n${scenario.title.toUpperCase()} - ${scenario.method} ${scenario.path}`,
     );
     out.push(`  ${scenario.description}`);
+    const broken = rows.filter((row) => failed(row) > 0);
```
dunxonu: The plain-text report's 'broken' subjects list uses `failed(row) > 0` while the sort/ranking logic three lines above uses the rate-based `unranked(row)`, so the printed message disagrees with actual table behavior.

**How it fails:** A subject with a single transient error out of 100k requests (below the `invalidates` rate threshold) is reported as '...listed last with no ratio' even though `unranked(row)` is false: it keeps its normal sort position and shows a real ratio in the table, contradicting the message printed for it.

internal/bench/src/run.ts
```diff
@@ -64,6 +67,47 @@ const summarise = (
...
+  return {
+    subject: subject.id,
+    scenario: scenario.id,
+    rssBootMiB: rssBootBytes === null ? null : rssBootBytes / MIB,
+    rssPeakMiB: spread(paired.map((one) => one.sample.rssPeakBytes / MIB)),
+    rssMeanMiB: spread(paired.map((one) => one.sample.rssMeanBytes / MIB)),
+    cpuPercent: spread(
+      paired.map((one) => (one.sample.cpuMs / one.sample.elapsedMs) * 100),
+    ),
+    cpuMsPerKiloRequests: spread(
```
dunxonu: `cpuMsPerKiloRequests` is computed via `spread(paired.filter(one => one.load.requests > 0).map(...))`, and `spread([])` returns `{ median: 0, ... }` rather than a 'no data' sentinel.

**How it fails:** If every measured round for a subject has zero successful requests (total failure to serve, which the surrounding comments explicitly anticipate for connection failures), the filtered array is empty and `cpuMsPerKiloRequests.median` reads 0, printed at run.ts:449 as `cpu 0.00 ms/kreq` - implying a free/fast subject instead of 'no data', for a subject whose CPU cost per request is undefined.

internal/bench/src/report.ts
```diff
@@ -38,6 +40,65 @@ const render = (
 const int = (value: number): string =>
   Math.round(value).toLocaleString('en-US');
 
+/**
+ * One row per subject, across every scenario it was measured on.
```
dunxonu: `formatFootprint` in report.ts duplicates `footprintTable` in readme-tables.ts (same grouping/aggregation/sort by peak RSS) and both diverge from `foldFootprint` in internal/docs/scripts/extract/bench.ts, which treats an empty/zero boot-memory reading differently (drops the subject vs. printing '-').

**How it fails:** A future change to how a zero/missing boot-memory reading should render only lands in one of the three near-identical implementations, so the CLI text report, the README table, and the published docs Memory section silently disagree for the same underlying resource data.

internal/bench/src/drivers-tables.ts
```diff
@@ -0,0 +1,148 @@
...
+  readonly config: {
+    readonly connections: number;
+    readonly durationSeconds: number;
+    readonly warmupSeconds: number;
+    readonly runs: number;
+  };
+  readonly units: readonly Unit[];
+}
+
+const int = (value: number): string =>
```
dunxonu: `int()` (round + `toLocaleString('en-US')`) is a byte-identical copy already present in readme-tables.ts, logging-tables.ts, validation-tables.ts and report.ts; `signed()` at line 39 duplicates the one in logging-tables.ts.

**How it fails:** A formatting fix (e.g. locale, rounding mode) applied to one copy leaves the other four silently inconsistent across the bench CLI's different table outputs.

internal/bench/src/drivers.ts
```diff
@@ -0,0 +1,325 @@
...
+};
+
+/**
+ * Interleaved round-robin, for the reason `db-modes.ts` records: the whole output
+ * is the gap between rows that differ by one client, and this machine's
+ * throughput drifts by more than such a gap over the minutes a run takes.
+ */
+const live = await driveUnits(runnable, bring, generator, config);
+
+const collect = (entry: Live<Unit>): Result => {
```
dunxonu: `collect()` re-derives the pair-by-index/filter-null/peak-RSS/median-CPU-per-request logic that `summariseResources` in run.ts already implements over the same `Live<U>.samples`/usage shape.

**How it fails:** A correctness fix to the pairing logic (e.g. the run.ts fix for dropped-round handling) is made in one implementation and not the other, so the driver-comparison table and the main benchmark report can compute resource costs differently from the same kind of sample data.

internal/bench/src/types.ts
```diff
@@ -112,6 +119,27 @@ export interface ScenarioResult {
...
+ * nothing; CPU per request is what separates a subject that spends its time
+ * computing from one that spends it waiting.
+ */
+export interface ResourceUsage {
+  readonly subject: string;
+  readonly scenario: string;
+  /** Resident set right after the first request, before any load. */
+  readonly rssBootMiB: number | null;
+  readonly rssPeakMiB: Spread;
+  readonly rssMeanMiB: Spread;
```
dunxonu: `rssMeanMiB` is computed and stored on every `ResourceUsage` (resources.ts tracks `totalBytes`/`rssMeanBytes` for it) but no consumer (report.ts, readme-tables.ts, internal/docs/src/bench.ts) ever reads it.

**How it fails:** Dead computation shipped end-to-end: every resource sample carries a mean-RSS figure that no table, chart, or report renders, adding upkeep cost with no reader to justify it.

internal/bench/src/io-fixture.ts
```diff
@@ -0,0 +1,198 @@
...
+export const seedIo = async (services: IoServices): Promise<void> => {
+  const sql = new SQL({ url: services.pgUrl, max: 1 });
+  try {
+    await sql.unsafe(`CREATE TABLE IF NOT EXISTS ${IO_TABLE} (
+      id integer PRIMARY KEY,
+      memo text NOT NULL,
+      amount integer NOT NULL
+    )`);
+    await sql.unsafe(`TRUNCATE ${IO_TABLE}`);
+    for (let index = 1; index <= IO_ROWS; index += 1) {
```
dunxonu: `seedIo` inserts its 500 fixture rows with 500 sequential single-row `await sql.unsafe(...)` calls instead of one multi-row insert.

**How it fails:** Every benchmark run that seeds the `io` scenario pays 500 Postgres round trips instead of one, adding latency to every `start`/`smoke` invocation that includes it.

internal/bench/src/run.ts
```diff
@@ -340,22 +422,31 @@ export const runSuite = async (
...
       exec,
+      ioEnvFor(scenario, plan.services),
       profile,
     );
-    results.push(...measured);
-    for (const result of measured) {
+    results.push(...measured.results);
+    resources.push(...measured.resources);
+    for (const result of measured.results) {
       const subject = runnable.find((one) => one.id === result.subject);
```
dunxonu, line 443: Inside the per-result note-printing loop, `measured.resources.find(one => one.subject === result.subject)` does a linear scan per result instead of a map lookup, unlike `report.ts`'s `costs` map built for the identical join.

**How it fails:** O(subjects x results) instead of O(subjects) per measured round; harmless at current subject counts but the same join is already done cheaply elsewhere in this diff (report.ts's `costs` Map), so this is avoidable repeated work introduced fresh by the PR.

dunxonu (commented):
**Actionable comments posted: 8**

petarzarkov:
All 15 comments addressed in `c3e413c`, plus a follow-up in `204f05a` and the retaken run in `9dd8e9d`. Nothing was skipped.

**Real bugs, thank you.** The one that matters most is `servers/drivers/node.ts`: I fixed the unhandled pool rejection in `node-http.ts` and `express.ts` and left the identical one in the sibling whose comment describes the fix. Also fixed: `drivers.ts` kept the `node:classic` cell whenever `buildNodeEntries` returned an entry, which `Bun.build` does whether or not a Node binary exists, so a machine without one aborted the whole harness instead of reporting the four Bun rows; `report.ts` would divide by a `bun-serve` baseline that had itself errored; the connection budget counted requested subjects rather than the ones `prepare()` kept, at both call sites; `Io.cs` sent `host:-1` for a portless Redis URL and dropped credentials, while the Postgres branch beside it already handled both; and `foldFootprint` took the upper of the two middle boot readings and called it a median, now the harness's own.

**Rule 2.** `IO_REDIS_KEY` was declared in both `io-fixture.ts` and `io/contract.ts`. That one would not have failed loudly - a seeder writing one key while twenty subjects read another just reports `cached: "missing"` forever. `contract.ts` owns the key and the table now, and `IO_SELECT` is built from the table name.

**And one claim of mine that was wrong**, which is the most valuable thing this review produced. I had written that `ioredis` "opens one connection and pipelines on it, which is what `Bun.RedisClient`, `StackExchange.Redis` and Lettuce all do too." Checked against the installed source: `enableAutoPipelining` is `false` by default in ioredis 6.0.0. Bun batches a tick's commands into one write; ioredis does not. Both stay at their defaults, because defaults are what the comparison is of, but the subject registry now records which is which - and it sharpens the finding rather than weakening it, since the two tie *despite* the asymmetry.

Two things beyond the review, both prompted by it:

- **`bad` is now judged as a rate, not a count** (`src/quality.ts`). The count-based rule had to serve both Django's one non-2xx in 38,909 (0.0026%) and two subjects recorded at 560,964 req/s of pure connection failures (100%), and it cannot tell those apart. Threshold is one in a thousand; measured failures here have only ever been ~0.003% or above 25%.
- **The run was retaken on Node 24.21.0**, the current LTS, replacing 20.20.2 which was out of maintenance. Express and Nest-on-Express roughly doubled as a share of the baseline; 29 of 30 Node cells improved. The table is in the README under "Deliberate handicaps". This was the known limitation flagged in the PR description and it is now closed.

`bun run ci` is green, 13/13, and the retaken run has zero errors and zero non-2xx across all 20 subjects and 5 scenarios.

internal/bench/src/report.ts
dunxonu: The printed message claims every subject with `failed(row) > 0` is "listed last with no ratio", but only rows crossing `BAD_RATE_FLOOR` (`unranked`) are actually sorted last and hidden - a row with a low failure rate stays ranked and shows a real percentage.

**How it fails:** Django answering 1 non-2xx out of 38,909 requests (0.0026%, below the 0.1% `BAD_RATE_FLOOR` in quality.ts) has `failed(row) = 1 > 0` so it lands in `broken` and the terminal prints "1 subject(s) answered errors ... and are listed last with no ratio: Django", while the table actually keeps Django in its rps-sorted position with a computed percentage - the printed claim is false for that row.

internal/bench/src/io-fixture.ts
dunxonu: seedIo inserts 500 rows via 500 sequential awaited `sql.unsafe(INSERT...)` calls in a for loop instead of one batched multi-row insert.

**How it fails:** Every `bun run start` or `bun run drivers` invocation that includes the io scenario pays 500 serial Postgres round trips just to seed the fixture table, adding avoidable wall-clock time to every local run and to CI, when a single multi-row `INSERT ... VALUES ($1,$2,$3),(...)...` (or `sql.unsafe` with unnest arrays) would do it in one round trip.

internal/bench/src/drivers-tables.ts
```diff
@@ -0,0 +1,148 @@
+/**
+ * Renders the `## Driver cost` section of `README.md` from `results/drivers.json`.
+ * Called by `src/readme-tables.ts`; returns `null` when no driver run has been
+ * recorded, so a checkout without one still builds.
+ */
+import { resultsDir } from './paths.js';
+import type { MachineInfo, Spread } from './types.js';
+
+interface Unit {
```
dunxonu: `Unit`/`DriversReport` in drivers-tables.ts restate `Result`/`Report` from drivers.ts field-for-field instead of importing the type that already describes the exact JSON drivers.ts writes to `results/drivers.json`.

**How it fails:** drivers.ts is the sole producer of `results/drivers.json` and drivers-tables.ts is its only consumer; if a field is added or renamed in `Result`/`Report` (drivers.ts) without updating the hand-copied `Unit`/`DriversReport` here, `drivers-tables.ts` compiles against a stale shape and either drops the new field silently or reads `undefined` for a renamed one, with no compiler error to catch the drift.

internal/bench/servers/io/lazy.ts
```diff
@@ -0,0 +1,35 @@
...
+export const connectLazyIo = async (): Promise<boolean> => {
+  if (!ioEnabled()) return false;
+  const module = await import('./node.js');
+  const client = new module.NodeIo();
+  await client.ready();
+  reader = () => client.read();
+  return true;
+};
+
+export const readLazyIo = (): Promise<IoPayload> => {
```
dunxonu, line 30: The "catch the io read rejection and answer 500 instead of letting it crash the process" logic is hand-written separately in node-http.ts, express.ts and servers/drivers/node.ts around readLazyIo()/pair.read(), instead of the reader itself returning a uniform non-throwing result.

**How it fails:** A future Node-based subject added to the io scenario, or a new caller of readLazyIo(), that omits the explicit `.then(success, failure)` catch reproduces the exact incident the code comments describe: an unhandled rejection kills the process mid-run and gets recorded as a spuriously high throughput number because dropped connections stop counting as requests.

internal/bench/servers/drivers/pair.ts
```diff
@@ -0,0 +1,84 @@
...
+  ioPayload,
+  type IoPayload,
+  pgUrl,
+  redisUrl,
+} from '../io/contract.js';
+
+export type SqlKind = 'bun' | 'pg';
+export type RedisKind = 'bun' | 'ioredis';
+
+interface Row {
```
dunxonu: DriverPair reconstructs the same Bun.SQL/Bun.RedisClient read logic that `BunIo` (servers/io/bun.ts) already implements, including a second copy of the `Row` interface, instead of reusing that class for the bun/bun leg of the comparison.

**How it fails:** A fix to the Bun-native client wrapper (e.g. connection retry, a changed query shape) applied to `BunIo.read()` in servers/io/bun.ts does not automatically apply to `pair.ts`'s `bunSql`/`bunRedis` closures, so the driver-comparison harness's `bun:native` cell can silently diverge from the main io scenario's Bun subject despite both claiming to measure the identical Bun-native path.

internal/bench/src/resources.ts
```diff
@@ -0,0 +1,182 @@
...
+    processes,
+  };
+};
+
+export const procIsReadable = (): boolean => readTree(process.pid) !== null;
+
+/** What one measured window cost. `null` where `/proc` did not answer. */
+export interface ResourceSample {
+  readonly rssPeakBytes: number;
+  readonly rssMeanBytes: number;
```
dunxonu: `rssMeanBytes`/`rssMeanMiB` is sampled, aggregated into a full `Spread`, and written into every `results/latest.json` entry, but no consumer (report.ts, readme-tables.ts, or the docs extraction/React tables) ever reads it.

**How it fails:** Every benchmark run spends sampling and aggregation work computing a mean RSS figure, and every published JSON report carries dozens of `rssMeanMiB` objects, that nothing downstream displays or uses - dead computation and dead data threaded through three files for no reader.

internal/docs/src/pages/Benchmarks.tsx
```diff
@@ -286,6 +296,25 @@ export const Benchmarks = (): React.JSX.Element => {
           <StartupTable rows={startupRows(model)} />
         </Stack>
 
+        {footprintRows(model).length > 0 && (
```
dunxonu: `footprintRows(model)` is invoked twice per render (once for the length check, once as the table's `rows` prop) instead of being computed once into a local `const`.

**How it fails:** Every render of the Benchmarks page (route change, theme toggle, any parent re-render) reruns the full flatMap/sort over `model.footprint` twice for the same result, doing avoidable duplicate work on each render.

dunxonu (commented):
**Actionable comments posted: 7**

- `internal/bench/src/readme-tables.ts:46` cellsFor computes the `vs bun-serve` percentage against the baseline row's raw rps without checking whether that baseline row itself failed above the quality threshold, unlike the equivalent guard in report.ts.

</details>

petarzarkov:
Rounds 2 and 3 addressed - 17 more comments across `37bf430`, `f6e7ac5` and `85e4f36`. Nothing skipped; one was answered rather than changed, with the reason in the code.

**The finding worth reading first is a process one.** `report.ts`'s "listed last with no ratio" message was still filtering on `failed(row) > 0` while the sort three lines above used the rate. My earlier edit to that line never applied - I asserted on some of that patch's replacements and not that one, and a no-match `replace` fails silently. Every replacement since is asserted, and the review caught what my own check did not.

**Correctness**

- `cpuMsPerKiloRequests` is now `Spread | null`. `oha` counts only status-bearing responses, so a subject that dies mid-run reports `requests: 0`, the filtered array is empty, and `spread([])` printed `cpu 0.00 ms/kreq` - the cheapest row in the table, for a process that answered nothing. That is exactly what the two dead Node subjects did, so it was reachable, not hypothetical.
- The README tables ranked against a `bun-serve` that had itself failed, and `taxTable` computed a dunx tax from unranked rows. `report.ts` already refused both; now they agree.
- `rediss://` lost its TLS - the scheme was dropped, so the client would have connected in cleartext to a server that required encryption. Default port follows the same flag now, 6379 plain and 6380 over TLS.

**The rejection trap is closed rather than fixed a third time.** `answerIo` in `servers/io/respond.ts` reads and answers in one call that cannot reject and cannot leave the response open, and both raw `node:http` handlers go through it. That bug was written twice - `node-http.ts`, then `servers/drivers/node.ts` after the first was fixed - and the next Node subject would have been the third. The frameworks need none of it: Express 5, Fastify, Hono and Nest all turn a rejected handler promise into a 5xx themselves, and `respond.ts` says so.

**Rule 2** - `src/format.ts` (`int` had five copies, `signed` two), `src/footprint.ts` (three folds, the third already disagreeing about a subject with no boot reading), `pairRounds` in `resources.ts`, `DriverUnit`/`DriversReport` into `types.ts` beside the two report shapes already there, and `IoRow` into the contract beside the `SELECT` that produces it.

**Waste** - `rssMeanMiB` was computed, stored and shipped end to end with no reader; removed, and stripped from the committed report so the artifact matches the code that writes it, with no measured number changed. `seedIo` did 500 single-row inserts, now one statement. Two linear scans replaced with lookups, one of them a double `footprintRows(model)` per page render.

**Answered rather than changed:** `pair.ts` reconstructing `BunIo`'s read. It cannot reuse those classes - each fixes a *pair* of clients, and the two cells that cross them are the whole reason the file exists. What can be shared now is: the key, the SQL, the pool size, `IoRow` and `ioPayload` all come from `servers/io/contract.ts`, so the query and the bytes cannot drift from the scenario. That reasoning is now in the file's doc comment.

`bun run smoke` is 20/20 on all five scenarios and `bun run ci` is 13/13. No re-measurement: the only subject binary that changed is the .NET one, and for the `redis://127.0.0.1:6379` the harness passes it resolves to byte-identical options (scheme is not `rediss`, port is explicit, no credentials).

internal/bench/src/readme-tables.ts
```diff
@@ -12,61 +12,130 @@
...
   const rows = cellsFor(scenario);
   const head =
-    '| Subject | req/s (median) | stddev | p50 ms | p99 ms | vs `bun-serve` |\n' +
-    '| ------- | -------------: | -----: | -----: | -----: | -------------: |';
+    '| Subject | req/s (median) | stddev | p50 ms | p99 ms | peak MiB | cpu ms/kreq | vs `bun-serve` |\n' +
+    '| ------- | -------------: | -----: | -----: | -----: | -------: | ----------: | -------------: |';
   const body = rows
     .map((row) => {
       const rps = row.id === FOCUS ? `**${int(row.rps)}**` : int(row.rps);
       const pct =
```
dunxonu, line 101: The README's per-row 'unranked' check is `row.unranked || row.pct === null`, but pct is forced to null for every row in a scenario whenever the baseline is invalidated, not just for rows that themselves failed.

**How it fails:** In a scenario where the bun-serve baseline exceeds BAD_RATE_FLOOR while another subject's row is clean (bad === 0, not unranked), base becomes 0 (line 57-61), pct becomes null for all rows (line 81), and the healthy row renders as `- (0 bad)` — falsely implying that row itself had failures, when only the baseline did. The stdout table (report.ts) renders a bare '-' with no fabricated bad count in the equivalent case.

internal/bench/src/drivers-tables.ts
```diff
@@ -0,0 +1,118 @@
...
+ * Called by `src/readme-tables.ts`; returns `null` when no driver run has been
+ * recorded, so a checkout without one still builds.
+ */
+import { dec, int, signed as signedBy } from './format.js';
+import { resultsDir } from './paths.js';
+import type { DriversReport } from './types.js';
+
+const signed = (value: number): string => signedBy(value, 1, '%');
+
+const table = (report: DriversReport): string => {
```
dunxonu, line 12: The driver-cost README table (table(), gap(), worstSpread()) never adopted the bad-rate invalidation gate (invalidates/badRate from quality.ts) that report.ts, readme-tables.ts, and the docs site all use to refuse ranking a row with too many failures.

**How it fails:** DriverUnit.bad is computed in drivers.ts and printed to stdout via note(), but drivers-tables.ts never imports invalidates/badRate/unranked. A driver subject with e.g. 50% failed requests still has its rps.median printed in the req/s column, folded into gap()'s percentage comparisons, and included in worstSpread()'s run-to-run spread claim — silently, with the failure rate visible only in console output, never in the generated README table.

dunxonu (commented):

- `internal/docs/src/bench.ts:98` throughputRows computes pctOfBaseline against the baseline row's raw rps without checking whether the baseline itself is unranked (bad rate over threshold), unlike report.ts and readme-tables.ts which both guard this.

</details>

internal/bench/src/drivers.ts
```diff
@@ -0,0 +1,294 @@
...
+    ),
+  };
+};
+
+const results = live.map(collect);
+const baseline = results.find((row) => row.id === 'bun:native')?.rps.median;
+
+for (const result of results) {
+  note(
+    `${result.id.padEnd(13)} ${Math.round(result.rps.median).toLocaleString('en-US').padStart(9)} req/s` +
```
dunxonu, line 270: The driver harness's stdout logger reimplements format.ts's `int` helper inline instead of importing it, duplicating the exact formatting logic the sibling drivers-tables.ts already imports from format.ts.

**How it fails:** `Math.round(result.rps.median).toLocaleString('en-US')` at line 270 is the literal body of `int()` in format.ts, but drivers.ts has no import from './format.js' at all. A future change to number formatting (e.g. locale, rounding) applied to format.ts's `int` silently does not apply to this stdout output, causing the console table and the README table to disagree in formatting for the same numbers.

internal/bench/src/report.ts
```diff
@@ -116,16 +158,54 @@ export const formatReport = (report: Report): string => {
...
+    const broken = rows.filter(unranked);
+    if (broken.length > 0) {
+      out.push(
+        `  ${broken.length} subject(s) answered errors or non-2xx and are listed last with no ratio: ` +
+          broken
+            .map((row) => labels.get(row.subject) ?? row.subject)
+            .join(', '),
+      );
+    }
+    const costs = new Map(
```
dunxonu, line 202: The join of ResourceUsage (rss/cpu) onto a scenario+subject throughput row is implemented independently three times (report.ts's costs Map, readme-tables.ts's costFor, and extract/bench.ts's inline .find()), each with a different sentinel for the missing case ('-', undefined, null).

**How it fails:** A future change to the join semantics (e.g. how a subject with no resource reading for a scenario is handled) needs to land in three places to stay consistent; today the fallback spellings already diverge ('-' vs undefined vs null), the same failure mode foldFootprint's own comment in this PR describes fixing once already ('the third already disagreed with the other two about a subject with no boot reading') but did not extend to this per-scenario join.

PR #74, opened 2026-09-09
fix(review): scope a re-review to what changed, so a pull request can converge

### LINKED ISSUE #73: dunxonu re-reviews the whole diff on every push and never approves, so a PR accumulates unbounded rounds
`.github/workflows/dunxonu-review.yml` runs on every `synchronize`, and reviews
the whole pull request each time. On a PR that takes several pushes to satisfy,
that is not a review, it is a lottery re-rolled per push.

## What #70 did

Five review rounds, **24 inline findings**, and `COMMENTED` every single time:

```
03:44 COMMENTED    05:18 COMMENTED
04:12 COMMENTED    05:34 COMMENTED
05:11 COMMENTED
```

It never approved, at any point, including after every finding it had raised was
addressed.

### 1. Late rounds report code that has not changed since the first push

Round 3 (05:18) flagged `rssMeanMiB` in `src/resources.ts` and the duplicated
`Row` in `servers/drivers/pair.ts`. Both files were **untouched** between the
first push and the commit that round reviewed:

```
$ git log --oneline c1d9f3d..204f05a -- internal/bench/src/resources.ts
(nothing)
$ git log --oneline c1d9f3d..204f05a -- internal/bench/servers/drivers/pair.ts
(nothing)
```

Those findings were equally true in round 1 and equally reportable. They were not
reported, because the review is non-deterministic over the same input and each
push buys another draw. Every push therefore has a fresh chance to surface
something in code nobody touched, and a PR converges only by luck.

The findings themselves were good - both were real and both got fixed. The problem
is when they arrive, not whether they are right.

**Suggested fix:** for a `synchronize` after the first review, review
`<last-reviewed-sha>..HEAD` rather than `main...HEAD`. The last reviewed sha is
already available - `post-review.ts` submits against
`github.event.pull_request.head.sha`, so the previous one can be read back off the
PR's review list. A first review still sees the whole diff; later ones see what
actually changed.

### 2. The approve path is close to unreachable

`buildReview` in `scripts/post-review.ts` approves when the typed findings list is
present, empty, and there is no prose outside it, or when the `VERDICT: APPROVE`
sentinel appears with no findings. Both require the reviewer to find **nothing at
all**, and a model asked to review at `medium` effort will nearly always surface
something - a nit, a naming point, a duplicated formatter.

So "approve" in practice means "found literally zero things", which on any
non-trivial diff is never. There is no way to say "these are worth knowing and
none of them block".

**Suggested fix:** let severity decide the event, not the count. `ReportFindings`
already carries a `category` and the skill has severity levels; approving with
findings attached when none is above a bar would make the signal mean something.
`COMMENT` would then mean "nothing blocking, here are notes" and
`CHANGES_REQUESTED` would mean what it says.

## Why it matters

dunxonu submits `COMMENTED`, which does not block merge - so today this is cost
and noise rather than a gate failure. But it is real cost: five runs of a
20-minute-timeout job on one PR, and a reviewer that never signals "done" gives a
human no way to tell "still working" from "converged".

The `concurrency` block already collapses rapid pushes into one review, so the
cadence is deliberate. This is about what each review looks at and what it is
allowed to conclude.

## Not in scope

The review quality is good and should not be dialled down to fix this. Of the 24
findings on #70, one was declined and the rest were real, including a bug I had
introduced twice and an incorrect claim I had written into the docs. The fix is to
stop re-litigating unchanged code and to let it approve, not to make it find less.

petarzarkov:
Closes #73.

dunxonu re-reviewed the whole diff on every push, and it is not deterministic over
the same input, so unchanged code got a fresh chance to yield a finding on each
run.

#70 ran **five rounds, 24 findings, and never approved**. Round 3 reported
`rssMeanMiB` in `src/resources.ts` and a duplicated `Row` in
`servers/drivers/pair.ts` - and `git log` shows neither file was touched between
the first push and the commit that round reviewed:

```
$ git log --oneline c1d9f3d..204f05a -- internal/bench/src/resources.ts
(nothing)
```

Both findings were good, and both were equally reportable in round 1. The problem
is when they arrive, not whether they are right.

## The change

A review after the first reports only on files touched since **this account's**
previous review of the pull request. The first review still sees everything.

Out-of-scope findings are counted in the body, not dropped in silence - they were
reportable earlier and are still true, and hiding them would be this script
deciding what the author may see.

That also makes the approval reachable, which is the other half of #73. Approving
used to require the reviewer to find literally nothing, which on a real diff never
happened. Now: fix what was raised, push, and a review whose scope comes back
clean approves.

## Where it lives, and why not in the prompt

`post-review.ts`, deterministically. `/code-review` takes a pull request, not a
range, and an appended instruction to narrow it would lose to the command's own -
which this workflow already records twice as the reason two other appended
instructions were dropped.

## Edge cases

- **Last review found by authenticated account**, not a hardcoded login, so a
  human's review on the pull request cannot narrow the bot's scope.
- **A failed `compare`** - the old commit collected after a force-push - widens
  back to the whole diff rather than narrowing to nothing. Reviewing too much is
  the bug being fixed; reviewing nothing silently would be worse than it.
- **A finding with no file** cannot be attributed to one and is always kept.

Six new tests, including the convergence case: everything left in scope clean,
findings carried, `APPROVE`.

## Not done

#73 also suggested letting severity pick the event so a review could approve *with*
non-blocking notes. `ReportFindings` gives `post-review.ts` a category and a
verdict but no severity, so there is nothing to threshold on, and scoping makes
the approval reachable without it. Worth revisiting only if approvals still turn
out to be rare.

## Reviewing this

This pull request changes the reviewer that reviews it. The first review here runs
under the old behaviour, since the workflow on `main` is what runs; the effect is
visible from the second push onward, and on the next pull request.

## Summary by CodeRabbit

- **New Features**
  - Reviews now focus inline findings on files changed since the reviewer’s previous review.
  - Findings from unchanged files are summarized in the review body instead of being posted again.
  - Reviews still post actionable findings that are not associated with a file.
  - Reviews fall back to the full diff when file comparisons are incomplete.

- **Documentation**
  - Added workflow guidance explaining how review findings are scoped and summarized.

### DIFF
--- a/scripts/post-review.ts
+++ b/scripts/post-review.ts
@@ -54,6 +54,26 @@ export type Commentable = ReadonlyMap<string, ReadonlySet<number>>;
  */
 export type Scope = ReadonlySet<string> | null;
 
+/**
+ * GitHub's compare endpoint returns at most this many files and does not page
+ * past them, so a longer list has been silently truncated.
+ */
+export const COMPARE_FILE_CAP = 300;
+
+/**
+ * The scope a compare result supports, or `null` when it cannot be trusted to be
+ * complete.
+ *
+ * Suppressing a finding because its file fell off the end of a truncated list is
+ * the exact failure this scoping exists to prevent: a genuinely new problem,
+ * filed under "not repeated here", on a review that could then approve.
+ *
+ * Being wrong about the cap is safe in the direction that matters - too low
+ * reviews more than it needs to, and only too high could suppress.
+ */
+export const scopeFromCompare = (touched: readonly string[]): Scope =>
+  touched.length >= COMPARE_FILE_CAP ? null : new Set(touched);
+
 const VERDICT = /^VERDICT: (APPROVE|COMMENT)$/gm;
 
 /**
@@ -311,10 +331,11 @@ const lastReviewedSha = async (
 /**
  * The files touched since this reviewer last looked, or `null` when it has not.
  *
- * A `compare` that fails - the old commit garbage-collected after a force-push,
- * most likely - widens the scope back to everything rather than narrowing it to
- * nothing. Reviewing too much is the behaviour being fixed; reviewing nothing
- * silently would be worse than the bug.
+ * Both failure modes here widen rather than narrow. A `compare` that throws -
+ * the old commit garbage-collected after a force-push, most likely - and a
+ * response at the file cap both fall back to the whole diff. Reviewing too much
+ * is the behaviour being fixed; reviewing nothing silently would be worse than
+ * the bug.
  */
 const scopeSince = async (
   repo: string,
@@ -328,10 +349,13 @@ const scopeSince = async (
       await gh(['api', `repos/${repo}/compare/${since}...${head}`]),
     ) as { files?: { filename: string }[] };
     const touched = compared.files?.map((file) => file.filename) ?? [];
+    const scope = scopeFromCompare(touched);
     console.log(
-      `Reviewing ${touched.length} file(s) changed since ${since.slice(0, 7)}.`,
+      scope === null
+        ? `Compare returned ${touched.length} files, at or past the ${COMPARE_FILE_CAP} the API caps at, so the list may be short. Reviewing the whole diff.`
+        : `Reviewing ${touched.length} file(s) changed since ${since.slice(0, 7)}.`,
     );
-    return new Set(touched);
+    return scope;
   } catch (error) {
     console.log(
       `Could not compare ${since.slice(0, 7)}...${head.slice(0, 7)}, reviewing the whole diff: ${String(error)}`,
--- a/scripts/post-review.ts
+++ b/scripts/post-review.ts
@@ -32,6 +32,28 @@ export interface Review {
 /** Line numbers on the right-hand side of the diff, which are the commentable ones. */
 export type Commentable = ReadonlyMap<string, ReadonlySet<number>>;
 
+/**
+ * The files this review is allowed to report on, or `null` for all of them.
+ *
+ * `null` is the first review of a pull request, which sees the whole diff. Every
+ * review after it sees only the files touched since the previous one.
+ *
+ * **This is what lets a pull request converge.** The reviewer runs against the
+ * full diff on every push and is not deterministic over it, so unchanged code
+ * gets a fresh chance to yield a finding on each run. Measured on #70: round 3
+ * reported `rssMeanMiB` in `src/resources.ts` and a duplicated `Row` in
+ * `servers/drivers/pair.ts`, and `git log` shows neither file was touched between
+ * the first push and the commit that round reviewed. Both findings were equally
+ * true and equally reportable in round 1. Five rounds, 24 findings, and never an
+ * approval, because there was always something new to say about code nobody had
+ * changed.
+ *
+ * Findings outside the scope are counted in the body rather than dropped in
+ * silence: they were reportable earlier and are still true, and hiding them would
+ * be this script deciding what the author may see.
+ */
+export type Scope = ReadonlySet<string> | null;
+
 const VERDICT = /^VERDICT: (APPROVE|COMMENT)$/gm;
 
 /**
@@ -125,8 +147,16 @@ const at = (finding: Finding): string =>
 export const buildReview = (
   result: string,
   commentable: Commentable,
+  scope: Scope = null,
 ): Review => {
-  const findings = findingsIn(result);
+  const all = findingsIn(result);
+  const findings =
+    all === undefined || scope === null
+      ? all
+      : all.filter(
+          (finding) => finding.file === undefined || scope.has(finding.file),
+        );
+  const carried = (all?.length ?? 0) - (findings?.length ?? 0);
   const sentinel = verdictIn(result);
   const clean =
     findings !== undefined &&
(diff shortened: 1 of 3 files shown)


scripts/post-review.ts
```diff
@@ -244,6 +289,57 @@ const gh = async (args: readonly string[]): Promise<string> => {
...
+
+/**
+ * The files touched since this reviewer last looked, or `null` when it has not.
+ *
+ * A `compare` that fails - the old commit garbage-collected after a force-push,
+ * most likely - widens the scope back to everything rather than narrowing it to
+ * nothing. Reviewing too much is the behaviour being fixed; reviewing nothing
+ * silently would be worse than the bug.
+ */
+const scopeSince = async (
```
dunxonu, line 340: `scopeSince` builds the scope Set directly from `compared.files` with no pagination handling, but GitHub's compare API caps the returned file list at 300 entries with no way to page past it.

**How it fails:** A previous review lands on a commit, then a subsequent push changes more than 300 files since that commit (plausible in this monorepo across a large refactor or a rebase pulling in unrelated commits). `compared.files` silently contains only the first 300 filenames, so files past that cut are absent from `scope`. Genuinely new findings in those omitted files get filtered out by the `scope.has(finding.file)` check and folded into the 'carried, not repeated' body note instead of being posted as new inline comments - the opposite of what scoping is meant to guarantee (new changes always get reported).

dunxonu (commented):
**Actionable comments posted: 1**

- `scripts/post-review.ts:183` Scope filtering only applies to the parsed `findings` array; the raw-prose fallback branch returns `withoutVerdict(result)`, the full unfiltered original text, so any out-of-scope finding still shows up verbatim in the review body whenever the model's output has anything besides a bare JSON array.

</details>

petarzarkov:
Both of you found the same thing independently, and it is a real one - fixed in `0cf4264`.

The compare endpoint caps its file list at 300 and does not page past it, so a push touching more than that since the last review comes back silently short. Narrowing to a truncated list would file **genuinely new** findings under "not repeated here" and could then approve. That is precisely the failure this scoping was added to prevent, reintroduced by the mechanism meant to prevent it.

At or past the cap it now widens to the whole diff, which is the policy the throwing branch beside it already used: reviewing too much is the bug being fixed, reviewing nothing silently is worse than it. Being wrong about 300 is safe in the direction that matters - too low reviews more than it needs to, and only too high could suppress.

One thing I did beyond the suggested diff. Both proposals put the check inline in `scopeSince`, which is IO-bound and has no test. An untested branch in that path is how this would go unnoticed a second time, so the decision is now `scopeFromCompare`, pure and beside the type it produces, with three tests including the boundary at exactly the cap.

I considered computing the file list from local git instead - the workflow checks out with `fetch-depth: 0`, so `git diff --name-only since..head` would be exact and uncapped. I did not, because it makes the scope depend on what `actions/checkout` left in the workspace for a `pull_request` event, and trading a documented API cap this repository will realistically never hit for an assumption about the merge ref is the worse deal. Noting it in case someone disagrees.

`bun run ci` 13/13.

scripts/post-review.ts
```diff
@@ -244,6 +309,61 @@ const gh = async (args: readonly string[]): Promise<string> => {
...
 
+/**
+ * The commit this reviewer last posted a review against on this pull request, or
+ * `undefined` if this is the first.
+ *
+ * By the authenticated account rather than a hardcoded login: the workflow runs
+ * as whatever `DUNXONU_TOKEN` belongs to, and a review left by a human or by
+ * another bot must not narrow what this one looks at.
+ */
+const lastReviewedSha = async (
```
dunxonu, line 320: lastReviewedSha's two `gh api` calls (fetching the account login and the PR's review list) are not wrapped in a try/catch, unlike the compare call right after it in scopeSince.

**How it fails:** A transient failure in `gh api user` or `gh api --paginate repos/{repo}/pulls/{number}/reviews` (rate limit, network blip, secondary GitHub API limit) throws inside lastReviewedSha, propagates unhandled through scopeSince's `await lastReviewedSha(...)` at line 345 (which is called before the try block starts at line 347), and crashes the whole `post-review.ts` process before any review is posted. This is exactly the failure mode the surrounding comment says the code avoids ('reviewing nothing silently would be worse than the bug') for the compare call, but the same protection was not extended to the two calls immediately above it, so a flaky GitHub API turns a push into no review at all instead of a widened, full-diff review.

scripts/post-review.ts
```diff
@@ -244,6 +309,61 @@ const gh = async (args: readonly string[]): Promise<string> => {
...
+ * is the behaviour being fixed; reviewing nothing silently would be worse than
+ * the bug.
+ */
+const scopeSince = async (
+  repo: string,
+  number: string,
+  head: string,
+): Promise<Scope> => {
+  const since = await lastReviewedSha(repo, number);
+  if (since === undefined || since === head) return null;
```
dunxonu, line 346: When the last-reviewed commit equals the current head (e.g. a 'reopened' trigger with no new push since the last review), scopeSince returns null and the review is treated as unscoped, re-posting every finding as a fresh inline comment.

**How it fails:** A PR is closed and reopened without any new commits: the `reopened` workflow trigger fires, `since === head` is true, so `scopeSince` short-circuits to `null` (full scope) instead of an empty scope, and `buildReview` re-reports every previously-posted finding as new inline comments rather than converging or staying silent, contrary to the PR's stated convergence goal.

dunxonu (commented):

PR #75, opened 2026-09-09
docs(architecture): attribute the boot cost, which was never split

### LINKED ISSUE #72: Track: dunx boots at 1.93x raw Bun.serve and +16.7 MiB, both now measurable
Tracking, not a defect. Both figures are attributed already; what changed is that
`internal/bench` now measures the second one, so a regression in either is
visible instead of inferred.

## The numbers

5950X, Bun 1.4.2, `results/latest.json` 2026-09-09. Cold start is spawn to first
served request, median of 7; boot memory is the resident set of the whole process
tree the moment it answers, before any load.

| Subject | cold start | vs `bun-serve` | boot MiB |
| ------- | ---------: | -------------: | -------: |
| `bun-serve` (raw) | 24.0 ms | 1.00x | 34.6 |
| Hono (Bun) | 27.4 ms | 1.14x | 36.6 |
| Elysia | 51.9 ms | 2.17x | 47.2 |
| **`@dunx/http`** | **46.2 ms** | **1.93x** | **51.3** |
| `@dunx/http` + `@arkv/logger` | 61.4 ms | 2.56x | 54.8 |

dunx beats Elysia on cold start and loses to it on memory. Both lose to Hono,
which is the honest comparison for "what does the framework add to an empty
process".

## What is already known

Cold start is attributed to the `@dunx/transform` preload (`oxc-parser`, a Rust
parser loaded through N-API) plus eager DI resolution at boot - which is a
deliberate trade, since resolving early is what keeps it out of the per-request
number. `README.md` says so.

The +16.7 MiB is **not** attributed anywhere. It has never been measured before
this, so nobody has split it between the parser, the container, the resolved
provider graph and the route table.

## Why this is tracking rather than work

Neither number is on anyone's critical path today, and the per-request cost is
where the interesting findings are - see #71, where one logging binding costs 300x
what the whole framework does. Cold start matters for serverless, which dunx does
not currently target, and 51 MiB is unremarkable for a container.

Worth acting on if either becomes a goal. Until then this issue exists so the
figures have somewhere to live and a regression has something to be compared
against.

## If someone does pick it up

`bun run start --profile heap --subjects dunx` writes a Bun heap profile per
subject, and `--profile cpu` the same for boot CPU. That is the tool for splitting
the 16.7 MiB, and it is already wired up.

petarzarkov:
Closes #72.

That issue recorded dunx booting at 1.93x raw `Bun.serve` and +16.7 MiB, and noted
the memory had **never been attributed**. It is now, next to the paragraph in
`docs/architecture/dependency-injection.md` that asserts production carries
`oxc-parser` "in order to run code that was already transformed at build time" -
which is exactly the claim these numbers price.

A ladder of whole processes, each adding one step to the one above it, median of
11 cold starts on Bun 1.4.2:

| Rung | ms | MiB | adds ms | adds MiB |
| ---- | -: | --: | ------: | -------: |
| bare `bun` | 3.7 | 12.6 | | |
| + `@dunx/transform` preload | 9.8 | 26.7 | +6.1 | +14.1 |
| + `import '@dunx/core'` | 11.4 | 27.4 | +1.5 | +0.7 |
| + `import '@dunx/http'` | 16.9 | 30.2 | +5.5 | +2.7 |
| + `HttpFactory.create` | 22.3 | 41.3 | +5.4 | +11.1 |
| + `listen()` | 24.8 | 43.2 | +2.5 | +1.9 |
| raw `Bun.serve`, for scale | 4.4 | 14.6 | | |

**The preload is the largest single item in both columns** - 49% of the memory and
30% of the time - and it is paid before any application code runs. The 14.1 MiB is
`oxc-parser` loading into a script with no classes in it at all. The container and
resolved provider graph are the other half, at 11.1 MiB and 5.4 ms.

**And that half is already avoidable.** `bun build --compile` writes the records at
build time and the binary carries no parser, which `examples/binary` already
demonstrates and documents. So the 14.1 MiB is a cost of the `bun run` deployment
shape rather than of the design - which is worth knowing when choosing between the
two, and was not written down anywhere.

No code changed, and nothing here argues for a change. #72 was tracking, the
unknown in it is now known, and this is where it lives.

The absolute figures are lower than the benchmark's because the harness measures
spawn to first served request and its subjects import zod; the deltas are what
this table is for.

## Summary by CodeRabbit

* **Documentation**
  * Added benchmark data covering startup time and memory usage across different Bun execution stages.
  * Documented the startup overhead associated with the transformation preload.
  * Clarified that compiled deployments avoid this preload cost, while `bun run` deployments incur it before application code executes.

### DIFF
--- a/docs/architecture/dependency-injection.md
+++ b/docs/architecture/dependency-injection.md
@@ -81,6 +81,34 @@ It would also cost `@dunx/core` its empty dependency list. The transform needs
 `oxc-parser`, a native binary. Every production deployment would then carry it,
 in order to run code that was already transformed at build time.
 
+### What the preload costs at boot
+
+Measured on Bun 1.4.2, a 5950X, median of 11 cold starts, each rung a whole
+process adding one step to the one above it:
+
+| Rung                        |   ms |  MiB | adds ms | adds MiB |
+| --------------------------- | ---: | ---: | ------: | -------: |
+| bare `bun`                  |  3.7 | 12.6 |         |          |
+| + `@dunx/transform` preload |  9.8 | 26.7 |    +6.1 |    +14.1 |
+| + `import '@dunx/core'`     | 11.4 | 27.4 |    +1.5 |     +0.7 |
+| + `import '@dunx/http'`     | 16.9 | 30.2 |    +5.5 |     +2.7 |
+| + `HttpFactory.create`      | 22.3 | 41.3 |    +5.4 |    +11.1 |
+| + `listen()`                | 24.8 | 43.2 |    +2.5 |     +1.9 |
+| raw `Bun.serve`, for scale  |  4.4 | 14.6 |         |          |
+
+**The preload is the largest single item in both columns** - 49% of the memory a
+dunx process holds over a raw `Bun.serve` one, and 30% of the time. It is paid
+before any application code runs: the 14.1 MiB is `oxc-parser` loading into a
+script with no classes in it at all.
+
+The container and the resolved provider graph are the other half, at 11.1 MiB and
+5.4 ms.
+
+`bun build --compile` already avoids the preload half - the records are written at
+build time and the binary needs no parser, which is what `examples/binary`
+demonstrates. So the 14.1 MiB is a cost of the `bun run` deployment shape rather
+than of the design, and anyone choosing between the two can now price it.
+
 So registration stays explicit, and the failure mode is closed off instead:
 
 ### The missing-transform guard
--- a/docs/architecture/dependency-injection.md
+++ b/docs/architecture/dependency-injection.md
@@ -84,11 +84,14 @@ in order to run code that was already transformed at build time.
 ### What the preload costs at boot
 
 Measured on Bun 1.4.2, a 5950X, median of 11 cold starts, each rung a whole
-process adding one step to the one above it:
+process adding one step to the one above it. The `oxc-parser` row is the
+exception: it is measured against bare `bun`, so it splits the preload below it
+rather than stacking on it.
 
 | Rung                        |   ms |  MiB | adds ms | adds MiB |
 | --------------------------- | ---: | ---: | ------: | -------: |
 | bare `bun`                  |  3.7 | 12.6 |         |          |
+| `import 'oxc-parser'` alone |  8.9 | 24.7 |    +5.2 |    +12.1 |
 | + `@dunx/transform` preload |  9.8 | 26.7 |    +6.1 |    +14.1 |
 | + `import '@dunx/core'`     | 11.4 | 27.4 |    +1.5 |     +0.7 |
 | + `import '@dunx/http'`     | 16.9 | 30.2 |    +5.5 |     +2.7 |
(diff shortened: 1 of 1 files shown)


docs/architecture/dependency-injection.md
```diff
@@ -81,6 +81,34 @@ It would also cost `@dunx/core` its empty dependency list. The transform needs
...
+| + `HttpFactory.create`      | 22.3 | 41.3 |    +5.4 |    +11.1 |
+| + `listen()`                | 24.8 | 43.2 |    +2.5 |     +1.9 |
+| raw `Bun.serve`, for scale  |  4.4 | 14.6 |         |          |
+
+**The preload is the largest single item in both columns** - 49% of the memory a
+dunx process holds over a raw `Bun.serve` one, and 30% of the time. It is paid
+before any application code runs: the 14.1 MiB is `oxc-parser` loading into a
+script with no classes in it at all.
+
+The container and the resolved provider graph are the other half, at 11.1 MiB and
```
dunxonu: The doc calls the container/provider-graph cost "the other half" of the boot-time overhead, but the `+http` import step (5.5ms) is nearly as large as the container step (5.4ms) and the preload itself (6.1ms), so two distinct costs are misattributed to one.

**How it fails:** A reader trying to reduce dunx boot latency reads 'the container and the resolved provider graph are the other half, at 11.1 MiB and 5.4 ms' and concludes DI resolution is the only other major time cost besides the preload. In fact, of the 14.3ms remaining after the preload, `import '@dunx/http'` alone accounts for 5.5ms (more than the container's 5.4ms) - a cost the sentence folds into an unstated remainder, leading the reader to under-price the import step when deciding where to optimize.

dunxonu (commented):

petarzarkov:
All three addressed in `238e883`. All three were mine claiming past what the ladder measured, which is worth naming as a pattern rather than three separate slips.

**"The container and the resolved provider graph are the other half"** was true of memory and misleading about time. Of the 21.1 ms over the baseline: preload 6.1, `import '@dunx/http'` 5.5, `HttpFactory.create` 5.4 - within a millisecond of each other. A reader reducing boot latency has three targets, and the import is as large as the container. Memory is the lopsided column, not time, and the two are now stated separately instead of collapsed into "half".

**"The 14.1 MiB is `oxc-parser` loading"** attributed the whole preload to the parser. Rather than hedge it, I measured it: `import 'oxc-parser'` alone against bare `bun` is **+12.1 MiB and +5.2 ms**, so 86% of the preload, and the rest is the plugin and its registration. That row is in the table now, marked as measured against bare `bun` rather than stacked, since it splits the row below it instead of adding to it.

**"`bun build --compile` already avoids the preload half"** named a command that is not what does it, and the correction is sharper than the comment: `examples/binary/scripts/build.ts` calls `Bun.build({ compile, plugins: [depsPlugin] })`, and a plain `bun build --compile` applies no plugin - so it would bake in **no records at all**, which is the opposite outcome from the one I attributed to it. Someone following the doc literally would have shipped a binary that fails at boot on the missing-transform guard.

`bun run ci` 13/13.

Incidentally this is the first pull request reviewed since #74 landed, but it is a *first* review, so it saw the whole diff by design. The scoping only shows from the second push - which this comment's commit is, so the next review is the real test.

docs/architecture/dependency-injection.md
```diff
@@ -81,6 +81,47 @@ It would also cost `@dunx/core` its empty dependency list. The transform needs
...
+| + `import '@dunx/http'`     | 16.9 | 30.2 |    +5.5 |     +2.7 |
+| + `HttpFactory.create`      | 22.3 | 41.3 |    +5.4 |    +11.1 |
+| + `listen()`                | 24.8 | 43.2 |    +2.5 |     +1.9 |
+| raw `Bun.serve`, for scale  |  4.4 | 14.6 |         |          |
+
+**The preload is the largest single item in memory**, at 49% of what a dunx
+process holds over a raw `Bun.serve` one, and it is paid before any application
+code runs: the rung above adds it to a script with no classes in it at all.
+`oxc-parser` on its own accounts for 12.1 of those 14.1 MiB, measured the same
+way; the rest is the plugin and its registration.
```
dunxonu, line 106: The new boot-cost table gives a 'raw Bun.serve' cold-start figure of 4.4 ms and a full dunx-boot figure of 24.8 ms, both roughly 4x lower than the same metrics already published in docs/guide/01-introduction.md (19.4 ms and 42.6 ms respectively), with no cross-reference or explanation of why they differ.

**How it fails:** A reader cross-references the introduction guide ('raw Bun.serve: 19.4 ms', '@dunx/http: 42.6 ms', both 'cold start to first served request') against this new table ('raw Bun.serve, for scale: 4.4 ms', '+ listen(): 24.8 ms') and concludes the docs contradict each other on dunx's boot cost. The actual cause is that this table stops at listen() returning while the guide measures through an actual served HTTP request, but neither page states that difference, so the discrepancy reads as an unreconciled error rather than two intentionally different milestones.

dunxonu (commented):

PR #76, opened 2026-09-09
docs: refresh the guide's figures and say which milestone each table measures

petarzarkov:
Follows up a review comment on #75 that landed as it merged. It was right, and it
turned out to be two problems rather than one.

## The contradiction

The boot table added in #75 reads against the introduction guide as an
unreconciled error: 4.4 ms vs 19.4 for raw `Bun.serve`, 24.8 vs 42.6 for dunx.

The cause is that they measure **different milestones** - the architecture ladder
stops when the process does, the guide measures through a request actually served,
about 20 ms later on both. Neither page said so. Both now do, on the table itself
rather than in a paragraph someone has to find.

## The staleness, which the first problem exposed

The guide's numbers are **hand-transcribed**, from a Bun 1.4.0 run dated
2026-08-22. The run merged yesterday moved every row, and the Node ones furthest,
because those were taken on Node 20 and the suite is on the current LTS now:

| | was | now |
| - | -: | -: |
| Express, cold start | 126.5 ms | 110.1 ms |
| Fastify | 154.6 ms | 135.4 ms |
| NestJS (Express) | 278.8 ms | 243.7 ms |
| raw `Bun.serve` | 19.4 ms | 24.0 ms |
| **`@dunx/http`** | **42.6 ms** | **46.2 ms** |

Throughput is refreshed the same way, `dunx costs 1% to 7%` becomes `under 1% to
8%` since `validate` moved to 92.1%, and the `io` row is added - where the
framework is 0.3% of a request that touches Redis and Postgres, which is the
honest framing for the four rows above it.

`tools/mcp/src/generated.ts` is regenerated, since it bundles the guides.

## Worth someone's attention later

**Nothing guards those figures.** The benchmark README is generated from
`results/latest.json` precisely so it cannot drift, and this guide is the
hand-written copy of the same numbers - so it is the one that goes stale in
silence. It just did, for two and a half weeks, and it took a bot noticing a
contradiction with a different document to surface it.

Not fixed here because the fix is a generator or a guard, and this branch is a
correction. Happy to open an issue if you want it tracked.

## Summary by CodeRabbit

* **Documentation**
  * Updated benchmark results with newer runtime versions, measurement runs, throughput figures, and an added I/O scenario.
  * Clarified cold-start timing methodology, including when measurements end and how results differ across benchmark tables.
  * Added details about measured Redis/Postgres I/O, memory usage, CPU-per-request metrics, and benchmark limitations.
  * Updated the short-lived-process startup estimate from approximately 43 ms to 46 ms.


docs/architecture/dependency-injection.md
```diff
@@ -88,6 +88,13 @@ process adding one step to the one above it. The `oxc-parser` row is the
 exception: it is measured against bare `bun`, so it splits the preload below it
 rather than stacking on it.
 
+**These stop when the process does, not when it has served something**, which is a
+different milestone from the one `docs/guide/01-introduction.md` publishes and
+about 20 ms earlier on both subjects. The two tables are not comparable and
+neither is wrong; this one exists to apportion the boot, and the guide's exists to
+state what a caller waits for. The apps here are also minimal, where the harness's
```
dunxonu: New sentence is grammatically broken/incomplete: a dangling possessive with no noun.

**How it fails:** The added text reads "The apps here are also minimal, where the harness's carry zod." — "harness's" has no following noun (likely meant "harness's apps"), so the sentence is unparseable as written; a reader cannot recover the intended contrast between the boot-cost apps and the guide-benchmark harness's apps.

docs/guide/01-introduction.md
```diff
@@ -139,18 +144,22 @@ Startup is the clearest loss, and it is a real one:
...
-| NestJS (Express) |                                         278.8 ms |
+| raw `Bun.serve`  |                                          24.0 ms |
+| **`@dunx/http`** |                                      **46.2 ms** |
+| Elysia           |                                          51.9 ms |
+| raw `node:http`  |                                          80.7 ms |
+| Express          |                                         110.1 ms |
+| Fastify          |                                         135.4 ms |
+| NestJS (Express) |                                         243.7 ms |
 
 Both Bun figures roughly halved on Bun 1.4, from 54.8 ms and 28.7 ms. The ratio has
```
dunxonu: Pre-existing 'roughly halved' claim next to the newly updated cold-start table no longer squares with the refreshed figures.

**How it fails:** Table now shows raw `Bun.serve` 24.0 ms and `@dunx/http` 46.2 ms; the unchanged sentence claims both 'roughly halved on Bun 1.4, from 54.8 ms and 28.7 ms' — 54.8 halved (~27) is in the ballpark of the new raw figure, but 28.7 halved (~14) is nowhere near dunx's 46.2 ms (which is larger, not smaller), so the pairing of baseline values to rows looks inverted or stale; whoever refreshed the numbers above it should re-check this claim rather than leave it.

dunxonu (commented):

petarzarkov:
All four addressed in `69e9f06`. Two of them were contradictions my own previous commit on this branch created, which is worth saying plainly.

**The broken sentence.** "The apps here are also minimal, where the harness's carry zod" is a dangling possessive with no noun. Unparseable. Rewritten to say which side imports zod.

**The halving claim.** Right, and the fix took two attempts. "Both Bun figures roughly halved on Bun 1.4, from 54.8 ms and 28.7 ms" does not square with the refreshed table - 28.7 against dunx's 46.2 is an increase, and the pairing reads inverted against the rows above. My first replacement said they "dropped by about a third", which is a fraction I cannot verify either: the old Bun is not measurable from here and I was inventing a number to replace a wrong one. So the historical claim is gone entirely and what remains is what the current run supports - **1.93x, read the ratio rather than the milliseconds**, since both absolutes move with the Bun release and have moved in both directions.

**The I/O exclusion**, both comments. "The harness does not measure ... anything with I/O" was made false by the `io` row **this same branch added two screens above it**. It now states what `io` covers - one Redis `GET`, one Postgres `SELECT` - and what it still does not: no filesystem, no upstream HTTP, nothing about a pool under contention. The "rounding error next to one query" sentence that followed is no longer an assertion either, since `io` measures it at 0.3%.

**And a fifth you did not flag:** the same section says a CLI "will feel the ~43 ms", which is the old figure. Now ~46.

`bun run ci` 13/13, corpus regenerated.

Worth noting how this one surfaced. The contradiction between the two documents was only visible to something reading both, and no gate compares them - which is the point I raised in the pull request body about those figures being hand-transcribed and unguarded. This round is the same class of problem twice more.

docs/guide/01-introduction.md
```diff
@@ -175,9 +185,13 @@ ajv, came in under the parse, so no throughput argument separates them. Pick on
 API, error quality and ecosystem.
 
 The harness does not measure absolute capacity, concurrency beyond one process,
```
dunxonu, line 187: The harness's "does not measure" disclaimer silently drops "memory" while correctly replacing the "anything with I/O" item with the new io-row explanation, leaving the benchmark's lack of memory measurement undocumented.

**How it fails:** A reader relies on this list to know what the throughput/startup harness covers; the harness still reports no memory figures (only ms and req/s), but the disclaimer no longer says so, understating what the numbers omit.

dunxonu (commented):

petarzarkov:
Both addressed in `416746e`, and I disagreed with half of one.

**Scoping the io conclusion** - agreed, and it is the same overclaim I have now made three rounds running. "In an application that talks to Postgres, every difference in the table above is rounding error next to one query" turned one Redis `GET` and one Postgres `SELECT` into a claim about every application. It now says what it is: 0.3% on that row, a result about this workload rather than a law, with a heavier query moving it further toward nothing and a service-free endpoint being the `plaintext` row instead.

**Restoring "memory" to the does-not-measure list** - not done, because it is no longer true. The premise is that "the harness still reports no memory figures (only ms and req/s)"; it has reported resident set and CPU per request since the `io` scenario landed, and `results/latest.json` carries a `resources` entry per subject per scenario. Putting "memory" back would reintroduce exactly the class of error the previous round removed - a disclaimer contradicting a measurement two screens away.

What *was* true is that this page showed neither. So it now gives the two figures a reader would want: **51.3 MiB at boot against 34.6, and 71.9 under the `io` row against 60.5**.

One process note, since it is the pattern this pull request keeps hitting: I first wrote that second pair as "72 against 63" from memory, then checked `results/latest.json` before pushing. It was 60.5. Same habit that produced the three claims you have already caught here - the difference this time is only that I checked first.

`bun run ci` 13/13, corpus regenerated.

docs/guide/01-introduction.md
```diff
@@ -175,9 +185,18 @@ ajv, came in under the parse, so no throughput argument separates them. Pick on
...
-anything with I/O, memory, behaviour under sustained load, TLS, HTTP/2, websockets
-or streaming. In an application that talks to Postgres, every difference in the
-table above is rounding error next to one query.
+behaviour under sustained load, TLS, HTTP/2, websockets or streaming. Its only
+I/O is the `io` row's one Redis `GET` and one Postgres `SELECT`; no filesystem, no
+upstream HTTP, and nothing about a pool under contention.
+
+It does measure resident set and CPU per request, which this page does not
```
dunxonu, line 192: New sentence claims the page does not reproduce resident-set/CPU numbers, then immediately reproduces the resident-set numbers in the very next sentence, and never gives the promised CPU-per-request numbers at all.

**How it fails:** A reader hits 'It does measure resident set and CPU per request, which this page does not reproduce.' followed by 'dunx boots at 51.3 MiB against raw `Bun.serve`'s 34.6, and peaks at 71.9 against 60.5 while serving the `io` row.' The RSS figures are reproduced right after being called unreproduced, and no CPU-per-request figure is given anywhere on the page despite being named as measured — the claim is self-contradicting and the CPU half is an unfulfilled promise.

dunxonu (commented):

PR #83, opened 2026-09-11
fix: two security defects found in a framework review

petarzarkov:
Two findings from a security pass over `packages/*` and `tools/*`. Each commit
carries a test that fails without its fix.

**`commands: false` did nothing.** `readOnlyMode` is a per-queue bull-board
adapter option, not a `UIConfig` key, so a read-only dashboard mount served every
mutating queue route: obliterate, clean, retry, and `POST .../add`, which injects
an arbitrary job name and payload into the app's workers. The hand-written
restatement of `BullMQAdapter` had erased the constructor's options parameter,
and the old test asserted the string appeared in the page.

**A `list` glob escaped the storage root.** `LocalStorage.list` checked `prefix`
and not `glob`; `Bun.Glob.scan` walks `..` and ignores `cwd` for an absolute
pattern, so `?glob=/etc/*` enumerated the host through a route that refuses
`?key=../..`. `examples/full` passed the query string in unchecked and
`@dunx/create-app` vendors that folder. Filenames only - `read` and `stat` go
through `resolveWithin`.

`bun run ci` green, 13/13.

Not done here: the example's dashboard still runs with `commands: true`, so the
Rule 4 coverage for the first fix is the package suite, which exercises real
bull-board against a real `Bun.serve`. Adding a board probe to `examples/full`
would make its dashboard demo open a broker socket, which that panel avoids on
purpose.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented file-listing glob patterns from accessing paths outside the configured storage directory, including absolute, parent-directory, and Windows-style traversal patterns.
  - Invalid traversal patterns now return a clear 400 Bad Request response.
  - Limited file-listing prefix and glob parameters to 200 characters.
  - Preserved support for valid filenames containing consecutive dots and colons.
  - Enforced read-only mode by rejecting queue-modifying requests with a 405 response.

### DIFF
--- a/packages/infra/src/files/local.ts
+++ b/packages/infra/src/files/local.ts
@@ -114,7 +114,7 @@ export class LocalStorage extends Storage {
     // Checked like the prefix above it, or the glob walks out of the root the
     // prefix was just held inside. S3 needs no equivalent: there the glob only
     // filters keys the bucket already returned.
-    const glob = new Bun.Glob(assertGlobWithin(options?.glob ?? '**/*'));
+    const glob = new Bun.Glob(assertGlobWithin(cwd, options?.glob ?? '**/*'));
 
     let yielded = 0;
     try {
@@ -124,9 +124,14 @@ export class LocalStorage extends Storage {
         dot: true,
         onlyFiles: true,
       })) {
+        const key = `${base}${toPosix(relative)}`;
+        // The pattern was checked, and so is every path it produced - so
+        // containment holds however `Bun.Glob` expands a brace or a bracket,
+        // rather than because a measurement said today's expansion is contained.
+        resolveWithin(this.#root, key);
         if (yielded >= limit) return;
         yielded += 1;
-        yield { key: `${base}${toPosix(relative)}` };
+        yield { key };
       }
     } catch (error) {
       // An absent directory lists as empty, matching what S3 does for a prefix
--- a/packages/infra/src/files/local.ts
+++ b/packages/infra/src/files/local.ts
@@ -2,6 +2,7 @@ import { resolve } from 'node:path';
 import type { BunFile } from 'bun';
 import { FileNotFoundError, UnsupportedOperationError } from './errors.js';
 import {
+  assertGlobWithin,
   guardMissing,
   isMissing,
   resolveDirWithin,
@@ -110,7 +111,10 @@ export class LocalStorage extends Storage {
     const cwd = resolveDirWithin(this.#root, prefix);
     const limit = options?.limit ?? Infinity;
     const base = prefix === '' ? '' : `${toPosix(prefix).replace(/\/+$/, '')}/`;
-    const glob = new Bun.Glob(options?.glob ?? '**/*');
+    // Checked like the prefix above it, or the glob walks out of the root the
+    // prefix was just held inside. S3 needs no equivalent: there the glob only
+    // filters keys the bucket already returned.
+    const glob = new Bun.Glob(assertGlobWithin(options?.glob ?? '**/*'));
 
     let yielded = 0;
     try {
--- a/packages/infra/src/files/local.ts
+++ b/packages/infra/src/files/local.ts
@@ -112,8 +112,8 @@ export class LocalStorage extends Storage {
     const limit = options?.limit ?? Infinity;
     const base = prefix === '' ? '' : `${toPosix(prefix).replace(/\/+$/, '')}/`;
     // Checked like the prefix above it, or the glob walks out of the root the
-    // prefix was just held inside. S3 needs no equivalent: there the glob only
-    // filters keys the bucket already returned.
+    // prefix was just held inside (docs/bun-apis.md). S3 needs no equivalent:
+    // there the glob only filters keys the bucket already returned.
     const glob = new Bun.Glob(assertGlobWithin(cwd, options?.glob ?? '**/*'));
 
     let yielded = 0;
--- a/packages/infra/src/files/path.ts
+++ b/packages/infra/src/files/path.ts
@@ -48,12 +48,16 @@ export const resolveDirWithin = (root: string, prefix: string): string =>
  * A `list` pattern, checked before `Bun.Glob` is constructed from it. `scan`
  * walks a `..` segment and ignores `cwd` outright for an absolute pattern, so a
  * checked `prefix` buys nothing while the glob beside it can name the whole
- * filesystem. Segments, not a substring: an inner `a..b` still lists.
+ * filesystem.
+ *
+ * The same `checkWithin` every key goes through, so there is one definition of
+ * "outside the root" rather than a second one that drifts. A wildcard is an
+ * ordinary path character to `resolve`, and the root itself is the pattern's
+ * own directory. What this cannot see is how `Bun.Glob` expands a brace or a
+ * bracket, so `LocalStorage.list` checks the paths it produced as well.
  */
-export const assertGlobWithin = (glob: string): string => {
-  if (hasParentSegment(glob) || /^([/\\]|[A-Za-z]:)/.test(glob)) {
-    throw new PathTraversalError(glob);
-  }
+export const assertGlobWithin = (root: string, glob: string): string => {
+  checkWithin(root, glob, true);
   return glob;
 };
 
--- a/packages/infra/src/files/path.ts
+++ b/packages/infra/src/files/path.ts
@@ -45,6 +45,19 @@ export const resolveDirWithin = (root: string, prefix: string): string =>
   checkWithin(root, prefix, true);
 
 /**
+ * A `list` pattern, checked before `Bun.Glob` is constructed from it. `scan`
+ * walks a `..` segment and ignores `cwd` outright for an absolute pattern, so a
+ * checked `prefix` buys nothing while the glob beside it can name the whole
+ * filesystem. Segments, not a substring: an inner `a..b` still lists.
+ */
+export const assertGlobWithin = (glob: string): string => {
+  if (hasParentSegment(glob) || /^([/\\]|[A-Za-z]:)/.test(glob)) {
+    throw new PathTraversalError(glob);
+  }
+  return glob;
+};
+
+/**
  * An object key with leading slashes and duplicate separators removed. `..` is
  * rejected outright rather than collapsed: an S3 key is opaque, so a caller who
  * wrote `..` meant a path, and under a configured prefix that would escape it.
--- a/docs/bun-apis.md
+++ b/docs/bun-apis.md
@@ -764,6 +764,39 @@ locally, `NetworkSink` for S3), preceded by an empty `Bun.write` to create paren
 and truncate. On 1.4.1 the local half is `Bun.write(path, stream)`; the S3 half is
 still a `NetworkSink`, because `S3Client`'s write takes no `ReadableStream`.
 
+### `Bun.Glob.scan` - `cwd` is where a pattern starts, not a boundary
+
+`scan({ cwd })` resolves the pattern against `cwd` and follows it wherever it
+leads. An absolute pattern ignores `cwd` entirely. Measured on 1.4.2 rev
+`744846f84`, one run, against a root holding `ok/a.txt` and `sub/s.txt` with a
+`secret/creds.env` beside it:
+
+| Pattern              | Yields                                           |
+| -------------------- | ------------------------------------------------ |
+| `**/*`               | `sub/s.txt`, `ok/a.txt`                          |
+| `../secret/*`        | `../secret/creds.env`                            |
+| `sub/../../secret/*` | `sub/../../secret/creds.env`                     |
+| `/etc/hostn*`        | `/etc/hostname`                                  |
+| `{ok,../secret}/*`   | nothing, and that includes the `ok/a.txt` branch |
+| `{.,..}/secret/*`    | nothing                                          |
+| `[.][.]/secret/*`    | nothing                                          |
+| `..?/secret/*`       | nothing                                          |
+
+A `..` segment and a leading `/` escape. The forms that spell a parent segment
+without writing one as a segment do not, and a brace group holding a `..` branch
+matches nothing at all rather than expanding to the branch that would have
+resolved. `dot: true` and `onlyFiles: true` change none of it.
+
+An escape stays visible in the result: a relative match keeps its `../` prefix
+and an absolute pattern yields absolute paths, so nothing is normalised away
+before the caller sees it.
+
+`LocalStorage.list` takes its `glob` from a caller, which in an app is usually a
+query parameter, so it checks the pattern through the same `checkWithin` every
+key goes through and checks each path the scan produced. The second check is
+what keeps containment off the last four rows of that table: they are today's
+expansion, not a promise.
+
 ### `Bun.S3Client` - the undocumented surface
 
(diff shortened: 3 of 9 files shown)


packages/infra/src/files/path.ts
```diff
@@ -44,6 +44,19 @@ export const resolveWithin = (root: string, key: string): string =>
 export const resolveDirWithin = (root: string, prefix: string): string =>
   checkWithin(root, prefix, true);
 
+/**
+ * A `list` pattern, checked before `Bun.Glob` is constructed from it. `scan`
+ * walks a `..` segment and ignores `cwd` outright for an absolute pattern, so a
+ * checked `prefix` buys nothing while the glob beside it can name the whole
+ * filesystem. Segments, not a substring: an inner `a..b` still lists.
+ */
+export const assertGlobWithin = (glob: string): string => {
```
dunxonu: assertGlobWithin only rejects a literal `..` segment, but Bun.Glob supports brace-alternation and bracket expressions that can spell an equivalent parent segment without the string `..` appearing as its own `/`-delimited segment.

**How it fails:** A request like `GET /files?glob=%7Bok%2C..%2Fsecret%7D%2F*` (glob=`{ok,../secret}/*`) splits into segments `{ok,..`, `secret}`, `*` — none exactly equal to `".."` — so `hasParentSegment` returns false and the absolute-path regex also doesn't match, letting the string through unchanged. `Bun.Glob("{ok,../secret}/*").scan({cwd})` then expands the brace and walks `../secret` relative to `cwd`, listing files outside the storage root that the PR's own fix and tests (`local.test.ts:314`, `negative.test.ts`) were written to block for the literal `../*` form.

packages/infra/src/files/path.ts
```diff
@@ -44,6 +44,19 @@ export const resolveWithin = (root: string, key: string): string =>
...
+  if (hasParentSegment(glob) || /^([/\\]|[A-Za-z]:)/.test(glob)) {
```
dunxonu: The absolute-path detector in assertGlobWithin (`/^([/\\]|[A-Za-z]:)/`) is a hand-rolled regex duplicating the generic `resolve`+`isInside` boundary check already in the same file (`checkWithin`, lines 11-29), instead of reusing it.

**How it fails:** The two absolute-path detectors can drift: if `checkWithin`/`isInside` is later extended to catch a new escaping form (a UNC path, a `file://` URL, a drive-relative path), only `resolveWithin`/`resolveDirWithin` gains the fix while `assertGlobWithin`'s separate regex stays exploitable via that form until someone remembers to update it too.

dunxonu: The `[A-Za-z]:` half of the absolute-path regex rejects any glob starting with a single ASCII letter followed by a colon, even on POSIX where such a prefix is not an absolute path and previously worked with the unrestricted `new Bun.Glob(...)` call.

**How it fails:** A relative glob like `a:backup.txt` or `x:*.log` (a legal POSIX filename fragment containing a colon) now throws `PathTraversalError` and turns into a 400 from `files.controller.ts`, where before this PR it was accepted and listed normally - a behavior regression for a pattern that never escaped the storage root.

dunxonu (commented):

packages/infra/src/files/path.ts
petarzarkov: Measured before changing anything, and it does not reproduce on Bun 1.4.2. A brace group holding a `..` branch matches **nothing at all** - it does not even match the branch that resolves:

```
cwd=<root>, root/ok/a.txt exists, ../secret/creds.env exists
"{ok,../secret}/*"  -> []      (not even ok/a.txt)
"{../secret,ok}/*"  -> []
"{ok,..}/*"         -> ["ok/a.txt"]
"[.][.]/secret/*"   -> []
"{.,..}/secret/*"   -> []
"?./secret/*"       -> []
```

So `hasParentSegment` returning false was not the hole it looks like: the expansion never walks out.

That said, the objection is right that the guarantee should not rest on how Bun expands a pattern, and a measurement of today's behaviour is a weak thing to hang containment on. `list` now checks every path the scan produced, not just the pattern going in (38e82e2, `local.ts:127-131`), so a change in brace or bracket expansion is a refusal rather than a leak. `local.test.ts` pins the brace case with the Bun version named.

packages/infra/src/files/path.ts
petarzarkov: Agreed, and fixed in 38e82e2. `assertGlobWithin` is now `checkWithin(root, glob, true)` - the same call `resolveWithin` and `resolveDirWithin` make, so there is one definition of "outside the root" and the drift you describe cannot happen. A wildcard is an ordinary path character to `resolve`, and `allowRoot` is true because a glob names its own directory.

petarzarkov: Correct, and it is gone with the regex in 38e82e2. Delegating to `checkWithin` gets the platform question right for free: on POSIX `resolve(root, 'c:notes.txt')` lands inside the root and lists, on Windows the same string is drive-relative and resolves out. `local.test.ts` covers `glob: 'c:*'` listing `c:notes.txt` so the regression stays fixed.

dunxonu: Confirmed, the fix holds. assertGlobWithin now delegates to checkWithin (packages/infra/src/files/path.ts:57-60) instead of the hand-rolled `[A-Za-z]:` regex, so a POSIX-relative pattern like `c:*` resolves via `resolve(root, key)` and lands inside the root instead of being rejected outright. local.test.ts:343-346 pins `glob: 'c:*'` listing `c:notes.txt`, so the regression is closed and regression-tested. Nothing further from me here.

packages/infra/src/files/path.ts
```diff
@@ -44,6 +44,23 @@ export const resolveWithin = (root: string, key: string): string =>
 export const resolveDirWithin = (root: string, prefix: string): string =>
   checkWithin(root, prefix, true);
 
+/**
+ * A `list` pattern, checked before `Bun.Glob` is constructed from it. `scan`
```
dunxonu, line 48: The runtime behavior this fix depends on (Bun.Glob's `scan` walking `..` segments, ignoring `cwd` for an absolute pattern, and matching nothing for a brace group containing a `..` branch) was measured on Bun 1.4.2 but never recorded in docs/bun-apis.md.

**How it fails:** CLAUDE.md states: "docs/bun-apis.md is not exhaustive - Bun ships undocumented APIs, and several documented ones misbehave. Probe the runtime before concluding anything, and record what you verify there." This PR probes exactly that (comments in local.ts line 114 and local.test.ts line 308-311 say "Measured on Bun 1.4.2"), but the measurement lives only in source/test comments. A future Bun upgrade that changes brace-expansion or absolute-pattern handling has no central place documenting the assumption `assertGlobWithin`'s design and the 'contains what a brace pattern expands to as well' test both rely on, so the regression would be caught only if that specific test still encodes today's behavior.

dunxonu (commented):

packages/infra/src/files/path.ts
petarzarkov: Valid, and it is the convention rather than a nice-to-have: CLAUDE.md says to record what you probe there, and I had left the measurement in source comments only.

Recorded in f2c122e, `docs/bun-apis.md`, under `### Bun.Glob.scan - cwd is where a pattern starts, not a boundary`. Every row is one run on 1.4.2 rev `744846f84` rather than the two separate probes the comments came from:

| Pattern | Yields |
| --- | --- |
| `../secret/*` | `../secret/creds.env` |
| `sub/../../secret/*` | `sub/../../secret/creds.env` |
| `/etc/hostn*` | `/etc/hostname` |
| `{ok,../secret}/*` | nothing, including the `ok/a.txt` branch |
| `{.,..}/secret/*`, `[.][.]/secret/*`, `..?/secret/*` | nothing |

The section closes on which half of the fix rests on which row: the pattern check covers the first three, and the per-entry check is what keeps containment off the last four, since those are today's expansion rather than a promise. `local.ts` and `local.test.ts` now point at the section instead of restating it.

dunxonu: Confirmed, f2c122e recorded it correctly: docs/bun-apis.md now has the "Bun.Glob.scan - cwd is where a pattern starts, not a boundary" section with the full table (the eight patterns, `..` segment, leading `/`, the brace/bracket/`?` forms matching nothing), one run on 1.4.2 rev `744846f84`, and it states which rows the pattern check covers versus which rely on the per-entry check. `local.ts` and `local.test.ts` point at it instead of restating it, which is exactly what [CLAUDE.md](https://github.com/petarzarkov/dunx/blob/main/CLAUDE.md) asks for ("Probe the runtime before concluding anything, and record what you verify there"). Closing this out.

packages/infra/src/files/local.ts
```diff
@@ -110,7 +111,10 @@ export class LocalStorage extends Storage {
     const cwd = resolveDirWithin(this.#root, prefix);
     const limit = options?.limit ?? Infinity;
     const base = prefix === '' ? '' : `${toPosix(prefix).replace(/\/+$/, '')}/`;
-    const glob = new Bun.Glob(options?.glob ?? '**/*');
+    // Checked like the prefix above it, or the glob walks out of the root the
+    // prefix was just held inside (docs/bun-apis.md). S3 needs no equivalent:
+    // there the glob only filters keys the bucket already returned.
+    const glob = new Bun.Glob(assertGlobWithin(cwd, options?.glob ?? '**/*'));
```
dunxonu, line 117: assertGlobWithin checks the glob against `cwd` (the resolved prefix directory), not the storage root, so an absolute or dotted glob that legitimately resolves inside the storage root but outside the current `prefix` is rejected as a traversal even though it never escapes the root.

**How it fails:** root=/data, prefix='reports' (cwd=/data/reports), glob='/data/other/*' where /data/other is a real subdirectory of the configured storage root: `checkWithin(cwd, glob, true)` computes `resolve(cwd, glob)` = '/data/other/*' (absolute pattern ignores cwd), which does not start with cwd+sep, so it throws PathTraversalError and the request 400s even though '/data/other' sits inside the true root and would be safe to list.

packages/infra/src/files/local.ts
```diff
@@ -120,9 +124,14 @@ export class LocalStorage extends Storage {
         dot: true,
         onlyFiles: true,
       })) {
+        const key = `${base}${toPosix(relative)}`;
+        // The pattern was checked, and so is every path it produced - so
+        // containment holds however `Bun.Glob` expands a brace or a bracket,
+        // rather than because a measurement said today's expansion is contained.
+        resolveWithin(this.#root, key);
```
dunxonu, line 131: resolveWithin is invoked unconditionally on every single entry a listing yields, even though the code's own comment says the only gap it covers is brace/bracket expansion in the pattern.

**How it fails:** A listing of thousands of files under a plain glob like `**/*` (no `{`/`[` constructs) pays a full `path.resolve` + containment check per entry for a guarantee that, per the adjoining comment, is only ever needed for brace/bracket patterns; a maintainer profiling a large `list()` call would find this check as avoidable per-item overhead rather than a one-time cost.

packages/infra/src/files/path.ts
```diff
@@ -44,6 +44,23 @@ export const resolveWithin = (root: string, key: string): string =>
...
+ * checked `prefix` buys nothing while the glob beside it can name the whole
+ * filesystem.
+ *
+ * The same `checkWithin` every key goes through, so there is one definition of
+ * "outside the root" rather than a second one that drifts. A wildcard is an
+ * ordinary path character to `resolve`, and the root itself is the pattern's
+ * own directory. What this cannot see is how `Bun.Glob` expands a brace or a
+ * bracket, so `LocalStorage.list` checks the paths it produced as well.
+ */
+export const assertGlobWithin = (root: string, glob: string): string => {
```
dunxonu, line 59: assertGlobWithin is a thin wrapper that calls checkWithin and then returns its own input unchanged, adding a new exported name for what call sites could get by calling checkWithin (via resolveDirWithin) directly.

**How it fails:** The only call site (local.ts:117) discards the return value's relationship to its input, using it purely for validation; a future reader has to open path.ts to learn `assertGlobWithin(root, glob)` is exactly `resolveDirWithin(root, glob)` plus `return glob`, which is one indirection with no behavioral payoff.

dunxonu (commented):

packages/infra/src/files/local.ts
petarzarkov: Mechanically right, and I checked what the refused case actually did before any of this. It was already broken.

`root=/data`, `prefix='reports'`, `glob='<root>/other/*'` on the pre-fix code yields:

```
reports//tmp/.../c1root/other/b.csv
```

the prefix concatenated onto an absolute path. That key does not round-trip: `read()` resolves it to `<root>/reports/tmp/...`, which does not exist. So the listing "that would be safe" produced keys nothing could use.

The contract also rules it out. `ListOptions.glob` is documented at `packages/infra/src/files/storage.ts:37` as "Glob pattern, matched against keys relative to `prefix`", so a glob naming a sibling subtree of the root is not a use the API offers - `prefix: 'other'` is. Checking against `cwd` is what enforces that sentence, and a 400 naming the refusal beats a key that silently does not resolve.

Leaving it as is.

packages/infra/src/files/path.ts
petarzarkov: Accurate: it is `resolveDirWithin(root, glob)` plus `return glob`.

Keeping it, for the reason the name exists. The call site reads `assertGlobWithin(cwd, options?.glob ?? '**/*')`, and `resolveDirWithin(cwd, glob)` with the result discarded would read as a mistake at a glance - a reviewer would ask why a resolved path is being thrown away. The doc comment on it is also where the `Bun.Glob` reasoning lives, so the indirection carries the one thing a future reader needs.

Noted as a judgement call rather than a defect, so say the word if you would rather see the direct call with a comment instead.

packages/infra/src/files/local.ts
petarzarkov: This one was right and I had waved it through on intuition. Measured on 5,000 files under one root, medians of seven runs, same script:

| Listing | Median |
| --- | --- |
| `**/*`, every entry checked | 7.0 ms |
| `**/*`, no entry checked | 3.7 ms |
| the check alone, 5,000 calls | 1.8 ms |

So a plain listing was paying roughly double for a guarantee it could not need.

Fixed in #84: the check now runs only for a pattern carrying `{`, `[` or `(`. A wildcard matches one entry name and never a separator, so `*`, `**` and `?` cannot spell a parent segment once the pattern check has refused a literal `..` and an absolute path, and extglob is covered because `@(`, `+(`, `?(` and `!(` all carry a paren. `local.test.ts` pins that assumption with a listing over `*/*`, `**/*`, `?./*` and `..?/*`, and the table is in `docs/bun-apis.md` beside the scan behaviour.

packages/infra/src/files/local.ts
dunxonu: Fair, the finding was wrong. I didn't check what the pre-fix behavior actually did for that case, and you're right it never worked: the key gets `prefix` prepended in front of an absolute path and doesn't round-trip through `read()`. Combined with the documented contract at packages/infra/src/files/storage.ts:37 ("matched against keys relative to `prefix`"), a glob reaching outside the current prefix isn't a supported case at all, so checking against `cwd` is the correct enforcement, not a bug. Dropping this.

PR #84, opened 2026-09-11
perf(infra): gate the per-entry glob containment check

petarzarkov:
Follow-up to the last review round on #83, which landed before three comments
arrived.

`LocalStorage.list` ran `resolveWithin` on every entry it yielded, for a gap its
own comment scoped to brace and bracket expansion. 5,000 files, medians of seven:
`**/*` took 7.0 ms with the check and 3.7 ms without.

It now runs only for a pattern carrying `{`, `[` or `(`. A wildcard matches one
entry name and never a separator, so `*`, `**` and `?` cannot spell a parent
segment once the pattern check has refused a literal `..` and an absolute path;
extglob is covered because `@(`, `+(`, `?(` and `!(` all carry a paren. A test
pins that assumption and the table sits in docs/bun-apis.md beside the scan
behaviour.

The other two comments from that round are answered on #83 rather than here: the
`cwd` one describes a case that produced a malformed key before any of this, and
the wrapper one is a naming call.

`bun run ci` green, 13/13.

## Summary by CodeRabbit

- **New Features**
  - Improved file listing safety by preventing glob patterns from traversing outside the intended directory.
  - Added support for validating complex glob patterns, including braces, brackets, parentheses, and backslashes.

- **Bug Fixes**
  - Backslash-based parent-directory patterns now return no results.
  - Wildcard-only patterns cannot access parent-directory entries.

- **Documentation**
  - Documented glob pattern behavior and the performance impact of containment checks.

### DIFF
--- a/docs/bun-apis.md
+++ b/docs/bun-apis.md
@@ -807,10 +807,26 @@ root, medians of seven runs:
 | the check alone, 5,000 calls | 1.8 ms |
 
 So it runs only for a pattern that can expand into a segment the pattern check
-did not see: one carrying `{`, `[` or `(`. A wildcard matches a single entry
-name and never a separator, so `*`, `**` and `?` cannot spell a parent segment
-on their own, and extglob is covered because `@(`, `+(`, `?(` and `!(` all carry
-a paren. `**/*` therefore pays nothing and `{a,../b}/*` pays for every entry.
+did not see: one carrying `{`, `[`, `(` or a backslash. A wildcard matches a
+single entry name and never a separator, so `*`, `**` and `?` cannot spell a
+parent segment on their own, and extglob is covered because `@(`, `+(`, `?(` and
+`!(` all carry a paren. `**/*` therefore pays nothing and `{a,../b}/*` pays for
+every entry.
+
+**A backslash is not an escape**, which is why it is on that list rather than
+handled by the pattern check. On 1.4.2 it matches a literal backslash in a name:
+
+```
+ok/*          -> ["ok/a.txt"]      \o\k/*        -> nothing
+dot.dir/*     -> ["dot.dir/f.txt"] dot\.dir/*    -> nothing
+\.\./secret/* -> nothing           ..            -> nothing
+```
+
+So `\.\./secret/*` looks for a directory named `\.\.` and finds none. The day
+Bun aligns with the engines that do treat `\.` as an escaped dot, that pattern
+becomes `../secret/*` while `hasParentSegment` still sees `['', '.', '.']`,
+because it splits on the backslash. Gating on the character costs a real pattern
+nothing and does not wait for that.
 
 ### `Bun.S3Client` - the undocumented surface
 
--- a/docs/bun-apis.md
+++ b/docs/bun-apis.md
@@ -797,6 +797,21 @@ key goes through and checks each path the scan produced. The second check is
 what keeps containment off the last four rows of that table: they are today's
 expansion, not a promise.
 
+That second check is charged per entry, and it is not free. 5,000 files under one
+root, medians of seven runs:
+
+| Listing                      | Median |
+| ---------------------------- | ------ |
+| `**/*`, every entry checked  | 7.0 ms |
+| `**/*`, no entry checked     | 3.7 ms |
+| the check alone, 5,000 calls | 1.8 ms |
+
+So it runs only for a pattern that can expand into a segment the pattern check
+did not see: one carrying `{`, `[` or `(`. A wildcard matches a single entry
+name and never a separator, so `*`, `**` and `?` cannot spell a parent segment
+on their own, and extglob is covered because `@(`, `+(`, `?(` and `!(` all carry
+a paren. `**/*` therefore pays nothing and `{a,../b}/*` pays for every entry.
+
 ### `Bun.S3Client` - the undocumented surface
 
 `prototype`: `delete`, `exists`, `file`, `list`, `presign`, `size`, `stat`,
--- a/packages/infra/src/files/local.ts
+++ b/packages/infra/src/files/local.ts
@@ -114,7 +114,12 @@ export class LocalStorage extends Storage {
     // Checked like the prefix above it, or the glob walks out of the root the
     // prefix was just held inside (docs/bun-apis.md). S3 needs no equivalent:
     // there the glob only filters keys the bucket already returned.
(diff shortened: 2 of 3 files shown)


packages/infra/src/files/local.ts
```diff
@@ -114,7 +114,12 @@ export class LocalStorage extends Storage {
     // Checked like the prefix above it, or the glob walks out of the root the
     // prefix was just held inside (docs/bun-apis.md). S3 needs no equivalent:
     // there the glob only filters keys the bucket already returned.
-    const glob = new Bun.Glob(assertGlobWithin(cwd, options?.glob ?? '**/*'));
+    const pattern = assertGlobWithin(cwd, options?.glob ?? '**/*');
+    const glob = new Bun.Glob(pattern);
+    // A wildcard matches one entry name and cannot introduce a separator, so
+    // after that check only an alternation construct can still spell a parent
+    // segment. Those pay for a per-entry check and `**/*` does not.
+    const expands = /[{[(]/.test(pattern);
```
dunxonu: The `expands` gate (`/[{[(]/.test(pattern)`) only recognizes brace/bracket/extglob syntax as capable of producing a `..` escape, but a backslash-escaped pattern like `\.\./secret/*` contains none of `{`, `[`, `(`, so the per-entry `resolveWithin` safety check is skipped for it even though glob engines conventionally treat `\.` as an escaped literal `.` and would resolve it the same as the unescaped `../secret/*` case that `docs/bun-apis.md` (line ~777) measures as an actual escape.

**How it fails:** A caller passes `glob: '\\.\\./secret/*'` (or any pattern using `\` to escape a `.` so the text never contains a literal `..` path segment). `hasParentSegment` in `packages/infra/src/files/path.ts` splits the string on `/` and `\` looking for a `..` segment, but `\.\.` splits into `['', '.', '.', ...]`, not `'..'`, so `assertGlobWithin` (path.ts:59-62) does not throw. `expands` in local.ts:122 is then `false` because the pattern has no `{`, `[`, or `(`, so the previously-unconditional `resolveWithin` check at local.ts:136 is skipped, and if `Bun.Glob` interprets the backslash as an escape (turning the pattern into the equivalent of `../secret/*`), the scan can yield a key outside `cwd`/root that is never checked and is handed straight to the caller.

dunxonu (commented):

packages/infra/src/files/local.ts
petarzarkov: The exploit does not reproduce, but the reasoning gap behind it is real, so the gate now carries a backslash too (e02cb21).

Measured on Bun 1.4.2, POSIX: a backslash in a pattern is **not** an escape, it matches a literal backslash in a name.

```
ok/*          -> ["ok/a.txt"]       \o\k/*        -> nothing
dot.dir/*     -> ["dot.dir/f.txt"]  dot\.dir/*    -> nothing
\.\./secret/* -> nothing            ..            -> nothing
```

`dot\.dir/*` missing a directory that `dot.dir/*` finds is the decisive pair: nothing is being unescaped. So `\.\./secret/*` looks for a directory literally named `\.\.`, finds none, and the scan never leaves cwd.

What holds regardless of that measurement is your point about the checks: `hasParentSegment` splits on the backslash and sees `['', '.', '.']`, so if Bun ever aligns with the engines that do unescape, that pattern becomes `../secret/*` and neither the pattern check nor the old gate would see it. One character in the character class closes it, costs a real pattern nothing (a pattern with a backslash on POSIX is hunting a literal backslash; on Windows it is a separator, where paying the check is right anyway), and does not wait for the Bun change.

`local.test.ts` covers `\.\./*`, `\.\./secret/*` and `.\./*`, and the escaping probe is recorded in `docs/bun-apis.md` beside the scan table.

dunxonu: Agreed, and thanks for measuring it rather than taking my reasoning at face value. The exploit as I described it doesn't hold: if backslash isn't an escape on Bun's Glob, `\.\./secret/*` never resolves to `../secret/*` in the first place, so there was no live containment bypass. What I got right was narrower than I framed it: `hasParentSegment` and the `expands` gate both blindly split/match on structural characters without accounting for a backslash, so the day Bun's Glob does start treating `\.` as an escape, the check would have silently missed it. Gating `expands` on the backslash too closes that forward-looking gap for free, which is exactly the right fix and matches [CLAUDE.md's guidance to probe the runtime rather than assume before concluding anything about an undocumented Bun surface](https://github.com/petarzarkov/dunx/blob/e02cb2190424b57d7721eff471d37697c3b4e0b9/CLAUDE.md#L142). e02cb21 plus the local.test.ts cases and the bun-apis.md entry look correct to me. Nothing further to add here.

dunxonu (commented):
The gate logic is verified sound. Given the small, well-documented, and measurement-backed scope of this diff, I don't find a correctness bug, and cleanup/altitude/conventions angles turn up nothing actionable either (no em-dashes, no CLAUDE.md violations, no reuse/simplification issues in the ~40 changed lines).

```json
[]
```

docs/bun-apis.md
```diff
@@ -797,6 +797,37 @@ key goes through and checks each path the scan produced. The second check is
...
+did not see: one carrying `{`, `[`, `(` or a backslash. A wildcard matches a
+single entry name and never a separator, so `*`, `**` and `?` cannot spell a
+parent segment on their own, and extglob is covered because `@(`, `+(`, `?(` and
+`!(` all carry a paren. `**/*` therefore pays nothing and `{a,../b}/*` pays for
+every entry.
+
+**A backslash is not an escape**, which is why it is on that list rather than
+handled by the pattern check. On 1.4.2 it matches a literal backslash in a name:
+
+```
```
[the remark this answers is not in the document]
petarzarkov: Skipping this one, with numbers rather than a preference.

`docs/bun-apis.md` has 25 bare opening fences against 6 ```` ```ts ```` and 1 ```` ```text ````. Every probe-output block in the file - the `AsyncLocalStorage` retention output, the `Bun.write` data-loss table, the `Bun.Glob` scan results two blocks above this one - is a bare fence. Tagging only the newest block makes it the odd one out among its own neighbours.

markdownlint also is not one of this repo's gates: no reference in `package.json`, `scripts/ci.ts` or `.github/workflows/ci.yml`, and `bun run ci` is green. So MD040 is not failing anything here.

If MD040 is worth being clean on, the change is all 25 fences plus markdownlint added to `PHASES` in `scripts/ci.ts` and to a job in `ci.yml` (the drift test requires both), which is its own PR rather than a line in this one. Happy to open that if wanted.

PR #87, opened 2026-09-11
feat(core): read config from YAML, TOML and JSON files

### LINKED ISSUE #79: Feature request: Support configuration files such as YAML by default
## Feature request: Support configuration files such as YAML by default

The current configuration system is primarily based on environment variables. I think it would be useful if Dunx provided built-in support for common configuration file formats as well.

For example:

```text
application.yml
application.yaml
application.json
application.toml
```

A typical configuration could look like:

```yaml
server:
  port: 3000

database:
  url: postgres://localhost:5432/app
  poolSize: 10

redis:
  url: redis://localhost:6379
```

### Why

For larger applications, using environment variables for all configuration can become difficult to maintain, especially for nested configuration.

Configuration files provide:

* Better organization for nested configuration
* Easier local development
* Easier configuration management
* Better readability
* A natural way to maintain environment-specific configuration

Ideally, the configuration system could support a flow such as:

```text
application.yml
      ↓
application-{env}.yml
      ↓
environment variables override
      ↓
schema validation
      ↓
ConfigService
```

For example:

```text
application.yml
application-development.yml
application-production.yml
```

Environment variables should still have the highest priority so that the configuration remains compatible with 12-factor deployment.

### Suggested API

Something along the lines of:

```ts
ConfigModule.forRoot({
  schema: configSchema,
  files: [
    'application.yml',
    `application-${env}.yml`,
  ],
})
```

Or an even simpler default convention:

```ts
ConfigModule.forRoot({
  schema: configSchema,
})
```

which automatically discovers supported configuration files.

The implementation does not necessarily need to couple the configuration module tightly to every parser. A small loader abstraction could allow YAML, JSON, TOML, and other formats to be supported cleanly.

I think this would make Dunx much more suitable for larger applications while keeping the existing environment-variable-first approach for production deployments.

petarzarkov:
Closes #79.

`ConfigModule.forRoot({ files })` merges the listed files under the environment. `Bun.YAML`, `Bun.TOML` and `JSON.parse` are all native, so `@dunx/core` keeps zero dependencies.

Env override is by exact key name only. No `DATABASE__POOLSIZE` → `database.poolSize` convention: that needs a separator, a case rule and a coercion rule, which is the schema DSL this module does not have.

`examples/full` moves its seed user list into `application.yml` with no schema default, so the file is load-bearing and the suite fails if it is not read.

## Summary by CodeRabbit

- **New Features**
  - Added support for loading YAML, TOML, and JSON configuration files.
  - Configuration files are parsed with types preserved, deep-merged in order, and overridden by environment values.
  - Missing files are skipped, while invalid or unsupported files produce clear startup errors.
  - Added generated application configuration modules and file-based configuration examples.

- **Documentation**
  - Expanded configuration guidance, including file loading, layering, overrides, validation, and secrets.


packages/core/src/config/files.ts
```diff
@@ -0,0 +1,98 @@
...
+
+const isPlainObject = (value: unknown): value is ConfigValues =>
+  typeof value === 'object' && value !== null && !Array.isArray(value);
+
+/** Later wins, per key: two objects merge, anything else replaces. */
+const merge = (base: ConfigValues, overlay: ConfigValues): ConfigValues => {
+  const out: ConfigValues = { ...base };
+  for (const [key, value] of Object.entries(overlay)) {
+    const existing = out[key];
+    out[key] =
```
dunxonu, line 36: The recursive `merge()` used to combine config files has no guard against `__proto__`, so a config file with that top-level key repoints the merged object's prototype instead of setting a data property.

**How it fails:** A YAML/TOML/JSON config file containing `__proto__: {isAdmin: true}` causes `out["__proto__"] = value` (files.ts:15) to invoke the inherited `Object.prototype.__proto__` setter rather than creating an own key, so the merged `ConfigValues` object silently inherits arbitrary attacker- or typo-supplied properties (e.g. `config.values.isAdmin` reads `true` even though no such key was ever legitimately set), and this repeats through every nested `merge()` call for any nested `__proto__` key at any depth.

packages/core/src/config/files.ts
```diff
@@ -0,0 +1,98 @@
...
+      if (!(await file.exists())) continue;
+
+      const text = await file.text();
+      let parsed: unknown;
+      try {
+        parsed = parse(path, text);
+      } catch (cause) {
+        // The parser names the line but not the file: it was handed a string.
+        throw new ConfigError(
+          `Config file "${path}" did not parse: ${(cause as Error).message}`,
```
dunxonu: `load()`'s catch-all around `parse()` re-wraps `parse()`'s own "no parser" `ConfigError` into a second, contradictory message.

**How it fails:** Given a file `app.ini`, `parse()` throws `Config file "app.ini" has no parser. Use .yml, .yaml, .toml or .json.` (files.ts:33), which is caught by the try/catch in `load()` (files.ts:73-80) and rethrown as `Config file "app.ini" did not parse: Config file "app.ini" has no parser. Use .yml, .yaml, .toml or .json.` - the filename appears twice and "did not parse" contradicts "has no parser", making the boot-failure message confusing for whoever is debugging a misnamed config file.

dunxonu: `(cause as Error).message` assumes every thrown parse error is an `Error` instance, so a non-Error throw renders as "undefined" in the boot failure message.

**How it fails:** If `Bun.YAML.parse` or `Bun.TOML.parse` ever throws a plain string or non-Error object for a malformed file, the rethrown `ConfigError` reads `Config file "x.yml" did not parse: undefined`, discarding whatever detail the parser provided about where the file is malformed.

dunxonu (commented):

petarzarkov:
All six review findings addressed in 2f8bbb1. Notes on each, since two were wrong about severity in opposite directions.

**Prototype pollution (`merge`)** - real, and I reproduced it before fixing: a file with `__proto__: { isAdmin: true }` had `config.get('isAdmin')` answer `true`. `__proto__` is now dropped at every depth, with tests at the top level and nested.

**Double-wrapped "no parser" message** - real. The parser is now chosen before the `try`, so the two messages cannot nest. The test asserts the second message is absent rather than just that the first is present.

**Non-`Error` throw** - taken, because reading it through an `instanceof` check also removes a cast.

**`undefined` in `source`** - real and taken. An unset variable is absent from `Bun.env` entirely, so a hand-written `source` has to mean the same thing; `{ PORT: undefined }` no longer wipes a file's value. Regression test added.

**Discriminate `validate` on `files`** - taken, as two `forRoot` signatures rather than a four-way union, so contextual typing of an unannotated `validate` still works. `files` moved out of `ConfigModuleBase` into `ConfigFilesOption`, because the general options type otherwise matched neither signature.

No compile-time negative test: the repo has no `@ts-expect-error` harness and adding one for this alone is more machinery than the case warrants.

**Guide declaration block** - was stale on both the generic and `files`.

packages/core/src/config/files.ts
```diff
@@ -0,0 +1,108 @@
+import { resolve } from 'node:path';
+import { ConfigError } from './service.js';
+
+/** A source whose values are already typed, as a parsed file's are. */
+export type ConfigValues = Record<string, unknown>;
+
+const isPlainObject = (value: unknown): value is ConfigValues =>
```
dunxonu: `isPlainObject` only checks `typeof === 'object' && !Array.isArray`, so a `Date` (or any class instance) is treated as mergeable, causing `merge()` to silently collapse it into `{}` when two config files set the same key to a date-like value.

**How it fails:** `application.yml` sets `released: 2024-01-01` and `application-production.yml` overlays `released: 2024-06-01`. If `Bun.YAML.parse` (or `Bun.TOML.parse`, which has a mandatory native date-time type per spec) resolves either to a JS `Date`, `merge()`'s `isPlainObject(existing) && isPlainObject(value)` branch is taken for both, and `merge(existingDate, valueDate)` iterates `Object.entries(date)` (zero own enumerable properties), replacing the key's value with `{}` instead of the intended date. `packages/http/src/client/json.ts` already had to fix this exact class of bug for the same reason (its own doc comment: 'the reference's typeof === object ... answered true for a Date'), and `files.test.ts` has no test covering a date-like value in any format.

docs/guide/12-configuration.md
```diff
@@ -31,23 +31,81 @@ its path, rather than whatever shape the library throws.
...
 which keys are wrong. `schema` is the same step handed to a Standard Schema
 instead, and its issues become a `ConfigError` naming each path.
 
-That is the whole contract. There is no `envFilePath`, no `load: [...]`, no
+That is the whole contract, plus `files`. There is no `envFilePath` and no
 `expandVariables`.
 
+## Configuration files
+
+`files` reads YAML, TOML and JSON and merges them under the environment:
```
dunxonu, line 58: The guide's own `files` example uses paths resolved against `process.cwd()` (`ConfigFiles`'s default), with no note that this is cwd-dependent, while `examples/full` deliberately avoids that footgun by building `configFiles` from `import.meta.dir`.

**How it fails:** A reader copies `files: ['application.yml', 'application-${NODE_ENV}.yml']` verbatim into a module that gets loaded from a different working directory (a compiled binary, a process manager, a forked worker with a changed cwd). Because `ConfigFiles` silently skips a missing file by design, the app boots with an empty/partial config instead of failing loudly - the opposite of `ConfigModule`'s stated goal - and nothing in the guide warns that `import.meta.dir`-based paths are the safer choice `examples/full` itself uses.

dunxonu (commented):

petarzarkov:
Both new findings were real and are fixed.

**`isPlainObject` and dates** - right about the class of bug, but the mechanism is worse than `Date`. TOML has a first-class date type and `Bun.TOML.parse` returns `Temporal.PlainDate` / `Temporal.Instant`, which are objects with zero own enumerable keys. Reproduced before fixing: merging `released = 2024-01-01` under `released = 2024-06-01` gave `{"released":{}}`. YAML is unaffected, since Bun parses its dates as primitive strings. Rather than write a second predicate, the correct prototype check moved from `@dunx/http` (where it was neither used internally nor public) into `@dunx/core`, per Rule 2.

**Empty JSON file** - real. The emptiness check now runs before the parser, so absent and empty mean the same thing in all three formats. Test covers `.json`, `.toml` and whitespace-only `.yaml`, not just the YAML case it had.

**cwd-relative paths in the guide** - real. A relative path resolves against `process.cwd()` and a missing file is skipped by design, so a copied example boots on defaults instead of failing. The guide now shows the `import.meta.dir` form `examples/full` already used.

packages/core/src/config/files.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/** Later wins, per key: two objects merge, anything else replaces. */
+const merge = (base: ConfigValues, overlay: ConfigValues): ConfigValues => {
+  const out: ConfigValues = { ...base };
+  for (const [key, value] of Object.entries(overlay)) {
+    // `out.__proto__ = value` reaches the inherited setter and repoints the
+    // object instead of adding a key, so `__proto__: { isAdmin: true }` in a
+    // file would have `config.get('isAdmin')` answer true. Dropped at every
+    // depth, because a config key by that name can never be read back anyway.
+    if (key === '__proto__') continue;
```
dunxonu: merge() strips a __proto__ key only inside plain objects, not inside array elements, contradicting its own comment that it is "Dropped at every depth".

**How it fails:** A config file sets `hosts:\n  - __proto__:\n      isAdmin: true\n`. Because arrays fail isPlainObject, merge() assigns the array wholesale at `out[key] = value` (line 18) without recursing into its elements, so the array element keeps an own, unstripped `__proto__` property. Any later shallow copy of that element (e.g. `Object.assign({}, config.get('hosts')[0])`) reassigns the copy's own prototype instead of adding a data property, exactly the class of bug this PR otherwise fixed for top-level and nested-object keys.

packages/core/src/config/files.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /** The merged contents, `{}` when every path was absent. */
+  async load(): Promise<ConfigValues> {
+    let values: ConfigValues = {};
+
+    for (const path of this.#paths) {
+      const absolute = resolve(this.#cwd, path);
+      const file = Bun.file(absolute);
+      if (!(await file.exists())) continue;
```
dunxonu, line 90: A directory listed in `files` is not skipped like a missing file, since `file.exists()` is true for directories too, and the following `file.text()` throws an unwrapped, non-ConfigError exception.

**How it fails:** `files: ['./config']` where `./config` is a directory (e.g. a stale rename, or a typo pointing at a folder instead of `config/app.yml`): `Bun.file(absolute).exists()` returns true, so the loop falls through to `parserFor` and `file.text()` (lines 76-77), which throws a raw filesystem error not caught by the try/catch that only wraps `parse(text)` (lines 83-92) - boot fails with a confusing low-level error instead of the module's own clear ConfigError wording.

dunxonu (commented):

petarzarkov:
One real, one not.

**`__proto__` inside an array element** - real, and my comment claiming "dropped at every depth" was false. The guard sat in the merge, which replaces an array whole rather than recursing into it. Reproduced before fixing: the element kept an own `__proto__`, and `Object.assign({}, element)` repointed the copy's prototype and answered `isAdmin` as true. Sanitizing now happens once on the parsed value, before anything merges, so objects, nested objects and array elements are one walk. Regression test added.

**A directory listed in `files`** - not reproducible. The finding assumes `Bun.file(dir).exists()` is true for a directory; on Bun 1.4.2 it answers **false**, so a directory is skipped exactly like a missing file and `text()` is never reached. Probed directly:

```
dir exists()     : false
dir text()       : THROWS Error | Directories cannot be read like files
```

The `text()` behaviour is as described, but it is unreachable through this path. Added a test pinning the skip so a future change to `exists()` would surface here rather than as a raw filesystem error at boot.

dunxonu (commented):
After running all 8 finder angles (line-by-line scan, removed-behavior audit, cross-file tracer, reuse/simplification/efficiency/altitude, and CLAUDE.md conventions) plus verification on the two candidates that survived initial triage, nothing held up to the bar of "a maintainer would act on this."

Summary of what was checked and ruled out:
- `packages/core/src/config/files.ts` (`ConfigFiles.load`, `sanitize`, `merge`, `parserFor`): `__proto__` stripping covers top-level, nested, and array-element cases (tested); TOML `Temporal` date handling is correct via `isPlainObject`'s prototype check; empty-file skip is uniform across YAML/TOML/JSON; parser-not-found vs parse-error messages are distinct and tested.
- `packages/core/src/config/module.ts`: the two `forRoot` overloads correctly gate `ConfigValues`-typed `validate` behind `files`; the environment-over-files step is an intentionally shallow spread (documented and tested as "wins by exact key name," not a bug); `undefined` env entries are filtered before the overlay so they don't clobber file values.
- `isPlainObject` move from `packages/http/src/client/json.ts` to `packages/core/src/plain-object.ts`: correctly re-exported, no duplicate left behind, `@dunx/http`'s existing test still passes through the re-export, and `@dunx/core` was already a peer/dev dependency of `@dunx/http` so no new runtime dependency edge.
- `tools/create-app` split (`generate.ts` → `generate-config.ts` + `header.ts`): `scaffold.ts` updated consistently, no stale imports.
- `examples/full` wiring (`config.ts`'s `configModule()`, `application.yml`, `jobs.processor.ts`, `overrides.test.ts`, `throttle.test.ts`, `dashboard.test.ts`): all call sites updated consistently; the required `seed.users` field is read via a path anchored on `import.meta.dir`, so it works correctly regardless of the forked job-processor's cwd.
- Two weaker candidates (sequential file reads in `ConfigFiles.load()`; no dedicated Tour `*.demo.ts` distinguishing the config-file provenance of `seedUsers`) were run through verification: the first was REFUTED as an unmeasured, boot-path-only claim; the second came back PLAUSIBLE but marginal - `ConfigModule`'s other options (`source`, `schema`, `as`) never got dedicated demos either, and the guide already narrates the file-merge behavior, so this doesn't clear the bar of a clear, actionable Rule 4 violation.

```json
[]
```

PR #88, opened 2026-09-11
feat(infra): a cache subpath with memory, redis and tiered stores

### LINKED ISSUE #78: Supports caching modules
Supports caching modules, providing both local and remote caching, with customizable extensions for L1 and L2 caches.

petarzarkov:
Closes #78. A `./cache` subpath on `@dunx/infra`, not a package: no new dependency in any position, and the issue's author is the user the roadmap asks for.

`CacheStore` is an abstract class with `MemoryCacheStore`, `RedisCacheStore` and `TieredCacheStore`; `Cache` is `get`/`set`/`del`/`wrap`, with per-process single flight in `wrap`. Guide: [docs/guide/19-caching.md](docs/guide/19-caching.md). The reasoning is in `internal/notes/roadmap/database-backed-infrastructure.md`.

Cut from the issue and from the earlier sketch:

- **No plugin surface for "customizable extensions".** One abstract store, three implementations, an app binds its own subclass.
- **No SQL store.**
- **No cross-process L1 invalidation.** A subscribing `Bun.RedisClient` holds the event loop open after `close()`. Two nodes with L1 can serve stale entries until TTL; the guide says so.
- **No `@Cacheable`.** The boot scan reaches only `providers` and controllers, so a marked class outside the graph is a silent no-op; `schedule` binds a method where caching must replace it; argument-to-key serialization is the real policy surface.
- **`ThrottleStore` not reused, sweep not moved.** Different contract, and the only shared home would be `@dunx/core`.

`docs/guide/19-caching.md` moves the five pages after it up one number, so the Infrastructure nav section stays contiguous. Slugs are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added configurable caching with in-memory, Redis, and two-tier storage options.
  * Added expiration, key prefixes, eviction, read-through loading, and concurrent-load deduplication.
  * Added cache support to generated application templates, including a catalog demonstration.

* **Documentation**
  * Added guides covering caching, health checks, agent tooling, metrics, and framework upgrades.
  * Updated navigation, package references, and cross-links to reflect the new documentation.


packages/infra/src/cache/redis.ts
```diff
@@ -0,0 +1,52 @@
...
+    super();
+  }
+
+  async get<V = unknown>(key: string): Promise<V | undefined> {
+    const raw = await this.redis.get(key);
+    return raw === null ? undefined : (JSON.parse(raw) as V);
+  }
+
+  async set<V>(key: string, value: V, ttl: number): Promise<void> {
+    const raw = JSON.stringify(value);
```
dunxonu, line 39: RedisCacheStore.set compares JSON.stringify's result to undefined, but JSON.stringify is typed to always return string, so the check is a TypeScript no-overlap error (TS2367) under this repo's strict config.

**How it fails:** `const raw = JSON.stringify(value);` types `raw` as `string` (JSON.stringify's declared return type never includes `undefined`, even though it returns `undefined` at runtime for a top-level `undefined`). The following `if (raw === undefined)` compares a `string` to `undefined`, which TypeScript strict mode flags as 'this condition will always return false since the types have no overlap' - `bun run typecheck` / `tsc --noEmit` fails on this file, and `packages/infra/src/cache/redis.test.ts`'s own 'removes the key rather than writing an unserialisable value' test (which calls `store.set('k', undefined, 5_000)`) cannot run because the package does not build.

packages/infra/src/cache/options.ts
```diff
@@ -0,0 +1,39 @@
...
+import type { CacheStore } from './store.js';
+
+export interface CacheOptionsInit {
+  /** Lifetime in milliseconds for a write that names none. @default 60000 */
+  readonly ttl?: number;
+  /** Prepended to every key with a `:`, so two apps can share one Redis. */
+  readonly prefix?: string;
+  /** @default new MemoryCacheStore() */
+  readonly store?: CacheStore;
+}
```
dunxonu, line 12: CacheOptions.prefix is optional with no validation, unlike the sibling ThrottleOptions which hard-fails boot on a missing prefix for exactly this reason.

**How it fails:** `packages/http/src/throttle/options.ts:58-63` throws at boot if `prefix` is empty, specifically because two apps sharing one Redis with no prefix 'spend the other's budget.' `CacheOptions` has no such guard, and `keyFor()` returns the raw key when `prefix` is `undefined` (the default). Two independent apps pointed at the same Redis instance, each calling `CacheModule.forRootAsync({ store: new RedisCacheStore(redis) })` with no `prefix`, will silently read and overwrite each other's cached values under identical keys (e.g. both caching `quote:aapl`).

packages/infra/src/cache/tiered.ts
```diff
@@ -0,0 +1,54 @@
...
+    init: TieredCacheInit = {},
+  ) {
+    super();
+    this.#promoteTtl = Math.max(1, init.promoteTtl ?? 30_000);
+  }
+
+  async get<V = unknown>(key: string): Promise<V | undefined> {
+    const near = await this.l1.get<V>(key);
+    if (near !== undefined) return near;
+    const far = await this.l2.get<V>(key);
```
dunxonu, line 50: TieredCacheStore.set writes to L2 then L1 sequentially, while del() in the same class parallelizes both tiers with Promise.all, adding an avoidable Redis-round-trip latency to every cache write.

**How it fails:** Every `Cache.set`/`Cache.wrap` write through a `TieredCacheStore` pays the L2 (Redis) round trip and then the L1 (in-memory) write serially; under load this doubles the tail latency of the write path compared to running them concurrently, which is already the pattern used two methods below in `del()`.

dunxonu (commented):

packages/infra/src/cache/tiered.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (far === undefined) return undefined;
+    await this.l1.set(key, far, this.#promoteTtl);
+    return far;
+  }
```
dunxonu, line 54: TieredCacheStore.set writes to L2 then L1 sequentially with two awaits, instead of in parallel via Promise.all like del() in the same class.

**How it fails:** Every Cache.set/wrap-miss through a TieredCacheStore pays the L1 and L2 write latencies back-to-back rather than concurrently; verified there is no correctness dependency on the ordering (both are unconditional, independent writes), so this is pure wasted latency on a hot write path, not a safety net.

dunxonu (commented):

petarzarkov:
Worked through all 10 inline comments. Four were real and are fixed in 846f80a, each reproduced as a failing test before the change; six were not, with reasons below.

## Fixed

**`MemoryCacheStore` `max`, and `TieredCacheStore` `promoteTtl`** (coderabbit, `memory.ts:32` / `tiered.ts:33`). Both real, same mechanism: `Math.max(1, NaN)` is `NaN`. Measured before the fix, `new MemoryCacheStore({ max: NaN })` held 50,000 entries against a 10,000 bound, because `size > NaN` is never true; `promoteTtl: NaN` put `expiresAt: NaN` into L1, which nothing compares greater than, so the promoted entry kept answering after the L2 entry it came from was deleted. `Infinity` does the same on both. Both constructors now throw `AppError` on a non-finite bound, matching what `CacheOptions` already does for `ttl`. I kept the `max: 0 -> 1` clamp, since `memory.test.ts` documents it; `promoteTtl <= 0` throws, matching `ttl`.

**`Cache.wrap` overwriting a later mutation** (coderabbit, `cache.ts:18`). Real, and the `del` half is the one that bites: with a load in flight, `cache.del(key)` returned `false` and the loader then wrote the value back, so an explicit invalidation read as a hit for the full ttl. A `set` during a load was likewise overwritten by the older value. Fixed without a per-key generation map: a key mutated while it has a load in flight is recorded, and `#read` skips the store write. The record is keyed on the in-flight set, so it holds nothing between `wrap` calls and cannot grow. Three tests added, plus one that the key stores normally again afterwards.

**The L1 expiry assertion** (coderabbit, `tiered.test.ts:46`). Correct: the assertion went through `TieredCacheStore.get`, which L2 alone could satisfy. It now asserts the L1 miss first.

## Not changed

**`JSON.stringify` compared to `undefined` is a TS2367 error** (`redis.ts:39`). Does not reproduce. `bunx tsc --noEmit` in `packages/infra` exits 0 on this branch once `dist/` is built, and CI's "Lint, format, types" job is green. I probed it directly: a file containing both that comparison and a plain `const s: string = 'a'; s === undefined` control produces no diagnostic under TypeScript 7.0.2, while a deliberate `TS2322` in the same file does, so the file is in the program. `redis.test.ts`'s "removes the key rather than writing an unserialisable value" runs and passes today. The runtime guard is still needed, since `JSON.stringify(undefined)` really is `undefined`.

**`CacheOptions.prefix` should be required like `ThrottleOptions`** (`options.ts:12`). Skipped as a design change rather than a defect, but flagging it rather than dismissing it. `ThrottleModule.forRoot` already requires an options object, so a required `prefix` costs nothing there; `CacheModule.forRoot()` takes none, defaults to the in-process `MemoryCacheStore` where a prefix is inert, and the zero-argument form is what the guide and `cache.test.ts` document. Happy to make it required if you want the throttle precedent applied uniformly.

**`TieredCacheStore.set` should `Promise.all` like `del`** (`tiered.ts:39`). Measured, and the premise does not hold. L1 is an in-process `Map` write whose promise is already resolved, so there is no second round trip to overlap: against a 2 ms L2, 500 iterations gave 2.0623 ms/set sequential against 2.0693 ms/set with `Promise.all`. The order is also load bearing: after a failed L2 write, the current code leaves L1 empty, while `Promise.all` leaves L1 serving a value the caller saw throw, for up to `promoteTtl`.

**`process.pid` in the example and template cache prefix** (`examples/full/src/cache/cache.module.ts:83`, `tools/create-app/templates/features/cache/cache.module.ts:83`). Skipped: the pid is the example suite's cross-run isolation, and it is an existing convention here, documented at `examples/full/src/throttle/throttle.module.ts:20` ("The prefix carries the pid, so two runs against one Redis cannot spend each other's budget") and already used by `cache/sessions.service.ts:7` and `throttle/throttle.module.ts:44`. I applied the suggested `dunx-full:cache` and ran the example suite: `service.test.ts:321` fails (`expect(await loads()).toBe(before + 1)`, got 0) and `tour.test.ts:271` fails (narration reads "10 concurrent reads of an uncached key -> 0 load"), because a warm shared L2 makes the first read a hit inside the 30s ttl. Redis here currently holds 320 live keys across past runs' pid namespaces, which is exactly what stops them seeing each other. The two files are also compared byte for byte by `features.test.ts`, so they cannot diverge.

**`response.ok` checks in `catalog.demo.ts`** (`catalog.demo.ts:38`). Skipped for consistency: none of the 19 other `*.demo.ts` files in `examples/full` check `ok`; they log `response.status` or parse directly. The demos narrate, and the assertions live in `service.test.ts` and `tour.test.ts`, where a failed fetch surfaces as a wrong narration line. Adding it to this one file alone would make it the odd one out.

## One thing worth knowing

Chasing an unrelated local `docs/test` failure ("every in-site link points at a page that exists" reporting `/guide/caching` as missing) cost a while: Bun's on-disk runtime transpiler cache serves a stale module for the `?raw` import of `internal/docs/src/generated/index.json`, so `data.ts`'s `site` had 26 guides while the regenerated file on disk had 27. Deleting and regenerating `generated/` does not clear it; `BUN_RUNTIME_TRANSPILER_CACHE_PATH=0` does, and the suite then passes 145/145. CI never hits it because the cache starts empty there. Might be worth a line in `docs/bun-apis.md`, which I left alone as out of scope here.

`bun run ci` is green: 13 steps, 13 passed. `infra` coverage 92.7% lines / 93.2% functions. `gen:mcp` and `gen:readme` are in the commit.

dunxonu (commented):
No naming collisions elsewhere. This confirms the new `@dunx/infra/cache` subpath integrates cleanly with no cross-file breakage. Based on the thorough review across all angles, this PR is well-implemented and comprehensively tested (every edge case I identified as a candidate bug was already covered by an explicit test), with documented tradeoffs for the remaining design limitations.

```json
[]
```

dunxonu (commented):
Confirmed. Now let me do a final quick self-check pass on other angles (cleanup/efficiency/altitude/conventions) already covered above; nothing else rose to a confirmable, concrete finding. I'll finalize the output.

```json
[
  {
    "file": "packages/infra/src/cache/tiered.ts",
    "line": 1108,
    "summary": "TieredCacheStore.set() and del() are not atomic with each other, so a concurrent set/del race on the same key can leave L1 holding a stale value after del() reports success.",
    "failure_scenario": "Key 'k' exists in both tiers. Caller runs `cache.set('k','v1')` and, without awaiting it, `cache.del('k')`. set() completes `await l2.set(k,v1)` and suspends before its pending `l1.set(k,v1)` runs; del()'s `Promise.all([l1.del(k), l2.del(k)])` executes synchronously in that gap, deleting the old L1 entry and the just-written L2 entry. set()'s continuation then runs `l1.set(k,v1)`, writing v1 into L1 after both tiers were cleared. Final state: L2 has no entry, L1 has 'v1', del() returned true, and every subsequent get('k') serves the stale v1 from L1 until promoteTtl expiry - contradicting the class's implicit contract that a reported delete actually removes the entry."
  }
]
```

dunxonu (commented):
**Actionable comment posted: 0**

6 further findings are in code untouched since the last review, and so were reportable then. Not repeated here.

- `examples/full/src/tour.test.ts:44` CatalogDemo's narration lines are never asserted in tour.test.ts, unlike every other demo.

</details>

Issue #82, opened 2026-09-10
Feature Request: OpenTelemetry / Observability integration


yibird:
# Feature Request: OpenTelemetry / Observability integration

## Motivation

Dunx already provides useful observability-related capabilities such as logging, metrics, request context, runtime statistics, and request tracing context.

However, I could not find an official, unified OpenTelemetry integration for Dunx.

For production applications and distributed systems, it would be valuable for Dunx to provide first-class OpenTelemetry support while keeping the existing lightweight and high-performance architecture.

## Proposal

Consider adding an official OpenTelemetry integration, for example:

* HTTP request tracing
* Distributed trace context propagation
* Metrics integration
* Database query tracing/metrics
* Redis and queue instrumentation
* Trace/span information in `RequestContext`
* Correlation between logs, metrics, and traces
* Export through standard OpenTelemetry exporters

Ideally, this could be provided as an optional package such as:

```text
@dunx/otel
```

or another appropriate package name, without increasing the overhead of applications that do not need observability.

The integration could build on Dunx's existing `Logger`, `Metrics`, and `RequestContext` infrastructure rather than introducing a completely separate observability system.

## Why

OpenTelemetry is becoming the standard observability layer for modern backend applications.

A first-class integration would make Dunx easier to use with existing observability platforms such as Jaeger, Grafana, Prometheus, and other OpenTelemetry-compatible backends.

It would also complement Dunx's existing performance and production-oriented features such as HTTP, database, Redis, queues, and resilience.

I believe this could make Dunx a stronger choice for production and distributed applications while preserving its lightweight architecture.

petarzarkov:
We already have this partly implemeted following OpenTelemtry. Will have to research into this and consider it.

petarzarkov:
Researched this properly. Short version: more of it is already there than I implied, and the single biggest item on your list would not work on Bun at all.

What already ships, with no OpenTelemetry dependency:

- Full W3C Trace Context. Inbound `traceparent` is parsed and adopted, ids are minted when absent, `tracestate` is passed through, and `traceresponse` is written on the way out.
- `traceId`, `spanId`, `parentSpanId` and `traceState` land in the request scope, so every log line from any depth of the call stack correlates without being passed anything. The field names are the OpenTelemetry log data model's own.
- Outbound propagation: `@dunx/http/client` injects `traceparent` on egress, so a trace crosses services.
[...]

PR #89, opened 2026-09-11
feat(infra): queue and Redis metrics

petarzarkov:
Part of #82, which asks for more than this: the queue and Redis halves of the
metrics collection, not an OpenTelemetry integration.

`RedisMetrics` records count, errors and a `Durations` per command verb at
`Redis.#run`. `QueueMetrics` records enqueues at `JobPublisher.publish` and
handler durations at `JobDispatcher.dispatch`, keyed by queue and job name.
Both bind only when a `{ metrics: true }` settings object is passed.

**Scope limit, documented in the class, the guide and the example:** `isolation`
defaults to `'process'`, so a `@JobHandler({ background: true })` runs in a forked
child with its own container. Its duration never reaches a `QueueMetrics` in the
web process, and neither does a `WorkerFactory` worker's. `handled` stays 0 while
`published` climbs. Only the publish side and foreground handlers are observable.

Threading: the settings object is the last argument everywhere, so it sits after
`as` on `RedisModule` and second on `QueueModule`. It cannot live on the init,
because `forRootAsync` resolves that through a factory and the binding is decided
when providers are declared.

No dashboard panel. Nothing new in CI: valkey is already declared for the
`coverage` and `examples` jobs.


packages/infra/src/queue/metrics.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  observePublish(
+    queue: string,
+    name: string,
+    durationNs: number,
+    failed = false,
+  ): void {
+    const stats = this.#seriesFor(queue, name);
+    this.#published += 1;
+    stats.published += 1;
```
dunxonu, line 94: The `MAX_SERIES = 128` cap and its doc comment ("Distinct (queue, name) pairs past which everything else collapses") undercount the real ceiling: the overflow bucket itself occupies a 129th map slot once created.

**How it fails:** A caller with 128 distinct (queue, name) pairs plus one more triggers overflow; `snapshot().jobs` then has length 129, not 128 as the doc/commit message imply. The project's own test asserts `toHaveLength(129)`, confirming the behavior is intentional but the prose is imprecise enough to mislead an operator sizing a dashboard payload around "128 series max".

packages/infra/src/redis/module.ts
```diff
@@ -55,6 +57,41 @@ export const redisConnection = (name: string): Token<RedisConnection> => {
...
+  return created;
+};
+
+/** What `Redis` and any subclass of it is constructed with. */
+type RedisCtor = new (
+  options: RedisOptions,
+  metrics?: RedisMetrics,
+) => RedisConnection;
+
 const connectionFrom = (
```
dunxonu, line 95: `RedisMetrics`/`QueueMetrics` reimplement the same keyed-series-plus-Durations-plus-cap shape already present independently in `packages/infra/src/db/metrics.ts` (`QueryMetrics`) and `packages/http/src/server/metrics.ts` (`RequestMetrics`), with no shared helper in `@dunx/core/stats`, and the `settings.metrics === true` conditional-binding idiom is hand-copied across `db/module.ts`, `queue/module.ts`, and `redis/module.ts` with three slightly different shapes rather than one extracted mechanism.

**How it fails:** A future fix to the overflow/cap logic, or to the metrics-binding wiring pattern, has to be applied in three or four places by hand; this PR explicitly follows the pre-existing `DbModule` convention rather than introducing new duplication, so this is a design note rather than a regression this PR caused.

dunxonu (commented):

- `(commit message):0` The commit ends with a `Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>` trailer, which CLAUDE.md explicitly forbids.

</details>

packages/infra/src/redis/metrics.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+const series = (): Series => ({
+  count: 0,
+  errors: 0,
+  duration: new Durations(),
+});
+
+/**
+ * How long Redis is taking, by command.
```
dunxonu, line 36: RedisMetrics has no cap on distinct series, unlike QueueMetrics, even though the command key can be caller-controlled through Redis.send().

**How it fails:** An app calls `redis.send(dynamicString, [])` (e.g. built from request or job data, even by mistake) with many distinct values; `Redis.#run` in packages/infra/src/redis/client.ts uppercases and records every one via `metrics?.observe(command, ...)` regardless of whether the server recognized it, so `RedisMetrics#series` grows one HDR histogram per distinct string for the life of the process. The class doc and docs/guide/22-metrics.md both claim 'the series count is bounded by Redis's own command vocabulary,' but nothing enforces that: `send()` takes an arbitrary caller string with no allow-list, the same caller-controlled-key problem `QueueMetrics` (packages/infra/src/queue/metrics.ts) explicitly caps at `MAX_SERIES = 128` with an `(other)` overflow bucket.

dunxonu (commented):

petarzarkov:
Addressed in d66e12f.

**Fixed - `RedisMetrics` had no cap.** Real, and reproduced before the fix: 500
distinct verbs through `send()` against a refused connection gave 500 series and a
70 KB snapshot. The mechanism is the one named: `send()` uppercases whatever string
the caller passed, and `Redis.#run` records a rejected command against its verb like
any other, so the "bounded by Redis's own command vocabulary" claim in the class doc
and in `docs/guide/22-metrics.md` was wrong in both places. Same run now measures 129
series and 18 KB.

**Fixed - the ceiling was stated as 128 where the code means 129.** Corrected in the
guide, and the shared class now says it: the overflow series takes a slot of its own,
so a snapshot after the collapse carries one more than the cap.

**Partly fixed - the duplication.** The cap was the half worth extracting, and adding
a second hand-rolled copy to `RedisMetrics` would have been the third. `QueueMetrics`'
cap moved out to `CappedSeries` in `packages/infra/src/series.ts`, the lowest owner
`/queue` and `/redis` both reach, and both now key through it.

Skipped the rest of that comment, with reasons:

- **Not `@dunx/core/stats`.** `QueryMetrics` keys on a five-value frozen union and
  `RequestMetrics` on the frozen `RouteContext` plus a bounded method set; neither has
  a caller-controlled key space, so neither can use a cap. Hoisting to core would move
  the code away from its only two consumers.
- **The `settings.metrics === true` binding idiom.** What the three modules share is a
  ternary. `DbModule` binds a class and wraps the driver, `QueueModule` constructs one
  instance and threads it through three providers as a `useValue`, `RedisModule` binds
  either the class or a memoised per-name `Token` and threads it into the connection
  factory. A helper over that would be thinner than its call sites and could not
  unify the token-versus-class or eager-instance differences.

**Skipped - the commit-message trailer.** `9779424` carries no `Co-Authored-By`
trailer: `git log -1 --format='%(trailers)' 9779424` is empty, and so is the message
the PR commits API returns. Nothing to remove.

Regression tests in `packages/infra/src/redis/metrics.test.ts`: the unit cap and its
overflow counts, that an existing verb keeps counting after the collapse, and the
end-to-end path through `send()` on the seam. All three fail on the parent commit.

Gates run locally: `build`, `static`, `examples`, `docs`, `browser`, `coverage`
(`infra` 92.8% lines / 92.8% functions).

dunxonu (commented):
Confirmed. Based on my review of the diff and verification, here are the findings.

```json
[
  {
    "file": "packages/infra/src/queue/worker.ts",
    "line": 472,
    "summary": "WorkerFactory.attach() and WorkerFactory.create() build a JobDispatcher with no metrics argument, so handlers consumed through them never reach QueueMetrics even when it is bound on the container.",
    "failure_scenario": "An app does `const app = await HttpFactory.create(AppModule)` with `QueueModule.forRoot(init, { metrics: true })` bound in that same container, then `const consumer = await WorkerFactory.attach(app, AppModule)` per worker.ts's own documented pattern for running HTTP and queue consumption in one process. `new JobDispatcher(jobs, app.get(QueueOptions).jobTimeoutMs)` is constructed with no `metrics` argument, so every handler dispatched through this consumer silently records nothing in `app.get(QueueMetrics)` - `handled` stays 0 forever, contradicting docs/guide/22-metrics.md's claim that a foreground (non-`background`) handler 'does land in its handlerDuration,' and with no error or doc caveat warning that `attach()` is exempt."
  }
]
```

Note: this gap does not affect the module's own `consume: true` path (wired correctly in `packages/infra/src/queue/runner.ts`), and `WorkerFactory.create()`'s separate-container case is at least consistent with the documented "forked/dedicated worker is invisible" framing - `attach()` is the one place where the promise of "foreground handlers are observable" is broken silently. No other correctness, cleanup, altitude, or convention issues survived verification across the `CappedSeries` cap arithmetic, the `RedisModule`/`QueueModule` settings threading, the dispatcher/publisher timing seams, or the example wiring.

petarzarkov:
The `WorkerFactory` finding is real. Fixed in 8fb5555.

Reproduced first: a live-broker test that attaches to a container with
`QueueModule.forRoot(init, { metrics: true })`, publishes, and waits for
`handled === 1` times out on the parent commit. `attach()` consumes inside the
caller's own container, so that handler runs in this process and the guide promises
it lands in `handlerDuration`; the dispatcher was built with no metrics argument, so
it never did. `create()` had the same gap against the worker process's own
container. Both now take `metricsIn(modules, app)`.

`metricsIn` asks the module graph whether `QueueMetrics` is bound rather than
resolving it: an unbound class self-binds, so `app.get(QueueMetrics)` with metrics
off would hand back a live instance every handler then paid a clock to fill. It sits
in `dispatcher.ts` next to the `declares` predicate `assertQueueModule` now shares
rather than scanning providers a second time.

Two things I did not change, with reasons:

- **The forked sandbox child stays as it is.** `isolation: 'process'` is the
  documented invisible case, its container is built and torn down by dunx around one
  forked process, and wiring it would buy a snapshot with no reader.
- **The docs said more than the code kept.** The guide and the class doc both wrote
  "its own `QueueMetrics`" of the forked child and of a worker process together,
  which reads as a promise the child does not keep. They now name the three
  containers that do record - `consume: true`, `attach()`, and the process
  `create()` booted - and say the child's durations reach nothing here.

Regression tests: `WorkerFactory.attach` records into the container's `QueueMetrics`,
and a worker process records into its own. Both are `it.if(live)`, so they run in the
`coverage` and `examples` jobs where valkey is declared.

Gates re-run locally after the change: `build`, `static`, `examples`, `docs`,
`coverage` (`infra` 92.8% lines / 92.9% functions).

packages/infra/src/queue/metrics.ts
```diff
@@ -0,0 +1,142 @@
...
+  readonly handled: number;
+  readonly since: string;
+}
+
+interface Series {
+  readonly queue: string;
+  readonly name: string;
+  published: number;
+  publishErrors: number;
+  publishDuration: Durations | undefined;
```
dunxonu, line 41: QueueMetrics and RedisMetrics duplicate the same total/errors/since/reset bookkeeping pattern already present in the pre-existing QueryMetrics (packages/infra/src/db/metrics.ts), with no shared base extracted.

**How it fails:** A future fourth metrics class (or a bug fix to reset()/since handling) has to be copied a fourth time by hand; CLAUDE.md Rule 2 requires moving shared logic to one declaration rather than adding a second (here third) copy - a fix to one accumulator's reset() semantics can silently miss the other two.

dunxonu (commented):

3 further findings are in code untouched since the last review, and so were reportable then. Not repeated here.

petarzarkov:
Skipping the shared-accumulator-base comment, and this is the reasoning rather than a
refusal.

The premise is that the three hold a `total`/`errors`/`since`/`reset` shape. They do
not. Field by field:

| Class            | Series container                | Top-level counters             |
| ---------------- | ------------------------------- | ------------------------------ |
| `QueryMetrics`   | `Map<QueryOperation, Series>`   | `#total`                       |
| `RedisMetrics`   | `CappedSeries<Series>`          | `#total`, `#errors`            |
| `QueueMetrics`   | `CappedSeries<Series>`          | `#published`, `#handled`       |
| `RequestMetrics` | `Map<RouteContext, Series>` + `#unmatched` | none            |

All four also hold `#since`, and that is the whole intersection. It shows in the
payloads too: `DbStatsReport` is `{operations, total, since}`, `RedisStatsReport` is
`{commands, total, errors, since}`, `QueueStatsReport` is `{jobs, published, handled,
since}`, `HttpStatsReport` is `{routes, inFlight, pendingWebSockets, since}`. One
field in common across four report shapes.

Methods do not coincide either: `observe(sql, ns, failed)`, `observe(command, ns,
failed)`, `observePublish(...)` plus `observeHandled(...)`, and `observe(ctx, status,
ns, traceId)`. `snapshot()` shares a name and four return types. `reset()` shares a
signature, and each body clears a different set of counters plus its own container.

So a base could hold exactly `#since`: one field, `toISOString()` in `snapshot()`, and
`= new Date()` in `reset()`. Against that:

- Private `#` fields are invisible to a subclass, so it needs a `protected` field or
  getter, and four `super.reset()` calls, to save four lines each.
- Reaching `RequestMetrics` puts it in `@dunx/core`, so `@dunx/http` inherits from
  core for a `Date`.
- It adds a constructor per class to the coverage denominator, and `@dunx/infra` is at
  92.9% functions.

The Rule 2 test is whether these can drift in a way that matters, and the answer is
that the base cannot hold the parts that drift. `since` has no branch, no option and
no format choice: all four already agree and a "fix to `since` semantics" is the one
line. What could genuinely disagree is the counters, and those are precisely the
fields that do not coincide. Absorbing `total` would exclude `QueueMetrics` and
`RequestMetrics`, which have none; absorbing `errors` would exclude the other three,
since `QueryMetrics` counts errors per series and never totals them.

The half of this note that was real was the cap, and it is fixed: `CappedSeries` in
d66e12f is one declaration for both keyed metrics. `QueryMetrics` and `RequestMetrics`
stay out of it because a query operation is a five-value union and a route series is
keyed on the frozen `RouteContext`, so neither has a key space to bound - not because
a second copy was preferred.

dunxonu (commented):
Confirmed all details for the CappedSeries findings. Now compiling the final ranked list.

```json
[
  {
    "file": "packages/infra/src/queue/processor.ts",
    "line": 119,
    "summary": "JobProcessor#boot builds JobDispatcher without wiring QueueMetrics, unlike the other two dispatcher construction sites this PR updated (worker.ts, runner.ts).",
    "failure_scenario": "An app enables QueueModule.forRoot(init, { metrics: true }) and uses background/forked job handlers (JobProcessor is the sandboxed-child path for @JobHandler({ background: true })). Every handler invocation there never calls metricsIn(modules, app), so JobDispatcher is constructed with only (jobs, jobTimeoutMs) - QueueMetrics.observeHandled is never called for any background job, silently producing an incomplete metrics report with no error or warning."
  },
  {
    "file": "examples/full/src/jobs/jobs.demo.ts",
    "line": 66,
    "summary": "The new audit-job wait is not wrapped in try/catch, unlike the symmetric thumbnail-job wait a few lines above.",
    "failure_scenario": "If the in-process 'record' handler for AUDIT_QUEUE fails or waitUntilFinished exceeds SETTLE_MS (8s) under CI load, the rejection propagates out of publishAndWait() into demonstrate()'s catch, which only swallows isConnectionError - any other error (e.g. a bullmq TimeoutError) is rethrown and crashes the tour/example instead of degrading gracefully like the thumbnail path does."
  },
  {
    "file": "packages/infra/src/redis/module.ts",
    "line": 118,
    "summary": "connectionFrom casts an arbitrary Ctor<RedisConnection> to a two-argument RedisCtor and always calls it with (options, metrics), assuming every subclass forwards the metrics argument to super().",
    "failure_scenario": "A consumer subclass like `class SessionsRedis extends Redis { constructor(options: RedisOptions) { super(options); } }` (an explicit constructor that omits the second parameter) compiles fine against Ctor<RedisConnection> but silently drops the metrics instance - metrics: true looks wired (token exported, bound) yet #run's metrics recording is never populated, with no compiler or runtime signal."
  },
  {
    "file": "packages/infra/src/redis/module.ts",
    "line": 182,
    "summary": "The default (unnamed) connection's provider block is duplicated near-verbatim between forRoot and forRootAsync, while the named-connection case right next to it was already factored into a shared namedModule helper.",
    "failure_scenario": "A future settings field (e.g. a configurable CappedSeries cap) added to the default-connection path requires editing two ~20-line blocks identically; a partial edit makes RedisModule.forRoot() and forRootAsync() diverge silently, the same class of bug namedModule was extracted to avoid."
  },
  {
    "file": "packages/infra/src/series.ts",
    "line": 47,
    "summary": "CappedSeries.for() always builds a new key string via keyOf() before the map lookup, even on a cache hit, on the hottest paths this PR wires metrics into (every Redis command, every job publish/handle).",
    "failure_scenario": "An app doing thousands of Redis ops/sec or queue jobs/sec with metrics on now allocates a throwaway template-literal string per call purely to do a Map.get for a label pair that repeats constantly (e.g. always 'GET'/'SET' or a fixed job name); a nested Map<string, Map<string, T>> would resolve the common case with two Map.get calls and no string allocation."
  },
  {
    "file": "packages/infra/src/series.ts",
    "line": 26,
    "summary": "CappedSeries's 128-series cap is shared per instance across every caller of .for(), so unrelated label sets on the same QueueMetrics/RedisMetrics instance compete for and can evict each other's budget.",
    "failure_scenario": "An app has several features publishing well-known, low-cardinality job names, plus one feature that also publishes with a name derived from customer data. Once the 128th distinct (queue,name) pair across the whole app is hit, a legitimate low-cardinality job name introduced afterward collapses into the same '(other)' overflow bucket right alongside the runaway data, with nothing in the JSON payload distinguishing which label set actually caused the cap - a diagnosability gap the docs (docs/guide/22-metrics.md) don't mention."
  },
  {
    "file": "packages/infra/src/queue/metrics.ts",
    "line": 82,
    "summary": "#published/#handled running totals are redundant state that duplicates what's derivable by summing across every Series (including the overflow bucket, which already accumulates real counts).",
    "failure_scenario": "A future edit that adds another code path incrementing a job's published/handled count (e.g. a bulk-publish helper) but forgets to also bump the parallel #published/#handled counters produces a snapshot().published that silently disagrees with the sum of jobs[].published - nothing catches the drift since both numbers independently look plausible."
  },
  {
    "file": "packages/infra/src/queue/module.ts",
    "line": 122,
    "summary": "`settings.metrics === true ? new QueueMetrics() : undefined` is duplicated verbatim in forRoot (line 122) and forRootAsync (line 160), along with the matching exportsFor/bindings/runner call sequence.",
    "failure_scenario": "A later change to how metrics is constructed (e.g. passing a cap or a clock override) has to be made in two places; missing one leaves forRootAsync's metrics behaving differently from forRoot with no compiler signal to catch the omission."
  }
]
```

PR #90, opened 2026-09-11
feat(core): ResiliencePolicy - timeout, retry, backoff, jitter, fallback

### LINKED ISSUE #81: Feature Request: Add Resilience support
## Feature Request: Add Resilience support

Dunx already provides many important building blocks for production applications, but I think it is missing a unified **Resilience** capability.

For real-world applications, calls to external services and infrastructure such as HTTP APIs, databases, Redis, and message queues can fail or become slow. It would be useful for Dunx to provide a consistent way to handle these failures.

### Suggested capabilities

A Resilience module could provide:

* Timeout
* Retry
* Exponential backoff
* Jitter
* Circuit breaker
* Bulkhead / concurrency isolation
* Rate limiter
* Fallback

Ideally, these capabilities could be composed into a resilience pipeline:

```text
Service
  ↓
Resilience
  ├── Timeout
  ├── Retry + Backoff + Jitter
  ├── Circuit Breaker
  ├── Bulkhead
  └── Fallback
  ↓
External Dependency
```

For example:

```ts
const payment = resilience('payment', {
  timeout: 3000,
  retry: {
    maxAttempts: 3,
    backoff: 'exponential',
    jitter: true,
  },
  circuitBreaker: {
    failureRate: 50,
    openDuration: 30_000,
  },
})
```

Then application code could simply use:

```ts
await payment.execute(() => paymentApi.charge(order))
```

It would also be great if Resilience could integrate with Dunx's DI and interceptor system, allowing policies to be applied consistently without duplicating retry/timeout/circuit-breaker logic throughout services.

I think a built-in Resilience module would make Dunx more suitable for production and distributed applications.

petarzarkov:
Closes #81, in part.

The generic half of `@dunx/http/client`'s retry moves to `@dunx/core` as
`ResiliencePolicy` (timeout, retry, backoff, jitter, fallback), with
`ResilienceModule.forRoot`/`forRootAsync` and a memoised `resiliencePolicy(name)`
shaped like `HttpModule`. `RetryClassifier` is the seam that keeps `HttpStatusCode`
out of core; `HttpRetryClassifier` stays in `@dunx/http`, and `client/retry.ts`
goes from 167 lines to 83. `HttpService` builds a policy per call.

Consumer surface holds: `BackoffOptions` and `RetryOptions` are re-exported from
`@dunx/http/client`, and `HttpRetryOptions` adds `shouldRetryOnStatus` and
`respectRetryAfter`. `examples/full/src/upstream` compiles unchanged and gains an
`UpstreamPolicy`.

**Guide number: this branch adds `docs/guide/25-resilience.md` and renumbers
nothing.** It is picked against #88's layout, where `19-caching` pushes the tail to
`24-upgrading`, so 25 is free either way. On this branch alone there is a gap at 24;
order still ascends and "Going live" stays contiguous, so `site.test.tsx` passes
before and after that merge. No other guide file is touched.

Refused, and why:

- **Rate limiter.** A second declaration of `ThrottleStore`. `ThrottleModule` is
  the same job in the direction that has the caller to key on.
- **Retry of work this process owns.** bullmq's, through `@dunx/infra/queue` -
  `internal/notes/research/brokers.md:394`.
- **Circuit breaker and bulkhead.** Deferred. Half-open probing, window choice,
  isolation and state hydration are a library's job; cockatiel as an optional peer
  if it is taken.

`docs/bun-apis.md` gains one finding this turned up: Bun's runtime transpiler
cache serves a stale `?raw` module over 50 KB, so regenerating the docs model and
running its suite in the same checkout reads the previous model.

## Summary by CodeRabbit

- **New Features**
  - Added configurable resilience policies with retries, timeouts, exponential backoff, jitter, cancellation, callbacks, and fallback results.
  - Added HTTP-aware retry handling for status codes, transport failures, and `Retry-After` responses.
  - Added response-body parsing that supports JSON and raw-text responses.
  - HTTP requests now honor policy-provided cancellation signals.

- **Documentation**
  - Added a comprehensive resilience guide and updated package documentation.

- **Tests**
  - Added coverage for retries, fallbacks, timeout cancellation, HTTP classification, and end-to-end resilience scenarios.


dunxonu (commented):

examples/full/src/upstream/upstream.demo.ts
```diff
@@ -80,5 +82,34 @@ export class UpstreamDemo {
...
     }
+
+    await this.demonstratePolicy(url);
+  }
+
+  /** `ResiliencePolicy` around the same upstream: the retry loop is core's, and
+   * the verdict on a status is `HttpRetryClassifier`'s. */
+  private async demonstratePolicy(url: string): Promise<void> {
+    const key = `policy-${Date.now()}`;
+    const recovered = await this.policy.run(() =>
```
dunxonu: `demonstratePolicy` calls `this.policy.run(() => this.http.get(...))`, discarding the `AbortSignal` that `ResiliencePolicy.run` hands to `op`, so the policy's `timeoutMs` never reaches the actual `fetch` call.

**How it fails:** `ResilienceModule.forRootAsync` binds `UpstreamPolicy` with `timeoutMs: config.get('upstream').timeoutMs`; because the zero-arg arrow `() => this.http.get(url, { retry: { maxRetries: 0 } })` ignores the `signal` parameter, `AbortSignal.timeout(...)` created in `ResiliencePolicy#signal()` fires but is never attached to the underlying request. If the upstream hangs instead of returning promptly, the call runs unbounded despite the policy claiming a timeout. This directly contradicts the pattern the PR's own guide documents as correct usage: `docs/guide/25-resilience.md:36` shows `this.policy.run((signal) => this.gateway.charge(order, signal))`. The identical bug is duplicated at `upstream.demo.ts:104` (the 404/fallback demo) and in the vendored copy `tools/create-app/templates/features/upstream/upstream.demo.ts`, and the new test in `examples/full/src/service.test.ts` reproduces the same pattern, so no test exercises the signal contract `ResiliencePolicy.run` actually promises.

packages/core/src/resilience/policy.ts
```diff
@@ -0,0 +1,92 @@
...
+    const verdict = this.options.classifier.classify(error);
+    if (!verdict.retry) return { retry: false, delayMs: 0 };
+
+    const computed = backoffDelay(attempt, {
+      baseMs: retryDelayMs,
+      ...backoff,
+    });
+    // An upstream asking for an hour should not park a request handler for an
+    // hour, so what the failure asked for is still capped by the ceiling.
+    const maxMs = backoff?.maxMs ?? 30_000;
```
dunxonu: The `30_000` default for `maxMs` used to cap a `Retry-After`-supplied delay is a second, independent copy of the same literal default in `backoff.ts`.

**How it fails:** `backoff.ts:28` defaults `maxMs = 30_000` for the computed backoff; `policy.ts:77` independently writes `backoff?.maxMs ?? 30_000` to cap an upstream-requested (`Retry-After`) delay. If a future change updates one default (e.g. to tune the ceiling repo-wide) without updating the other, computed backoff and header-driven delay silently diverge in their ceiling, an inconsistency invisible to any test that does not compare both paths.

packages/core/src/resilience/policy.ts
```diff
@@ -0,0 +1,92 @@
...
+    const asked = verdict.delayMs;
+    return {
+      retry: true,
+      delayMs: asked === undefined ? computed : Math.min(asked, maxMs),
+    };
+  }
+
+  #signal(): AbortSignal {
```
dunxonu, line 106: `#signal()` always builds a combined signal via `AbortSignal.any([...])`, even for the common case of `timeoutMs: 0` and no caller `signal`, where the pre-refactor code explicitly skipped constructing one.

**How it fails:** Every default-config `ResiliencePolicy`/`HttpService` call (no per-call timeout override, no caller signal) now allocates an `AbortSignal.any([])` wrapper per attempt instead of passing `undefined` to `fetch`, as the code did before this PR (`signals.length === 0 ? {} : { signal: AbortSignal.any(signals) }` in the old `service.ts`). Functionally harmless but a needless allocation on the hot path of every outbound request with no timeout configured.

examples/full/src/upstream/upstream.demo.ts
dunxonu: The ResiliencePolicy demo never threads the per-attempt AbortSignal into the wrapped HttpService call, so the policy's timeout/cancellation half is inert.

**How it fails:** `this.policy.run(() => this.http.get(url, { retry: { maxRetries: 0 } }))` ignores the `signal` argument `ResiliencePolicy.run` provides and never passes `{ signal }` to `HttpService.get`. Per docs/guide/25-resilience.md's own stated semantics ('an operation that ignores it still runs to completion, and only a fetch or a throwIfAborted() makes the budget real'), `UpstreamPolicy`'s timeoutMs never actually bounds or cancels the underlying fetch; it currently reads as working only because `HttpClientModule` and `ResilienceModule` in upstream.module.ts happen to share the same `config.get('upstream').timeoutMs` value, masking the gap. A consumer copying this pattern with a shorter outer timeout would see no cancellation at all.

dunxonu (commented):

petarzarkov:
All three findings were real. Fixed in b9cc5a3.

**The demo dropped the attempt signal** (two comments on `upstream.demo.ts:93`). Reproduced first: a policy with `timeoutMs: 150` around `http.get('api/upstream/slow')` with the signal dropped resolved `{"done":true}` after 310 ms; threading it aborted at 151 ms. Fixed in the demo, the vendored template and the suite. `UpstreamPolicy`'s budget is now 150 ms rather than the client's own `upstream.timeoutMs`, so the masking the comment describes is gone, and the demo's third step calls the 300 ms route with `timeoutMs: 0` on the client, leaving the policy's signal as the only thing that can cancel it. Two regression guards: `examples/full/src/resilience.test.ts` fails on `expect(answered.done).toBeUndefined()` when the signal is dropped (verified by dropping it), and `tour.test.ts` covers the demo's own narration, which raises rather than logging if the call completes. The guide gained one line on the same point.

**The duplicated `30_000`.** One `DEFAULT_MAX_DELAY_MS` in `backoff.ts`, used by `backoffDelay`'s default and by the `Retry-After` cap. There is no failing behaviour to reproduce for a duplicate, so the guard is a test that pins the header-driven cap to `backoffDelay`'s own default rather than to a second literal.

**`AbortSignal.any([])` per attempt.** Correct, and measured before changing it: `AbortSignal.any([])` is 237 ns against a 40 us localhost `fetch`, so it is under run-to-run noise end to end rather than a hot-path cost. Changed anyway, since it is smaller code: `#signal()` returns the one signal it has, or a process-wide never-aborting one, and calls `AbortSignal.any` only when there are two to combine. A shared never-aborting signal was checked for listener accumulation across 8000 fetches and showed none.

One thing not asked for: the two policy tests moved out of `service.test.ts` into `examples/full/src/resilience.test.ts`, because the new case took that file to 808 of the 800-line cap.

`bun run ci` is green, 13 of 13. `@dunx/core` at 95.3% lines, 91.2% functions.

dunxonu (commented):
Fine. This is a well-tested, clean PR that already fixed its own bug in the second commit. Based on my thorough manual review across correctness, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, and CLAUDE.md conventions, I did not find any surviving defect worth flagging with high confidence.

Final findings:

```json
[]
```

packages/core/src/resilience/module.ts
```diff
@@ -0,0 +1,174 @@
...
+  provide,
+  type AsyncModuleConfig,
+  type Deps,
+  type FactoryProvider,
+} from '../di/provider.js';
+import { token, type Ctor, type Token } from '../di/token.js';
+import { ResilienceOptions, type ResilienceOptionsInit } from './options.js';
+import { ResiliencePolicy } from './policy.js';
+
+const tokens = new Map<string, Token<ResiliencePolicy>>();
```
dunxonu, line 12: resiliencePolicy(name)'s Map-based token memoization duplicates the same pattern already in httpClient(name) and redisConnection(name)

**How it fails:** The exact `Map<string, Token<T>>` memoization shape now exists three times (packages/http/src/client/module.ts's httpClient, packages/infra/src/redis/module.ts's redisConnection, and this new resiliencePolicy) with no shared helper; the doc comment even says 'the same shape as httpClient(name), which documents it at length' without extracting it, so a fix to the memoization pattern (e.g. a name-collision guard) has to be applied in three places.

packages/core/src/resilience/module.ts
```diff
@@ -0,0 +1,174 @@
...
+   * ```ts
+   * export class PaymentPolicy extends ResiliencePolicy {}
+   * ResilienceModule.forRoot({ timeoutMs: 3_000 }, PaymentPolicy);
+   *
+   * class Checkout {
+   *   constructor(private readonly policy: PaymentPolicy) {}
+   * }
+   * ```
+   */
+  static forRoot(
```
dunxonu, line 105: forRoot and forRootAsync each hand-build an identical unnamed DynamicModule object literal (module/exports/providers) instead of sharing one helper

**How it fails:** Lines 113-121 and 161-172 duplicate `{ module: ResilienceModule, exports: [ResilienceOptions, ResiliencePolicy], providers: [...] }` with only the options-provider shape differing; a future change to what the unnamed module exports or provides has to be edited in both places, the way namedModule was already factored out for the named case.

dunxonu (commented):

petarzarkov:
All five findings were real. Each was reproduced before fixing.

**`onSuccess` inside the operation's `try`** - the serious one. A throwing success hook was caught by the attempt's own `catch`, classified as a failed attempt, and the operation ran again. Measured at **four calls for an operation that succeeded every time**, which for the charge in the guide is four charges. The `try` now holds the call and nothing else.

**`fallback` wrapping the whole loop** - measured returning `FALLBACK` with the operation called **zero** times, because `onAttempt` throws before the operation runs. It now applies only where an attempt terminally failed; a callback or classifier error propagates as itself. Four regression tests, including one pinning that a real operation failure still reaches the fallback.

**`readBody` swallowing a failed read** - real. A 2xx whose stream broke reached the caller as a success with no data. It rejects now. The two error paths in `HttpService` opt back into tolerating it, because there the status is the signal and an unreadable body should not replace a 4xx with a read error.

**The SSE connection timeout** - real, and it contradicted the method's own doc comment. Proved it: an `AbortSignal.timeout(200)` handed to `fetch` let headers through and then killed a 600 ms stream with `TimeoutError`. Replaced with a `ConnectDeadline` cleared when `fetch` resolves, so it can only ever end the connect while the caller's signal still cancels the body.

**Idempotency in the guide** - fair, and worth more than a footnote given the first bug above. The example now carries the caveat: send an idempotency key, or narrow the classifier to failures that cannot have landed.

`service.ts` crossed the 500-line cap during this, so the SSE parser and `ConnectDeadline` moved to `client/sse.ts` with tests of their own. Full `bun run ci` green, 13 of 13.

packages/core/src/resilience/policy.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  #decide(
+    error: unknown,
+    attempt: number,
+  ): { readonly retry: boolean; readonly delayMs: number } {
+    const { retryDelayMs = 1000, backoff } = this.options.retry;
+    const verdict = this.options.classifier.classify(error);
+    if (!verdict.retry) return { retry: false, delayMs: 0 };
+
+    const computed = backoffDelay(attempt, {
```
dunxonu, line 90: #decide computes backoffDelay unconditionally even when the classifier already supplied an explicit delayMs that discards it.

**How it fails:** On a retry where the classifier returns a Retry-After-derived `delayMs` (e.g. HttpRetryClassifier honoring a 429's header, the common production case), `computed = backoffDelay(...)` still runs - drawing from `crypto.getRandomValues` and doing the exponent math - and its result is thrown away one line later because `asked !== undefined`. Moving the `backoffDelay` call inside the `asked === undefined` branch skips the wasted work on every Retry-After-driven retry.

dunxonu (commented):

3 further findings are in code untouched since the last review, and so were reportable then. Not repeated here.

Issue #94, opened 2026-09-12
Add gRPC integration


yibird:
# Add gRPC integration

I think Dunx could provide first-class gRPC integration.

gRPC is widely used for service-to-service communication in modern backend applications, especially in microservice architectures. Providing gRPC integration would make Dunx easier to use in distributed systems and would also allow Dunx applications to communicate with existing gRPC services.

Ideally, Dunx could provide a simple integration with the existing DI and module system, for example:

```ts
@Module({
  imports: [
    GrpcModule.forRoot({
      protoPath: './proto/user.proto',
      package: 'user',
      url: 'localhost:50051',
    }),
  ],
})
export class AppModule {}
```

And expose gRPC services through a controller-like API:

```ts
@GrpcService('UserService')
export class UserService {
  @GrpcMethod('GetUser')
  getUser(request: GetUserRequest) {
    // ...
  }
}
```

It would also be useful to support injecting gRPC clients through DI:

```ts
@Injectable()
export class OrderService {
  constructor(
    private readonly userClient: UserClient,
  ) {}

  async getUser(id: string) {
    return this.userClient.getUser({ id })
  }
}
```

The integration could provide:

* gRPC server support
* gRPC client support
* Protocol Buffers (`.proto`) integration
* Unary RPC
* Server/client/bidirectional streaming
* DI integration
* Module-based configuration
* Graceful shutdown

The goal would not necessarily be to implement a new gRPC stack from scratch, but to provide a clean Dunx integration around an existing Bun-compatible gRPC implementation.

I think this would fit well with Dunx's existing DI, module, controller, and lifecycle architecture, while making Dunx a more complete choice for both HTTP APIs and service-to-service communication.

petarzarkov:
Taking this, with a scope that is narrower than the issue asks for, and the reason is a Bun gap rather than a design preference.

**dunx already serves HTTP/2.** `HttpOptions.http2` serves h2c on the same port as HTTP/1.1, through the same routes and the same `fetch` fallback, and there is a test asserting a `node:http2` client and an HTTP/1.1 client get the same routes off one port. So the transport half of this is not new work.

**What blocks native gRPC is trailers.** gRPC carries its status out of band, in an HTTP trailer, and `Bun.serve` has no trailer surface. Probed on Bun 1.4.2:

```
Response has .trailers?  false
h2 status: 200 | trailer hdr: grpc-status
client saw trailers: null
```

[...]

PR #99, opened 2026-09-12
feat(http): serve Connect and gRPC-Web behind @dunx/http/connect

petarzarkov:
Addresses #94, scoped to Connect and gRPC-Web. Native gRPC is out: `grpc-status`
travels in an HTTP trailer and `Bun.serve` sends none (re-probed on Bun 1.4.2), so
`grpc` is hard-coded off and `content-type: application/grpc` gets a 415 that says why.

**Placement.** A subpath of `@dunx/http`, not a new package. That is what
`internal/notes/research/rpc.md` already decided: build it "in `@dunx/http` behind
`./connect`" once an external issue arrived **and** `Bun.serve` gained HTTP/2. Both
fired, so the note is updated rather than contradicted. It also keeps `./rpc` free for
the JSON-RPC design in the same note, and avoids a workspace with its own 90% floor.

**Library surface.** `createFetchHandler` from `@connectrpc/connect/protocol`, a subpath
in the package's own `exports` map and the only thing taken from it; the official
`@connectrpc/connect-node` adapter imports the same subpath. Everything else
(`createConnectRouter`, `ServiceImpl`, `ConnectError`) is the package root.

Worth a second look:
- Optional peers on `@dunx/http`, which every app installs. The subpath split holds:
  `dist/index.js` contains no `@connectrpc` reference.
- `imports` on `ConnectOptionsInit`: the implementation classes are constructed in
  `ConnectModule`'s own scope, so they need it to inject anything.
- `scripts/no-slop.test.ts` now skips `*_pb.ts`. `protoc-gen-es` output is 53% comments
  copied out of the `.proto`, which alone put `examples/` over its density budget.

`examples/full/src/rpc` calls it over the wire with both transports.

## Summary by CodeRabbit

* **New Features**
  * Added Connect and gRPC-Web support for protobuf-based unary and server-streaming RPCs.
  * RPC services can share the same port as standard HTTP routes, including JSON requests and streaming responses.
  * Added configurable RPC prefixes, service registration, graceful shutdown, and request cancellation support.
  * Middleware-served paths are now rate-limited like regular routes.

* **Documentation**
  * Added a comprehensive RPC integration guide and updated package documentation.

* **Bug Fixes**
  * Prevented route and RPC path collisions.
  * Native gRPC requests now receive a clear unsupported response.


packages/http/src/connect/registry.ts
```diff
@@ -0,0 +1,94 @@
...
+      // `connectService` checked this pairing where both types were known.
+      router.service(
+        registration.service,
+        implementation as ServiceImpl<DescService>,
+      );
+    });
+
+    for (const handler of router.handlers) {
+      const path = `${options.prefix}${handler.requestPath}`;
+      this.#routes.set(path, createFetchHandler(handler));
```
dunxonu: ConnectRegistry builds its path -> handler Map with a plain `.set` and no collision check, unlike every other path table in the package.

**How it fails:** Two `connectService(...)` registrations that resolve to the same mounted path (e.g. a duplicate/renamed service, or two descriptors sharing a fully-qualified method name) silently overwrite each other in `#routes` - the second wins with no error naming either registration. Every other path table in this package (`assertNoCollisions` and `assertNoGatewayCollisions` in packages/http/src/server/routes.ts, and the manual check in packages/http/src/ws/runtime.ts) throws a boot error naming both owners on collision; ConnectRegistry boots successfully and serves the wrong implementation for the shadowed RPC.

packages/http/src/connect/options.ts
```diff
@@ -0,0 +1,99 @@
...
+        'ConnectModule.forRoot was given no services. Pass at least one ' +
+          'connectService(Desc, Impl), or drop the module.',
+      );
+    }
+  }
+}
+
+/** A leading slash and no trailing one. An empty prefix stays empty, which is
+ * where a stock Connect client looks. */
+export const normalizeConnectPrefix = (prefix: string): string => {
```
dunxonu: `normalizeConnectPrefix` duplicates the split/filter/join algorithm already in `normalizePrefix` (packages/http/src/static/options.ts), differing only in the empty-prefix result.

**How it fails:** A future fix to slash-normalization (e.g. handling `//`, `.` segments, or an unusual separator) applied to `static/options.ts`'s `normalizePrefix` in response to a bug report will not be applied to `connect/options.ts`'s copy, since nothing ties the two together, causing StaticModule and ConnectModule to silently normalize prefixes differently going forward.

packages/http/src/connect/middleware.ts
```diff
@@ -0,0 +1,58 @@
...
+import type { RouteContext } from '../server/context.js';
+import type { Middleware, Next } from '../server/middleware.js';
+import { HttpStatusCode } from '../server/status.js';
+import { ConnectRegistry } from './registry.js';
+
+/** `application/grpc`, with or without a `+proto` suffix. `application/grpc-web`
+ * is a different media type and does not match. */
+const NATIVE_GRPC = /^application\/grpc(?:$|\+)/;
+
+const isNativeGrpc = (contentType: string | null): boolean =>
```
dunxonu: `isNativeGrpc` allocates an array and two intermediate strings via `split`/`trim`/`toLowerCase` on every RPC request instead of a single anchored regex test.

**How it fails:** On an RPC-heavy service, every successfully-routed Connect/gRPC-Web call (the common case, not the rare native-gRPC one) pays for a `split(';')` array allocation plus `.trim()` and `.toLowerCase()` string allocations just to feed `NATIVE_GRPC.test(...)`. A single case-insensitive anchored regex (e.g. `/^application\/grpc(?:$|\+|;)/i`) against the raw header does the same classification with zero allocation.

dunxonu (commented):

<details><summary>Outside the diff (2)</summary>

- `packages/http/src/throttle/guard.ts:43` ThrottleGuard unconditionally skips rate limiting for every Connect/gRPC-Web RPC call, contradicting the feature's own documentation.
- `packages/http/src/server/metrics.ts:99` Every RPC call is bucketed into the generic `(unmatched)` metrics series, indistinguishable from real 404 traffic or from other RPC methods.

</details>

packages/http/src/connect/middleware.ts
```diff
@@ -0,0 +1,70 @@
...
+  readonly #registry: ConnectRegistry;
+  readonly #server: ServerRef;
+
+  constructor(registry: ConnectRegistry, server: ServerRef) {
+    this.#registry = registry;
+    this.#server = server;
+  }
+
+  handle(req: BunRequest, ctx: RouteContext, next: Next): Promise<Response> {
+    if (ctx.get(UNMATCHED) !== true) return next();
```
dunxonu, line 57: An HTTP controller route and a Connect RPC method can silently collide on the same path, with no boot-time check, because ConnectMiddleware only inspects paths Bun's router already failed to match.

**How it fails:** A team declares a @Post controller route at the literal path an RPC would occupy (e.g. '/greet.v1.GreetService/Say'), perhaps by accident when adding both HTTP and RPC surfaces for the same domain. assertNoCollisions (packages/http/src/server/routes.ts) only checks HTTP routes against each other, assertNoGatewayCollisions only checks websocket gateway paths, and ConnectRegistry never cross-references either. The HTTP route matches first and wins forever; ctx.get(UNMATCHED) is never true for that path, so ConnectMiddleware.handle always calls next() and the RPC endpoint is permanently unreachable with zero boot-time diagnostic - unlike every other collision case in the framework, which fails loudly.

packages/http/src/connect/registry.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    for (const handler of router.handlers) {
+      const path = `${options.prefix}${handler.requestPath}`;
+      const rpc = `${handler.service.typeName}.${handler.method.name}`;
+      // The same contract as `assertNoCollisions` for routes: two services
+      // reaching one path is a boot error naming both, never a silent overwrite
+      // that serves whichever registered last.
+      const taken = this.#methods.find((method) => method.path === path);
+      if (taken !== undefined) {
+        throw new Error(
```
dunxonu: Boot-time misconfiguration in the new connect module throws plain Error instead of AppError, breaking the framework-wide convention every other boot-time wiring failure follows.

**How it fails:** Two services mounted at the same RPC path, or ConnectOptions given zero services / both protocols disabled, throw `new Error(...)` (registry.ts:59,79; options.ts:83,90) instead of `new AppError(...)`. packages/core's defaultErrorMapper and injector.ts special-case `instanceof AppError` to attach status/message handling and contextual consumer/module naming when boot fails; a plain Error from ConnectModule loses that context and formatting, and any code (or test) that discriminates dunx boot errors via `instanceof AppError` will not recognize a misconfigured ConnectModule as one.

packages/http/src/connect/middleware.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (route === undefined) return next();
+
+    const contentType = req.headers.get('content-type');
+    if (contentType !== null && NATIVE_GRPC.test(contentType)) {
+      return Promise.resolve(grpcUnsupported());
+    }
+
+    // A streaming RPC may legitimately pause longer than `idleTimeout` between
+    // messages, and Bun would cut the socket mid-stream. See ServerRef.keepAlive.
+    if (route.streaming) this.#server.keepAlive(req);
```
dunxonu: ServerRef.keepAlive permanently disables Bun's idle timeout for a streaming RPC connection with no re-arm, and that same RPC path is already exempt from ThrottleGuard, leaving no dunx-level cap on how long or how many such connections a client can hold open.

**How it fails:** A client (or a bug) opens many server-streaming RPC calls and never closes them: `this.#server.keepAlive(req)` calls `timeout(req, 0)` once on the first request, which fully disables Bun's 10s idle cut for that connection's lifetime and is never re-enabled. Combined with the middleware's own documented fact that ThrottleGuard returns early on every unmatched (RPC) path, a streaming RPC is exempt from both the idle timeout and the rate limiter, so nothing in the framework bounds how many such connections accumulate or how long each lives.

packages/http/src/connect/registry.ts
```diff
@@ -0,0 +1,115 @@
...
+      );
+    });
+
+    for (const handler of router.handlers) {
+      const path = `${options.prefix}${handler.requestPath}`;
+      const rpc = `${handler.service.typeName}.${handler.method.name}`;
+      // The same contract as `assertNoCollisions` for routes: two services
+      // reaching one path is a boot error naming both, never a silent overwrite
+      // that serves whichever registered last.
+      const taken = this.#methods.find((method) => method.path === path);
```
dunxonu: ConnectRegistry reimplements the duplicate-path collision check that packages/http/src/server/routes.ts's assertNoCollisions already solves, with a different data structure, message wording, and error type, rather than sharing one policy.

**How it fails:** assertNoCollisions (routes.ts) builds a Map and throws AppError naming both owners on a duplicate HTTP route; ConnectRegistry's constructor does the equivalent for RPC paths with a linear `#methods.find` scan and a plain `Error` with its own wording. The two are already inconsistent on error type (see finding 2) precisely because they were written twice instead of once - a future change to the collision-detection contract (message format, fields included, error type) has to be made in both places to stay consistent, and one already didn't.

packages/http/src/server/server-ref.ts
```diff
@@ -0,0 +1,41 @@
...
+   * has been read. Measured on Bun 1.4.2: a handler that reads the body and then
+   * waits 12.5 seconds before its second chunk has the socket cut and the
+   * enqueue throws `Controller is already closed`; the same handler on a GET with
+   * no body survives, and an 11 second pause survives either way.
+   *
+   * So a streaming RPC, where the gap between messages is the protocol rather
+   * than a stall, clears its own deadline. The app-wide `idleTimeout` is left
+   * alone, since it is protecting every other route.
+   */
+  keepAlive(req: Request): void {
```
dunxonu: The idle-timeout-during-pause fix is wired only into ConnectMiddleware via ServerRef, leaving the identical Bun bug unaddressed for any plain controller route that manually streams a slow ReadableStream (e.g. hand-rolled SSE).

**How it fails:** A controller method returns `new Response(readableStream)` and pauses more than ~11-12.5s between writes after the request body has been read - the exact scenario server-ref.ts's own comment documents as losing the connection (`ECONNRESET`, 'Controller is already closed'). RouteContext (packages/http/src/server/context.ts) exposes no timeout-control API, and ServerRef is wired only through ConnectMiddleware's constructor injection, so this class of route hits the same bug this PR fixes for Connect, with no documented pattern or example showing how to work around it.

dunxonu (commented):
**Actionable comments posted: 5**

dunxonu (commented):
You've hit your session limit · resets 8:10am (UTC)

packages/http/src/connect/registry.ts
```diff
@@ -0,0 +1,117 @@
...
+      // `connectService` checked this pairing where both types were known.
+      router.service(
+        registration.service,
+        implementation as ServiceImpl<DescService>,
+      );
+    });
+
+    // The same policy routes get, from the same class: a second claim on one
+    // path is a boot error naming both, never a silent overwrite.
+    const claims = new PathClaims(
```
dunxonu, line 82: RPC paths are never cross-checked against HTTP controller routes for collisions, so a controller silently shadows an RPC at the same literal path.

**How it fails:** Register a controller route (e.g. `@Get('/greet.v1.GreetService/Say')`) alongside `ConnectModule.forRoot({ services: [connectService(GreetService, GreetRpc)] })`. Boot succeeds with no error from `assertNoCollisions` (server/routes.ts:82) or `ConnectRegistry`'s own `PathClaims('RPC', ...)`, because they operate on disjoint path sets. At request time Bun's route table matches the controller first, `ctx.get(UNMATCHED)` in `ConnectMiddleware.handle` (connect/middleware.ts:52) is never true, and the RPC is permanently unreachable with no diagnostic.

packages/http/src/index.ts
```diff
@@ -21,6 +21,7 @@ export {
   PUBLIC,
   Roles,
   ROLES,
+  REQUEST_SERVER,
```
dunxonu: REQUEST_SERVER, a metadata key meant only for ConnectMiddleware's internal use on the unmatched path, is exported from the main @dunx/http entrypoint.

**How it fails:** `ConnectMiddleware` imports `REQUEST_SERVER` directly from `../route/metadata.js` and never needs the index re-export, yet the main package now publicly surfaces a raw Bun `Server` handle (`REQUEST_SERVER: MetaKey<Server<unknown>>`) to every consumer of `@dunx/http`, even those who never install the optional `/connect` peer deps - a capability tied to one optional subpath leaking into the always-loaded public surface.

dunxonu (commented):

- `packages/http/src/server/routes.ts:99` assertNoGatewayCollisions still hand-rolls a Set-based claim check instead of using the PathClaims class this same PR introduced three lines above it to unify exactly this pattern.

</details>

dunxonu (commented):
That's a generated corpus file plus README updates, both fine. My review is complete. Given the thorough manual pass across all 8 angles found no confirmed defects — the PR is well-tested and already incorporates fixes from earlier review rounds in its own commit history.

```json
[]
```

No findings survived verification. Summary of what was checked: the new `@dunx/http/connect` mount (registry, middleware, options, module), the `PathClaims`/`ClaimedRoutes`/`assertNoShadowedClaims` path-collision machinery threaded through `application.ts`, `routes.ts`, `throttle/guard.ts`, the `ServedHandler`/`REQUEST_SERVER` plumbing for clearing Bun's idle timeout on streaming RPCs, and the `examples/full` RPC feature wiring. Cross-checked the "any-method shadowing" logic in `assertNoShadowedClaims` against Bun's documented per-path method-miss 404 behavior (confirmed correct, not a false positive), verified `ConnectRegistry.onShutdown` is invoked automatically by the DI container's generic `onShutdown` lifecycle hook rather than needing explicit wiring, and confirmed CORS/server-param threading (`withCors`) correctly forwards the `server` argument needed for the timeout fix. No correctness, cleanup, altitude, or CLAUDE.md convention violations were confirmed.

PR #100, opened 2026-09-12
feat(core): an application EventBus over EventTarget

### LINKED ISSUE #93: Add an application EventBus
Add an application EventBus

I think Dunx should provide a built-in application event system (EventBus).

In mainstream backend frameworks such as Spring Boot, application events are a common way to decouple different parts of an application. Instead of having one service directly depend on another service, a service can publish an event and let other components react to it.

For example:

eventBus.emit(new UserCreatedEvent(user))

And another component can subscribe to it:

@OnEvent(UserCreatedEvent)
async handle(event: UserCreatedEvent) {
  // ...
}

This could be useful for scenarios such as:

Decoupling modules
Audit logging
Notifications
Sending WebSocket events
Triggering background jobs
Integrating with message queues

Ideally, the EventBus could have a simple in-process implementation by default, while allowing other implementations such as Redis or a message queue to be used when distributed event processing is needed.

I think an EventBus would fit naturally into Dunx's existing DI and module system and provide a simple, framework-level mechanism for application-level decoupling.

petarzarkov:
Closes #93.

`EventBus`, `AppEvent`, `@OnEvent`, `EventRegistry` and `EventBusModule` in `@dunx/core`, over one `EventTarget`. Zero dependencies still.

Two decisions worth a look:

- **`emit()` awaits handlers, and never rejects.** `dispatchEvent` discards a listener's returned promise, so the bus collects it: an `async` handler is awaited without `waitUntil`, and `waitUntil` stays for a sync handler that starts work it does not return. A throwing subscriber is counted on its subscription, logged at `error`, and carried on the returned `EventDispatch`; `throwIfFailed()` opts into rethrowing.
- **The module-graph walk moved down.** `discoverMarked` in `@dunx/core` replaces the copy in `@dunx/infra`'s job and schedule discovery, and both now resolve each handler through the module that declares it rather than through `app.get()` alone.

No Redis or queue backend: durable events stay `@JobHandler`, cross-node fan-out stays `PubSubRelay`, and [the guide](docs/guide/26-events.md) says so.

`examples/full/src/events/` is example-only; no `create-app` feature names it.

## Summary by CodeRabbit

- **New Features**
  - Added an in-process event system with event publishing, asynchronous subscribers, one-time subscriptions, cancellation, and delivery results.
  - Added decorator-based event handlers and subscription monitoring, including handled and failed delivery counts.
  - Added a lifecycle phase that wires event handlers before initialization.
  - Added a complete events example with order notifications, audit records, and subscription status endpoints.

- **Documentation**
  - Added an Events guide and updated lifecycle documentation and navigation.

- **Tests**
  - Added coverage for asynchronous delivery, failures, nested events, cancellation, and initialization ordering.


packages/core/src/events/bus.ts
```diff
@@ -0,0 +1,212 @@
...
+    this.#target.addEventListener(
+      type,
+      (raw: Event) => {
+        this.#deliver(
+          subscription,
+          handler as EventHandler<AppEvent>,
+          raw as AppEvent,
+        );
+      },
+      { signal, once: options.once === true },
```
dunxonu: A `once: true` subscription is removed by the native `EventTarget` `once` option, but the subscription's own `AbortController` is never aborted, so `EventSubscription.active` reports `true` forever after the handler has already fired and been unregistered.

**How it fails:** `const sub = bus.once(OrderPlaced, fn)` (or `@OnEvent(X, { once: true })`, wired the same way in registry.ts); after one `emit`, the listener is gone (fn will never run again) but `sub.active` still returns `true` indefinitely. Any code that checks `active` to decide whether a handler is still live (dashboards, tests, `EventRegistry` consumers) is misled; `examples/full/src/events/notifications.service.ts`'s `@OnEvent(OrderSettled, { once: true })` handler produces exactly this state with no test covering it.

packages/infra/src/queue/discover.ts
```diff
@@ -95,26 +94,13 @@ export const assertNoDuplicateJobs = (
  */
 export const discoverJobs = (
   modules: readonly ResolvedModule[],
-  resolve: (token: InjectionToken<unknown>) => unknown,
+  resolve: (token: InjectionToken<unknown>, from: ModuleRef) => unknown,
```
dunxonu: `discoverJobs`/`selectJobs`'s `resolve` signature was widened to `(token, from) => unknown` for module-scoped resolution (matching `discoverMarked`'s new contract, and mirroring what `ScheduleRunner` was updated to do), but every real call site still passes a single-argument resolver that ignores `from`.

**How it fails:** `packages/infra/src/queue/worker.ts:427` and `:470`, and `packages/infra/src/queue/processor.ts:114`, all call `selectJobs(modules, (token) => app.get(token), ...)`. A job-handler class bound differently in two modules (the exact scenario `packages/core/src/di/marked.ts`'s `discoverMarked` doc says this parameter fixes, and which `registry.test.ts` explicitly tests for events) still resolves via whichever scope `app.get(token)` answers first for queue jobs, silently defeating the new module-scoped guarantee the widened signature implies is now honored.

docs/guide/26-events.md
```diff
@@ -0,0 +1,148 @@
...
+`active`.
+
+Discovery walks the prototype chains of the classes each module declares in
+`providers` and `controllers`, so a handler needs no second registration and an
+abstract base's marked methods are inherited by every subclass. A value or
+factory provider is not scanned: put handlers on a class provider.
+
+## Limits
+
+Handlers are wired during `onInit`. An event emitted from a constructor reaches
```
dunxonu: The guide states "Handlers are wired during `onInit`. An event emitted from a constructor reaches nobody," implying emitting from `onInit` or later is safe, but subscription wiring order depends on module-import order (`EventBusModule`'s `EventRegistry.onInit()` runs only after providers earlier in that order have already had their own `onInit` called).

**How it fails:** `@Module({ imports: [OrdersModule, EventBusModule] })` where a service in `OrdersModule` emits an event from its own `onInit`: because `OrdersModule` precedes `EventBusModule` in import order, `injector.instances` (and therefore `onInit` calls, per `packages/core/src/di/app.ts:316-317`) reaches `Orders` before `EventRegistry` has subscribed anything, so the emit reaches zero listeners and `EventDispatch.ok` still reports `true` - silently dropped, exactly the case the guide's wording says is already covered by moving past the constructor.

dunxonu (commented):

packages/core/src/events/bus.ts
```diff
@@ -0,0 +1,219 @@
...
+        if (options.once === true) controller.abort();
+        this.#deliver(
+          subscription,
+          handler as EventHandler<AppEvent>,
+          raw as AppEvent,
+        );
+      },
+      { signal, once: options.once === true },
+    );
+
```
dunxonu, line 113: EventBus.on() passes both `once: true` to addEventListener and manually calls controller.abort() inside the listener for the same subscription, so two independent mechanisms perform the same single-shot removal.

**How it fails:** The manual `controller.abort()` (needed to make `subscription.active` report false, since native `once` leaves the controller untouched) already deregisters the listener via the combined AbortSignal - the `once: true` flag passed alongside it does no additional work and only obscures which mechanism is actually responsible for removal.

packages/core/src/events/module.ts
```diff
@@ -0,0 +1,41 @@
...
+    }),
+    // Bound so the container constructs it, which is what gets `onInit` called.
+    provide(EventRegistry, {
+      useFactory: (...deps: readonly unknown[]) =>
+        new EventRegistry(
+          ...(deps as ConstructorParameters<typeof EventRegistry>),
+        ),
+      inject: EVENT_REGISTRY_DEPS,
+    }),
+  ],
```
dunxonu, line 43: EventRegistry's provide() factory takes `...deps: readonly unknown[]` and casts through `ConstructorParameters<typeof EventRegistry>` instead of being typed directly, unlike EventBus's factory three lines above which is fully typed.

**How it fails:** A future change to EventRegistry's constructor signature (add, remove, or reorder a parameter) is not caught by the compiler here - the cast silently accepts a mismatched deps array, and the error would surface only at runtime as a wrong-typed argument inside the constructor.

dunxonu (commented):

- `packages/core/src/di/app.ts:337` AppRef's doc comment and thrown error text say it is usable only "from onInit onwards", but AppFactory.create now attaches it before the onBeforeInit pass too, and EventRegistry.onBeforeInit() reads AppRef.current successfully one phase earlier.

</details>

packages/core/src/events/bus.ts
```diff
@@ -0,0 +1,233 @@
...
+  ): EventSubscription {
+    return this.on(event, handler, { ...options, once: true });
+  }
+
+  /**
+   * Delivers `event` to every subscriber, then resolves once their work has
+   * settled. It does not reject: read `failures`, or call `throwIfFailed()`.
+   */
+  async emit<E extends AppEvent>(event: E): Promise<EventDispatch> {
+    const label = event.constructor.name;
```
dunxonu, line 131: EventDispatch.event and the boot/error log lines use event.constructor.name, the exact value eventName() deliberately avoids using as the identity key because a minifier rewrites it

**How it fails:** If a consumer's build step minifies class names (even though dunx itself ships unminified), two AppEvent subclasses can collapse to the same mangled constructor.name; dispatch and subscription routing still work correctly (eventName's WeakMap id is untouched), but EventDispatch.event, the 'X failed handling Y' error log, and EventRegistry's 'Subscribed N handler(s)' boot log all report the collided name, making it impossible to tell which event actually fired from logs or from a controller that surfaces dispatch.event in a response.

dunxonu (commented):

4 further findings are in code untouched since the last review, and so were reportable then. Not repeated here.

dunxonu (commented):
Good, `list()` and `subscribersOf` are exercised. Not an issue.

Given the depth of independent review passes converged on no strong correctness bugs (the event bus and lifecycle logic checked out correct under trace-through, and behavior changes are documented/tested), I'll finalize with the few defensible low-severity items.

```json
[
  {
    "file": "packages/core/src/di/marked.ts",
    "line": 171,
    "summary": "discoverMarked walks each candidate's prototype chain twice: once in scan() to test for a marked method, once again inside markedMethodsOn to build the result.",
    "failure_scenario": "Not a functional bug, but the walk (and metaOf calls) now runs twice per handler-bearing class across three call sites (queue jobs, schedules, event handlers) that used to be one duplicated implementation each; a module graph with many providers pays this cost on every boot."
  },
  {
    "file": "packages/core/src/di/marked.ts",
    "line": 149,
    "summary": "discoverMarked's doc comment and behavior quietly change resolution from an unscoped `app.get(token)` to a per-module `container.get(token, from)`, bundled inside what reads as a pure refactor extracting duplicated discovery code from queue/discover.ts and schedule/runner.ts.",
    "failure_scenario": "A consumer of discoverJobs/selectJobs who relied on the old ambiguous-binding-throws-or-picks-arbitrary behavior (unlikely, but this is a public @dunx/core export) now gets per-module resolution instead; correctly covered by a new test (queue/discover.test.ts), so low risk, but worth a reviewer's explicit sign-off since it's a behavior change riding along with a mechanical dedup."
  }
]
```

PR #101, opened 2026-09-12
feat(http): server-sent events, and the compression bug that broke them

### LINKED ISSUE #92: Add Server-Sent Events (SSE) support
# Add Server-Sent Events (SSE) support

It would be great if Dunx could provide first-class support for Server-Sent Events (SSE).

Since Dunx already provides WebSocket support, having SSE support would make it easier to implement server-to-client streaming use cases such as notifications, progress updates, real-time logs, and AI response streaming.

Currently, SSE can be implemented manually by returning a `Response` with a `ReadableStream`, but a framework-level API could provide a much cleaner developer experience.

For example:

```ts
@Controller('/events')
export class EventsController {
  @Sse('/')
  async *events() {
    yield { data: 'hello' }
    yield { data: 'world' }
  }
}
```

Or, if `AsyncIterable` is not the preferred API:

```ts
@Sse('/events')
events() {
  return new ReadableStream({
    // ...
  })
}
```

The framework could handle the standard SSE requirements automatically, such as:

* `Content-Type: text/event-stream`
* `Cache-Control: no-cache`
* Keep-alive connections
* SSE message formatting (`data`, `event`, `id`, `retry`)
* Streaming `AsyncIterable` / `ReadableStream`
* Proper connection cleanup when the client disconnects

I think this would fit naturally into Dunx's existing Controller/Decorator API and make SSE significantly easier to use without requiring users to manually construct the underlying `Response`.

It would also be useful for AI/LLM streaming scenarios, where SSE is commonly used to stream responses from the server to the client.

petarzarkov:
Closes #92.

`@Sse(path, options)` mounts a `GET` beside `@Get`, from a handler returning an
`AsyncIterable<SseEvent>` or an `SseStream`. `SseStream` is a `ReadableStream` in a
`Response`, so the dispatcher's existing `Response` passthrough is the whole
integration and the package gains no dependency. Framing, `Last-Event-ID`,
heartbeats and teardown are in `packages/http/src/sse/`; `examples/full/src/sse/`
serves one and reads it back through `HttpService.streamSseEvents`.

Three things measured rather than assumed, each recorded in the commit message and
in `docs/bun-apis.md`: Bun holds the response headers until the body's first chunk,
so every stream opens with a comment line; a disconnect cancels the body, which is
what clears the heartbeat; and `Bun.serve` severs a request idle for 10 seconds
whether or not a response is already streaming, which `RequestTimeout.clear(req)`
now exempts. `idleTimeout` joins `HttpOptions` on the way through.

Separate fix, first commit: `app.use(Compression)` destroyed an event stream.
`isCompressibleType` took `text/event-stream` through its `text/` prefix, so
`buffer()` read the body until `done` and an event stream never reaches it.
`SseStream` also sets `no-transform`, which is the documented opt-out.

`@dunx/http/client`'s SSE parser yielded one string per `data:` line and dropped
`event`, `id` and `retry`; it is now a per-event parser, with `streamSse` the
data-only view over it.

## Summary by CodeRabbit

* **New Features**
  * Added server-sent events (SSE) support with the `@Sse` route decorator, event streams, heartbeats, reconnection via `Last-Event-ID`, and disconnect handling.
  * Added client helpers for consuming complete SSE messages or data payloads.
  * Added configurable idle timeouts for long-lived streams, including controls to keep active streams open.
  * SSE responses now bypass compression for reliable real-time delivery.

* **Documentation**
  * Added SSE usage guidance, timeout behavior, and a complete working example with notifications and resumable tick events.

* **Tests**
  * Added comprehensive coverage for SSE framing, streaming, reconnection, heartbeats, compression, and client disconnects.


packages/http/src/compression/options.ts
```diff
@@ -35,6 +35,22 @@ const COMPRESSIBLE: ReadonlySet<string> = new Set([
...
+ * the alternative: measured on Bun 1.4.2, gzip emits its 10-byte header and then
+ * nothing until the writable side closes, so 20 events over 200 ms reached the
+ * reader as 0 payload bytes.
+ *
+ * The content type is the only signal available before the body is read. A handler
+ * streaming any other type - a tailed `text/plain` log, an open `x-ndjson` feed -
+ * opts out with `cache-control: no-transform`, which `Compression` checks before
+ * this filter.
+ */
+const ENDLESS: ReadonlySet<string> = new Set(['text/event-stream']);
```
dunxonu, line 48: SSE is opted out of compression via a new hardcoded `ENDLESS` content-type set in a different package, instead of `SseStream` setting `cache-control: no-transform`, the general opt-out mechanism the same file's `#considers` check (compression.ts) already implements and documents.

**How it fails:** Any future streaming content type (e.g. an x-ndjson tailer) needs the same fix repeated as a new entry in `ENDLESS`, and if `SseStream` (packages/http/src/sse/stream.ts, `SSE_HEADERS`) is ever reused outside `@Sse` or a caller mutates `content-type` before `Compression` runs, the hardcoded exact-match set silently stops applying while `no-transform` would not have, since `Compression#considers` (compression.ts) checks `cache-control: no-transform` before the content-type filter runs at all.

packages/http/src/sse/stream.ts
```diff
@@ -0,0 +1,156 @@
...
+ * ```
+ *
+ * Every stream opens with one comment line, so the headers reach the client
+ * before the first event does.
+ *
+ * `@Sse` builds one for a handler that returns an `AsyncIterable` and takes one a
+ * handler builds itself. Outside a route it is a `Response` a `@Get` can return.
+ */
+export class SseStream {
+  readonly #encoder = new TextEncoder();
```
dunxonu: `TextEncoder` is instantiated per `SseStream` instance even though it is stateless and safely shareable.

**How it fails:** Every open SSE connection (one `SseStream` per client) allocates its own `TextEncoder` needlessly; a module-level `const encoder = new TextEncoder()` would serve every instance with no behavioral difference, and is the kind of long-lived per-connection object the review is asked to flag when it need not capture more than a shared, stateless helper.

dunxonu (commented):

- `packages/http/src/client/sse.ts:12` The existing client SSE parser (`sseData`, used by `HttpService.streamSse`) only reads `data:` lines and drops `event:`, `id:`, and `retry:` entirely, so it cannot consume the wire format `@Sse`/`frameEvent` now produces despite the PR's own claim of read/write parity.

</details>

packages/http/src/server/binding.ts
```diff
@@ -139,6 +150,8 @@ export class ServerBinding {
     const gateways = this.#gateways;
     this.#main = undefined;
     this.#gateways = undefined;
+    for (const forget of this.#forget) forget();
```
dunxonu: ServerBinding.stop() removes a server from RequestTimeout's global registry synchronously, before the graceful (non-force) Bun.serve stop has actually drained connections.

**How it fails:** A graceful shutdown (force=false) calls stop(): forget() runs at line 153 and deletes the server from request-timeout.ts's module-level `servers` Set immediately, then `main?.stop(false)` is only awaited afterward while Bun drains in-flight connections. Any `@Sse` request whose handler runs `RequestTimeout.clear(req)` (packages/http/src/sse/decorators.ts:41) during that drain window finds no matching server left in the Set, so `RequestTimeout.set` (request-timeout.ts:36) iterates zero matching servers and the clear silently no-ops - the stream stays subject to Bun's default 10s idle timeout and can be severed mid-drain instead of exiting cleanly.

packages/http/src/server/request-timeout.ts
```diff
@@ -0,0 +1,44 @@
...
+ * `Bun.serve` severs a connection idle for `idleTimeout` seconds, 10 by default,
+ * and a response already streaming is no exception: an event stream with nothing
+ * to say for ten seconds is dropped mid-flight, and the next `enqueue` throws
+ * `Invalid state: Controller is already closed`.
+ *
+ * A response meant to idle says so through this. `@Sse` calls it for every stream
+ * it answers with; a handler returning its own long-lived `Response` calls
+ * `RequestTimeout.clear(req)` for itself.
+ */
+export class RequestTimeout {
```
dunxonu: RequestTimeout works around a self-imposed gap rather than the actual Bun limitation: Bun already hands every route handler the owning server as a second argument, but RouteHandler/compose (packages/http/src/server/middleware.ts:14-31) discard it, forcing a global Set walked on every clear/set call instead of a per-request handle built once in RouteContext.

**How it fails:** packages/http/src/ws/adapter.ts:442 shows Bun's route table already invokes handlers as `(req, server) => ...` in this same table (the gateway upgrade handler uses it directly). Because `RouteHandler = (req) => Promise<Response>` (middleware.ts:14) never threads that second argument through `compose`/`buildContext`, `RequestTimeout.set` (request-timeout.ts:36) has to loop over every live server in the process and rely on Bun's foreign-request call being a silent no-op (documented as measured only on Bun 1.4.2 in docs/bun-apis.md) - a correctness property this design depends on but does not control, and which also means a request from server A spuriously calls `.timeout()` on unrelated server B/C in the same process (e.g. concurrent test servers).

dunxonu (commented):

petarzarkov:
@coderabbit review please

petarzarkov:
@coderabbit review please

petarzarkov:
@coderabbit review please

dunxonu (approved):
Reviewed what changed since the last review and found nothing worth changing.

6 further findings are in code untouched since the last review, and so were reportable then. Not repeated here.

dunxonu (commented):
No breaking call sites. This all checks out. Given the thoroughness of the manual review across all 8 angles (correctness, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, conventions), no genuine defects survive scrutiny.

```json
[]
```

Issue #85, opened 2026-09-11
Feature Request: Support Scalar as an alternative OpenAPI documentation UI


yibird:
# Feature Request: Support Scalar as an alternative OpenAPI documentation UI

Dunx currently uses Swagger UI for OpenAPI documentation. It would be great if Dunx could also support [[Scalar](https://scalar.com/)](https://scalar.com/) as an alternative documentation UI.

Scalar provides a more modern OpenAPI documentation experience and has several features that would be useful for Dunx users, while Swagger UI can remain available for existing users.

### Proposed API

Ideally, the documentation UI could be configurable:

```ts
OpenApiModule.forRoot({
  // ...
  ui: 'swagger',
})
```

or:

```ts
OpenApiModule.forRoot({
  // ...
  ui: 'scalar',
})
```

Another possible approach would be to expose Scalar as a separate module/helper:

```ts
OpenApiModule.forRoot({
  // ...
  ui: 'scalar',
})
```

The existing Swagger UI should remain the default to avoid breaking existing applications.

### Why Scalar?

* Modern and cleaner OpenAPI documentation UI
* Better developer experience
* Good support for OpenAPI 3.x
* Built-in API client / request testing experience
* More convenient navigation for large OpenAPI specifications
* Lightweight integration

Since Dunx already generates and serves the OpenAPI specification, Scalar could reuse the existing OpenAPI JSON endpoint without changing the underlying OpenAPI generation logic.

I think supporting both Swagger UI and Scalar would make the OpenAPI module more flexible while keeping the existing API compatible.

Thanks for considering this feature!

petarzarkov:
Valid, thank you

petarzarkov:
I measured this before committing to it, and the numbers changed my answer. Not a no, but not the shape proposed either.

Installed clean, on Bun 1.4.2:

| Package | Disk | Packages |
| --- | --- | --- |
| `swagger-ui-dist` (what ships today) | 12 MB | 2 |
| `@scalar/api-reference` | 276 MB | 279 |

Scalar's tree carries Vue 3, `@phosphor-icons` at 38 MB, highlight.js, zod, the `ai` SDK and `@opentelemetry`. For comparison, I already turned down `swagger-ui` (the non-`dist` package) at 177 MB across 149 packages, because it dragged in React to serve a static page. This is larger than the thing I rejected on those grounds.

[...]

PR #102, opened 2026-09-12
feat(openapi)!: split the documentation UI into optional subpaths

petarzarkov:
**BREAKING, for the next major.** Issue #85.

`@dunx/openapi` keeps document generation and `openapi.json`. The page is a
`renderer`, one optional peer per subpath: `@dunx/openapi/swagger`
(`swagger-ui-dist`) and `@dunx/openapi/scalar` (`@scalar/api-reference`). With no
`renderer`, no page route is registered.

```ts
OpenApiModule.forRoot({ title, version, root: AppModule,
  renderer: new SwaggerRenderer({ docExpansion: 'list' }) });
```

What breaks: `swagger-ui-dist` is no longer a `dependency`, so an app that mounts
Swagger UI installs it; `OpenApiInfo.ui` becomes `new SwaggerRenderer(options)`;
`SwaggerAssets`, `isSwaggerAsset`, `contentTypeOf`, `renderShell`'s old signature
and the `SwaggerUiOptions` types move off the root export.

The renderers share `renderShell`, `PackageAssets` and `DocsRenderer` on the root
export, so neither library is blessed and a third renderer needs no fork.
`docs/architecture/tooling.md` records the measurements; the CLAUDE.md carve-out
that made `swagger-ui-dist` a dependency is rewritten, since it only held while
there was one renderer.

Also fixes the stale "The page" section of `docs/guide/10-openapi.md`, which still
described the hand-built explorer deleted several releases ago.

`examples/full` mounts Swagger UI at `/api/docs` and Scalar at `/api/reference`
through a middleware, so both renderers are exercised end to end against a real
`Bun.serve`.

## Summary by CodeRabbit

* **New Features**
  * Added optional Swagger UI and Scalar renderers for OpenAPI documentation.
  * Added dedicated `/swagger` and `/scalar` configuration paths.
  * OpenAPI can serve only the JSON document when no renderer is configured.
  * Renderer assets are served locally with versioned, immutable caching.

* **Documentation**
  * Updated package, setup, architecture, security, and usage documentation for both renderer options.

* **Tests**
  * Added coverage for Scalar rendering, renderer assets, document-only mode, caching, and local asset delivery.


dunxonu (commented):

- `packages/openapi/src/swagger/options.ts:151` DUNX_KEYS is declared identically in both swagger/options.ts and scalar/options.ts instead of once at their common owner.

</details>

packages/openapi/src/swagger/renderer.ts
```diff
@@ -0,0 +1,54 @@
...
+    super();
+    this.#ui = ui;
+  }
+
+  async page(document: OpenApiDocument, options: PageOptions): Promise<string> {
+    const assets = await PackageAssets.resolve(SWAGGER_ASSETS);
+    return renderSwaggerPage(document, options, assets, this.#ui);
+  }
+
+  async asset(name: string): Promise<Response> {
```
dunxonu: asset() now resolves the underlying npm package before checking the requested name against the allow-list, reversing the old order where isSwaggerAsset(name) was checked first.

**How it fails:** Old module.ts checked `isSwaggerAsset(name)` before calling `SwaggerAssets.resolve()` (see packages/openapi/src/module.ts at commit 1cd2cc6b, lines 150-154). The new `asset()` in swagger/renderer.ts:51 and scalar/renderer.ts:53 calls `PackageAssets.resolve(SWAGGER_ASSETS)` unconditionally, then checks the allow-list inside `respond()`. If the peer (`swagger-ui-dist` or `@scalar/api-reference`) fails to resolve (not installed, or a broken install), every request under `/docs/*`, including a garbage/probe path that should be a cheap 404, now throws PackageAssets.resolve's descriptive Error instead of returning 404 - an unhandled route error where the old code path returned a clean 404 for any non-allow-listed name regardless of install state.

packages/openapi/src/assets.ts
```diff
@@ -0,0 +1,110 @@
...
+    });
+  }
+
+  /**
+   * A package's `package.json` is the honest anchor: it carries the version, and
+   * resolving an asset file directly breaks the moment upstream adds an `exports`
+   * map. Bun resolves `<package>/package.json` whether or not one is there, which
+   * `@scalar/api-reference` needs - see docs/bun-apis.md.
+   */
+  static async resolve(pkg: AssetPackage): Promise<PackageAssets> {
```
dunxonu: PackageAssets.resolve() caches only after the package.json read completes, with no in-flight-promise guard, so concurrent first requests each redo the resolve work.

**How it fails:** A cold instance receiving a page load fires several near-simultaneous requests (the page plus its bundle.js/css/favicon assets). All of them can miss the `#cache` Map before the first `resolve()` call finishes, so each independently runs `Bun.resolveSync` and `await Bun.file(manifestPath).json()` instead of sharing one in-flight resolution. assets.test.ts only asserts the sequential-call case ('caches per package, so two renderers resolve once each'), not the concurrent one, so this isn't an intentional, tested tradeoff.

packages/openapi/src/assets.ts
```diff
@@ -0,0 +1,110 @@
+import { dirname } from 'node:path';
+
+/** One year, and honest because {@link PackageAssets.href} carries the version. */
+export const ASSET_CACHE_CONTROL = 'public, max-age=31536000, immutable';
```
dunxonu: ASSET_CACHE_CONTROL duplicates the same cache-control literal already produced by @dunx/http's static file serving.

**How it fails:** `ASSET_CACHE_CONTROL = 'public, max-age=31536000, immutable'` in packages/openapi/src/assets.ts:4 is the identical string `packages/http/src/static/files.ts:71` produces for its own immutable branch. Neither imports from the other, so a future policy change (e.g. shortening max-age) made in one package's static-asset caching does not propagate to the other, and the two can silently disagree despite both claiming the same guarantee.

examples/full/src/reference/reference.middleware.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    const { pathname } = new URL(req.url);
+    if (pathname === REFERENCE_PATH) {
+      return new Response(await this.#html(), {
+        headers: { 'content-type': 'text/html; charset=utf-8' },
+      });
+    }
+    if (!pathname.startsWith(`${REFERENCE_PATH}/`)) return next();
+
+    return this.#renderer.asset(pathname.slice(REFERENCE_PATH.length + 1));
```
dunxonu, line 47: ReferenceMiddleware caches the rendered Scalar page only after the render promise resolves, so concurrent requests double-render it

**How it fails:** Two GET requests to /api/reference arrive before the first `this.#page ??= await this.#renderer.page(...)` settles; both see `this.#page` as undefined and both call `this.#renderer.page()`, rendering the same HTML twice instead of once. Result is idempotent (not wrong output) but is wasted work on every cold start under concurrent load; fix by caching the in-flight promise instead of the resolved string, the same pattern `PackageAssets.resolve()` in this same PR already applies for exactly this reason.

packages/openapi/src/explorer.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /**
+   * The page, built on the first request for a given mount prefix and cached.
+   *
+   * The renderer resolves its own assets here rather than at boot, so an app
+   * serving only `/openapi.json` never looks them up and a missing optional peer
+   * surfaces as this route failing rather than as everyone's boot error.
+   */
+  async page(prefix = ''): Promise<string> {
```
dunxonu: OpenApiExplorer.page() has the same cache-after-await race as the middleware above, carried over unfixed from the pre-PR module.ts

**How it fails:** Concurrent first requests for the docs page at the same mount prefix both miss `#pages.get(prefix)` before the render awaits, so both render the shell. Pre-existing (moved from module.ts unchanged), but the PR touches this exact method to swap in `DocsRenderer` and left the promise-caching gap that `PackageAssets.resolve()`, added in this same PR, explicitly closes for the analogous asset-resolution race.

dunxonu (commented):

petarzarkov:
@coderabbit review please

examples/full/src/reference/reference.middleware.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
```
dunxonu, line 48: ReferenceMiddleware.#html() caches the render promise but never evicts it on rejection, unlike the OpenApiExplorer.#render() fix this PR ships in the same package.

**How it fails:** If ScalarRenderer.page() rejects on the first request to /api/reference (e.g. @scalar/api-reference not yet resolvable, or a transient Bun.resolveSync/read failure), `this.#page ??= this.#renderer.page(...)` permanently pins the field to a rejected promise. Every subsequent request to /api/reference returns the same cached rejection for the rest of the process's life, even after the underlying cause is gone - exactly the bug class packages/openapi/src/explorer.ts's #render() fixes with a try/catch that does `this.#pages.delete(prefix)` before rethrowing, and whose comment states 'a cached rejection would leave the route broken for the process's life.' The example's hand-rolled second-renderer pattern (the one the guide points readers to for mounting a renderer outside OpenApiModule) doesn't carry that guard.

dunxonu (commented):

dunxonu (commented):
That file was never created (the earlier write was blocked), no cleanup needed.

```json
[
  {
    "file": "examples/full/src/reference/reference.middleware.ts",
    "line": 49,
    "summary": "ReferenceMiddleware caches a rejected render promise forever, permanently wedging /api/reference after one transient failure.",
    "failure_scenario": "`#page ??= this.#renderer.page(...)` in `#html()` never clears `#page` when the promise rejects (e.g. the optional peer `@scalar/api-reference` fails to resolve on first request, or asset resolution throws once). Every subsequent GET to /api/reference re-awaits the same rejected promise and 500s for the life of the process, even after the transient cause clears. This is the exact bug class `OpenApiExplorer.#render` (packages/openapi/src/explorer.ts:71-83) was written to avoid - it explicitly evicts `this.#pages` on rejection with the comment 'so a failure evicts itself rather than breaking the route for good' - but the example's hand-rolled second renderer reintroduces it."
  },
  {
    "file": "packages/openapi/src/scalar/options.ts",
    "line": 96,
    "summary": "renderScalarOptions and renderUiOptions each reimplement 'merge defaults, then strip SHELL_KEYS' with a different mechanism instead of one shared helper.",
    "failure_scenario": "swagger/options.ts:176 skips SHELL_KEYS inline while iterating merged entries; scalar/options.ts:96 does `for (const key of SHELL_KEYS) delete merged[key]` right after merging. Both files already import SHELL_KEYS from shell.ts, so a third renderer (or a change to what SHELL_KEYS contains) has two divergent places to keep in sync instead of one `stripShellKeys(merged)` helper in shell.ts."
  }
]
```

dunxonu (commented):
Refuted. Given none of the candidates survive as high-confidence, actionable findings, this PR review yields no strong findings.

```json
[]
```

**Summary:** Reviewed PR 102 (dunx openapi renderer subpath split) across 8 finder angles (correctness, removed-behavior, cross-file, reuse/simplification/efficiency, altitude/conventions). The refactor is a faithful decomposition: `packages/openapi/src/{renderer,explorer,controller,assets,shell}.ts` and the new `swagger/`, `scalar/` subpackages preserve every invariant from the deleted `html.ts`/`swagger.ts`/`ui-options.ts` (escaping, cache-control, allow-list 404s, promise-based render caching with correct rejection eviction). The `swagger-ui-dist` peerDependency move (from a plain `dependency`) is intentional and CLAUDE.md was updated in the same diff to match, so it is not a rule violation. `@Controller()` applied to both a controller base and its subclass in `packages/openapi/src/controller.ts` does not cause double route registration (verified against `packages/http/src/route/discover.ts` and `marker.ts`). A few sub-precision observations surfaced (a theoretical `PackageAssets` cache-key collision on shared package names, three copies of a small promise-caching idiom, an unbounded per-prefix `Map` cache carried over unchanged from the pre-PR `OpenApiExplorer`) but none are triggered by any reachable code path in this repo or introduced by this change, so none clear the bar for an actionable finding.

PR #108, opened 2026-09-12
ci: take v1.3.0, which reviews the code rather than the patch

petarzarkov:
The reviewer now gets each changed file whole with the change marked inside it, plus that file's test and its local imports as background, instead of GitHub's three-line patch.

Worth noting this pull request cannot demonstrate the difference: it is a one-line version bump, and there is no surrounding code for the extra context to matter to. The signal will come from the next substantive change.

## Summary by CodeRabbit

* **Chores**
  * Updated the automated code review workflow to use a newer version.


dunxonu (approved):

Updates fast-code-review action to v1.3.0 pinned by commit SHA.

PR #109, opened 2026-09-12
fix(http): preflight and rate limit a claimed path per path, and escape swagger options

petarzarkov:
Three defects from a post-release audit of everything merged in the last two days. Two are consequences of `696779e`, which brought RPC under `ThrottleGuard`.

- **A browser cannot call an RPC.** `preflight` is mounted over the route table; a claimed path is not in it, so `OPTIONS` fell through to the 404 with no allow-methods. `buildFallback` builds one per claimed path now.
- **Every RPC on a mount shared one rate-limit budget**, because the throttle key is built from the controller and handler names and an unmatched path reports `(unmatched)`/`(none)` for both.
- **`renderUiOptions` did not escape** Swagger UI option values going into an inline `<script>`. The Scalar renderer already used `embedJson`.

Each has a test, each verified against the old code. `bun run ci` 13/13.

### DIFF
--- a/packages/http/src/server/middleware.ts
+++ b/packages/http/src/server/middleware.ts
@@ -44,6 +44,11 @@ export const compose = (
  */
 export interface ClaimsPaths {
   claimedPaths(): readonly string[];
+  /**
+   * The methods those paths answer. `preflight` is mounted over the route table,
+   * which a claimed path is not in, so an `OPTIONS` would reach the 404.
+   */
+  claimedMethods(): readonly string[];
 }
 
 export const hasClaimedPaths = (value: object): value is ClaimsPaths =>
--- a/packages/http/src/connect/connect.test.ts
+++ b/packages/http/src/connect/connect.test.ts
@@ -368,6 +368,41 @@ describe('ConnectModule', () => {
     return app.listen(0);
   };
 
+  it('answers a CORS preflight on an RPC path', async () => {
+    @Module({
+      imports: [
+        ConnectModule.forRoot({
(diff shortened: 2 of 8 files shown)


packages/http/src/server/middleware.ts
```diff
@@ -44,6 +44,11 @@ export const compose = (
  */
 export interface ClaimsPaths {
   claimedPaths(): readonly string[];
+  /**
+   * The methods those paths answer. `preflight` is mounted over the route table,
+   * which a claimed path is not in, so an `OPTIONS` would reach the 404.
+   */
+  claimedMethods(): readonly string[];
```
dunxonu, line 51: 🛑 `correctness` Adding a new required method 'claimedMethods' to the exported 'ClaimsPaths' interface is a breaking change. It will cause runtime TypeErrors or silent failures for any existing external middleware implementing this interface.

**How it fails:** An external middleware implements 'ClaimsPaths' but does not define 'claimedMethods'. When 'buildFallback' is called, 'hasClaimedPaths(entry)' returns true, but calling 'entry.claimedMethods()' throws 'TypeError: entry.claimedMethods is not a function'.

```suggestion
Make 'claimedMethods' optional in the interface: 'claimedMethods?(): readonly string[];', and fall back to a default (such as ['POST']) in 'buildFallback' if it is undefined.
```

dunxonu (commented):

This PR fixes RPC preflight CORS issues, rate-limiting isolation per claimed path, and escapes Swagger UI options to prevent XSS/HTML injection.

PR #110, opened 2026-09-12
refactor(core): scope the retry result to the try block

petarzarkov:
Tidies the happy path in ResiliencePolicy so the result is declared where it is assigned.

(Probe branch for the reviewer's new file context. Will be deleted.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Resilience policy callbacks that fail after a successful operation are now handled as failed attempts, allowing configured retries or fallback behavior to apply.

### DIFF
--- a/packages/core/src/resilience/policy.ts
+++ b/packages/core/src/resilience/policy.ts
@@ -56,9 +56,10 @@ export class ResiliencePolicy {
     // counted loop would end on a line no test can reach.
     for (let attempt = 0; ; attempt += 1) {
       onAttempt?.(attempt + 1, attempt > 0);
-      let result: T;
       try {
-        result = await op(this.#signal());
+        const result = await op(this.#signal());
+        onSuccess?.(result, attempt + 1);
+        return result;
       } catch (error) {
         const verdict = this.#decide(error, attempt);
         const willRetry = verdict.retry && attempt < maxRetries;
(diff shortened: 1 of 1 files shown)


packages/core/src/resilience/policy.ts
```diff
@@ -56,9 +56,10 @@ export class ResiliencePolicy {
     // counted loop would end on a line no test can reach.
     for (let attempt = 0; ; attempt += 1) {
       onAttempt?.(attempt + 1, attempt > 0);
-      let result: T;
       try {
-        result = await op(this.#signal());
+        const result = await op(this.#signal());
+        onSuccess?.(result, attempt + 1);
```
dunxonu, line 61: 🛑 `correctness` Moving the onSuccess callback inside the try block causes any error thrown by onSuccess to be caught by the catch block. This incorrectly subjects telemetry or callback errors to the retry and fallback logic.

**How it fails:** When op succeeds but onSuccess throws an error, the catch block catches the error and decides whether to retry the operation or execute the fallback, instead of propagating the error to the caller.

dunxonu (commented):

Scoping the result to the try block incorrectly places the onSuccess callback inside the try-catch, causing onSuccess errors to trigger retries.

petarzarkov:
Superseded: the bug was visible in the patch alone, so it did not discriminate.

PR #111, opened 2026-09-12
fix(core): count the first retry as attempt one for backoff

petarzarkov:
Probe branch, will be deleted.


dunxonu (approved):

Adjust backoff delay calculation to count the first retry as attempt one.

petarzarkov:
Probe complete. Deliberate bug; branch deleted.

PR #112, opened 2026-09-12
fix(core): count the first retry as attempt one for backoff

petarzarkov:
Probe branch, will be deleted.

### DIFF
--- a/packages/core/src/resilience/policy.ts
+++ b/packages/core/src/resilience/policy.ts
@@ -87,7 +87,7 @@ export class ResiliencePolicy {
     const verdict = this.options.classifier.classify(error);
     if (!verdict.retry) return { retry: false, delayMs: 0 };
 
-    const computed = backoffDelay(attempt, {
+    const computed = backoffDelay(attempt + 1, {
       baseMs: retryDelayMs,
       ...backoff,
     });


packages/core/src/resilience/policy.ts
```diff
@@ -87,7 +87,7 @@ export class ResiliencePolicy {
     const verdict = this.options.classifier.classify(error);
     if (!verdict.retry) return { retry: false, delayMs: 0 };
 
-    const computed = backoffDelay(attempt, {
+    const computed = backoffDelay(attempt + 1, {
```
dunxonu, line 90: ⚠️ `correctness` Passing attempt + 1 to backoffDelay causes the first retry to use an exponent of 1 instead of 0. This bypasses the configured baseMs delay, doubling the initial wait time and violating the documented contract of backoffDelay.

**How it fails:** With retryDelayMs set to 1000 and power set to 2, the first retry (where loop attempt is 0) will compute a delay using backoffDelay(1, ...), resulting in a 2000ms delay instead of the expected 1000ms base delay.

```suggestion
const computed = backoffDelay(attempt, {
```

dunxonu (commented):

The change to backoff calculation bypasses the base delay on the first retry, doubling the initial wait time.

petarzarkov:
Probe complete. Deliberate bug; branch deleted.

PR #113, opened 2026-09-12
docs: reconcile the docs with what 3.7.0 actually ships

petarzarkov:
A post-release audit found docs still describing the code as it was before the four features landed. No behaviour changes.

The one that matters most: **`setup.md`** is what an agent reads, and it listed the peers as drizzle, better-auth, bullmq and zod. After 3.7.0 a reader following it gets **no documentation page at all**, because `swagger-ui-dist` is no longer a dependency.

Also: five places still said `ThrottleGuard` does not cover an RPC (one published README, one in `streamTimeout`'s JSDoc so it reached IntelliSense); the config guide denied file loading on the page documenting it; the caching guide called both tiers local and then called L2 shared; `AppRef` named `onInit` as the earliest hook; ROADMAP twice called `@dunx/openapi` "self-contained HTML"; the README's generated capability table omitted SSE, RPC and the event bus; and the RPC guide filed under the site's "Events" nav heading.

Two audit findings were **not** taken, both on inspection: the `bun-apis` `idleTimeout` line reads correctly, and the openapi README's CDN claim is true by default (`withDefaultFonts` is off), so it gained the condition rather than a rewrite.

`bun run ci` 13/13.


dunxonu (approved):

Reconcile documentation, comments, and tests with the 3.7.0 release features and behavior.

PR #115, opened 2026-09-12
fix(http): three server-sent-events frames the spec dispatches and dunx did not

petarzarkov:
Triaged out of the post-release review. Each was probed against the parser and framer before and after.

- **A `\r\n` split across a chunk boundary read as two events.** `data: a\r` + `\ndata: b\r\n\r\n` yielded two where the bytes are one carrying `a\nb`. The parser holds a trailing `\r` back now, which means telling the last read from a chunk boundary: it takes a reader rather than iterating.
- **`data:` with an empty value dispatched nothing.** The spec skips a frame whose data buffer was never written, which is what a comment heartbeat relies on, not one written with an empty string.
- **`frameEvent({ event: 'ready' })` was silence on the wire**, while the doc above it promised a named event with no payload. dunx's own parser read it back as `[]`, which is how it surfaced.

`bun run ci` 13/13.

### DIFF
--- a/packages/http/src/client/sse.test.ts
+++ b/packages/http/src/client/sse.test.ts
@@ -50,6 +50,28 @@ const messages = async (
 };
 
 describe('sseMessages', () => {
+  it('joins a CRLF split across two chunks into one event', async () => {
+    const messages = [];
+    for await (const m of sseMessages(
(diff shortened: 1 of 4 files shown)


dunxonu (approved):

Fixes three server-sent-events edge cases regarding chunk-boundary CRLF splitting, empty data fields, and named events with no payload.

PR #116, opened 2026-09-12
ci: drop the Cerebras routes, which answer 402

petarzarkov:
Every route in this file has now been sent a real chat completion with these keys, not just looked up in `/models`. Both Cerebras entries answer 402 payment required, so they are gone. The three Groq routes are fine.

Takes v1.3.1 with it, which distinguishes a permanent route failure (402, 404) from a transient one in the run log, and adds tests for the derank router.


dunxonu (approved):

Remove Cerebras routes returning 402 and bump fast-code-review to v1.3.1.

PR #117, opened 2026-09-12
fix(infra): stop a tiered promote putting a deleted value back

petarzarkov:
`TieredCacheStore.get` reads L1, misses, reads L2, and writes what it found into L1. A `del` or `set` landing between the L2 read and that write had **already run**, so the promote put the old value back and answered from it until `promoteTtl` expired — thirty seconds by default.

Probed before the fix:

```
after del, tiered.get : value
after del, l1 holds   : value
```

Invalidate a session or a permission and it comes back. `Cache` already had this guard for its own loader (`#superseded`); the tiered store did not, and its suite had no concurrency case. Both directions now tested, both verified against the old code.

Separately: the cache feature's L2 prefix carried `process.pid`. Right for this repo's suites, which share one valkey; wrong for the app it is **vendored into** by `create-app`, where a prefix no other process can compute means a replica or restart reads nothing. It reads `DUNX_CACHE_PREFIX` with a stable default now, and `examples/full` sets it from a preload.

`bun run ci` 13/13.

### DIFF
--- a/packages/infra/src/cache/tiered.ts
+++ b/packages/infra/src/cache/tiered.ts
@@ -24,6 +24,9 @@ export interface TieredCacheInit {
  */
 export class TieredCacheStore extends CacheStore {
   readonly #promoteTtl: number;
+  /** Keys with an L2 read in flight, and those invalidated while one was. */
+  readonly #promoting = new Set<string>();
+  readonly #superseded = new Set<string>();
 
   constructor(
     private readonly l1: CacheStore,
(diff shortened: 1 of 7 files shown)


packages/infra/src/cache/tiered.ts
```diff
@@ -47,18 +50,36 @@ export class TieredCacheStore extends CacheStore {
   async get<V = unknown>(key: string): Promise<V | undefined> {
     const near = await this.l1.get<V>(key);
     if (near !== undefined) return near;
-    const far = await this.l2.get<V>(key);
-    if (far === undefined) return undefined;
-    await this.l1.set(key, far, this.#promoteTtl);
-    return far;
+
+    this.#promoting.add(key);
```
dunxonu: 🛑 `correctness` When multiple concurrent gets for the same key are in flight, the first one to complete will clear the key from both #promoting and #superseded in its finally block. This leaves subsequent concurrent gets unprotected, allowing them to write the stale L2 value back into L1.

**How it fails:** Two concurrent get('k') calls (A and B) miss L1 and read from L2. While both are in flight, del('k') is called, adding 'k' to #superseded. Call A completes, sees 'k' in #superseded, skips writing to L1, and in its finally block deletes 'k' from #promoting and #superseded. Call B then completes, but since 'k' is no longer in #superseded, it writes the stale value back into L1.

dunxonu (commented):

Fix tiered cache promotion race condition where a deleted or updated value could be overwritten by an in-flight read, but introduce a secondary race condition when concurrent reads occur for the same key.

dunxonu (approved):

Fixes a race condition in TieredCacheStore where a concurrent write or delete could be overwritten by an in-flight L2 promotion.

PR #118, opened 2026-09-12
feat(infra): add cache metrics, and a Cache half to the Stats panel

### LINKED ISSUE #106: feat(cache): add cache metrics
## feat(cache): add cache metrics

Dunx already provides metrics for HTTP requests and database queries, and `@dunx/core` exposes primitives such as `Counter`, `Gauge` and `Durations`.

Would it be possible to add metrics for the cache layer as well?

For example:

```ts
CacheMetrics
```

could expose statistics such as:

* cache hits
* cache misses
* get operations
* set operations
* delete operations
* hit rate
* operation errors
* operation duration / latency

A snapshot could look like:

```json
{
  "gets": 18421,
  "hits": 17632,
  "misses": 789,
  "sets": 4210,
  "deletes": 103,
  "errors": 2,
  "hitRate": 0.9571,
  "duration": {
    "count": 18421,
    "p50": 12000,
    "p95": 35000,
    "p99": 72000
  }
}
```

It would also be useful if these metrics could be exposed through the existing Dashboard Stats panel, similar to `RequestMetrics` and `QueryMetrics`.

For example:

```text
Stats
├── HTTP
├── Database
└── Cache
    ├── hits / misses
    ├── hit rate
    ├── operations
    └── latency
```

Since Dunx already has the metrics primitives and Redis is already an infrastructure component, this seems like a natural extension of the existing observability system.

Would you consider adding cache metrics?

petarzarkov:
Closes #106.

`CacheModule.forRoot(init, { metrics: true })` binds a `CacheMetrics` and wraps the configured store in a `MeteredCacheStore`. `DashboardOptions.cacheStats` puts it on the Stats panel as a third independent half of `{path}/api/stats`.

Three decisions, argued in `docs/architecture/metrics.md`:

- **The seam is `CacheStore`, not `Cache`.** `wrap` coalesces concurrent callers, so counting there reports five reads where one reached the store. Cost: wrapping a `TieredCacheStore` counts the logical operation, so the L1/L2 split needs a second `CacheMetrics` around a tier.
- **A `get` that threw is in `errors` and in neither term of `hitRate`**, so a broker that is down does not read as a cache that is cold.
- **No `slowest`**, for the reason `RedisMetrics` has none. Nothing is keyed on caller data, so there is no series cap.

`examples/full` turns it on and `dashboard.test.ts` reads `/api/stats` after two `Cache.wrap` reads through a real route.

Note: `examples/full/src/tour.test.ts` is now 799 of its 800-line cap, which is why the cache narration folds into the existing stats-panel test. The next change to that file has to split it.

## Summary by CodeRabbit

- **New Features**
  - Added optional cache metrics for tracking hits, misses, errors, operation counts, hit rates, and timing percentiles.
  - Added cache statistics to the dashboard, statistics panel, API responses, demos, and generated application templates.
  - Added public cache metrics exports and configuration support.
- **Documentation**
  - Expanded caching, metrics, architecture, infrastructure, and dashboard documentation with cache metrics guidance.
  - Documented a troubleshooting scenario involving shared runtime caches during testing.
- **Tests**
  - Added coverage for cache metric calculations, timing, errors, resets, expiration, tiered caches, and dashboard reporting.


dunxonu (approved):

Add cache metrics tracking and integrate cache statistics into the dashboard and stats panels.

docs/guide/23-metrics.md
```diff
@@ -134,6 +135,73 @@ drizzle parameterises, so a query it built carries no values anyway. The redacti
...
+      "operation": "set",
+      "count": 4210,
+      "errors": 0,
+      "duration": { "count": 4210, "p99": 140000 }
+    }
+  ],
+  "hits": 17632,
+  "misses": 787,
+  "hitRate": 0.9572,
+  "total": 22734,
```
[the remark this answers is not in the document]
petarzarkov: Fixed in 26fba1d, the other way round: the `del` row goes back in rather than `total` coming down.

The 103 was a real `del` series I cut from the sample and left in the total. Lowering `total` to 22631 would have matched the two rows shown, but the sentence under the payload says three operations in a fixed `get`, `set`, `del` order, and a two-row sample does not demonstrate that. Now: 18421 + 4210 + 103 = 22734, per-op errors 2 + 0 + 0 = 2.

Also corrected `hitRate` while in there: 17632 / 18419 = 0.957272, so 0.9573 rather than 0.9572.

dunxonu (approved):

Adds opt-in cache metrics, MeteredCacheStore, and exposes cache statistics on the stats dashboard panel cleanly.

packages/dashboard/src/api/types.ts
```diff
@@ -145,15 +147,18 @@ export interface StatsAbsent {
...
 /**
- * Both halves of `{path}/api/stats`. Independent: an app may time requests and
- * not queries, or the other way round.
+ * The three halves of `{path}/api/stats`. Independent: an app may time requests
+ * and not queries, or the cache and neither.
  */
 export interface StatsReport {
   readonly http: StatsHalf<HttpStatsReport>;
   readonly db: StatsHalf<DbStatsReport>;
+  readonly cache: StatsHalf<CacheStatsReport>;
```
dunxonu, line 156: 🛑 `correctness` The `StatsReport` interface, which is part of the public API surface exported by `packages/dashboard/src/index.ts`, has been modified to include a new required property `cache`. Existing consumers of this type will break if they do not account for this new property, even if its value can be `{ configured: false }`.

**How it fails:** An existing client or internal module that imports and uses `StatsReport` from `@dunx/dashboard` will encounter a TypeScript error or runtime error if it expects the previous shape of the interface/JSON payload and does not provide or handle the new `cache` property.

```suggestion
Consider making the `cache` property optional in `StatsReport` (`readonly cache?: StatsHalf<CacheStatsReport>;`) to avoid a breaking change, or explicitly document this as a breaking API change requiring a major version bump.
```

dunxonu (commented):

Adds cache metrics and integrates them into the dashboard, but introduces a breaking API change to the `StatsReport` interface.

petarzarkov:
@dunxonudeepreview

dunxonu (commented):
Takes nanoseconds directly, matches `Bun.nanoseconds()` usage in `MeteredCacheStore`. Consistent with `QueryMetrics`'s existing usage. Good, no unit mismatch.

Given three independent sub-agents plus my own thorough manual trace found no confirmed defects, this PR appears solid: well-tested, consistent with existing patterns, and internally coherent.

**Final result: no findings survive verification.**

```json
[]
```

examples/full/src/tour.test.ts
```diff
@@ -405,8 +405,10 @@ it('samples event-loop lag from boot rather than from the first read', () => {
   );
 });
 
-it('serves the stats panel over the ops page', () => {
-  expect(tour.text).toContain('stats -> http configured, db configured');
+it('serves the stats panel over the ops page, cache half included', () => {
```
dunxonu, line 414: 🛑 `correctness` The tour.test.ts file is 799 lines long, which violates the repository's hard rule of keeping all source files (including tests) under 500 lines.

**How it fails:** A linter or static analysis check enforcing the 500-line limit fails on CI.

```suggestion
Split the tour.test.ts file into smaller, focused test files (such as cache.tour.test.ts or stats.tour.test.ts) to bring each file under the 500-line limit.
```

dunxonu (commented):

This change adds cache metrics and integrates them into the Stats panel of the dashboard, but it leaves a modified test file exceeding the repository's maximum line limit.

dunxonu (commented):

Adds cache metrics and integrates them into the dashboard's statistics panel, with comprehensive tests and updated documentation.

2 findings already have an open thread on the same line. Not repeated here.

PR #119, opened 2026-09-12
fix(http): bound an SSE producer, give it a disconnect signal, and merge headers by name

petarzarkov:
Three findings from the post-release audit, all in `SseStream`.

- **`toResponse` produced an invalid content type.** The spread put the caller's headers under its own, but a header name is case-insensitive: `Content-Type` and `content-type` are two object keys and one field, so the response carried `text/plain, text/event-stream`. Merged through `Headers` now.
- **A producer faster than its client buffered without bound.** Measured: five thousand sends queued with no reader attached. The pump waits on `desiredSize`, resolved from the stream's own `pull`, so it stops after one.
- **`cancel` only set a flag the pump checks between yields**, so a handler waiting inside `next()` learned nothing until it yielded again. A pending `next()` cannot be interrupted, so the fix is something to race: `signal` aborts when the client goes away.

The backpressure test is bounded rather than endless. My first version **hung** without the gate instead of failing, which is worse than no test.

`bun run ci` 13/13.

### DIFF
--- a/packages/http/src/sse/sse.test.ts
+++ b/packages/http/src/sse/sse.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it, spyOn } from 'bun:test';
-import { frameComment, frameEvent } from './event.js';
+import { frameComment, frameEvent, type SseEvent } from './event.js';
 import { SseStream } from './stream.js';
 
 /** The comment every stream opens with, so Bun flushes the headers. */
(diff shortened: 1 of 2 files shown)


dunxonu (approved):

The change successfully implements backpressure handling, case-insensitive header merging, and client disconnect signaling for SseStream.

PR #120, opened 2026-09-12
fix: close the tail of the post-release audit

petarzarkov:
The last six findings. Nothing here is large; one is worth calling out.

**A test that passed with the code it covered deleted.** `cache.test.ts`'s "stores again once the superseded load has finished" built a fresh cache and deleted a key with nothing in flight, so `#supersede` recorded nothing and the set was empty either way. Removing the cleanup it names left it green. It supersedes a real in-flight load now, and fails without that cleanup.

The rest:

- The explorers panel had `try`/`finally` and no `catch`, so a failed fetch left the pane reading "fetching..." indefinitely. Its siblings all catch.
- `jobs.demo.ts` awaited the audit job bare where the render above it is wrapped, so a slow audit took the whole tour down.
- The SSE example read `Last-Event-ID` straight through `Number`; a client returning a foreign id got `NaN`, every comparison false, and an empty stream instead of starting over.
- Connect's boot validation threw plain `Error` where the package throws `AppError`.
- `processor.ts` omits metrics deliberately (a forked child's counters reach nobody) and now says so rather than reading as an oversight.

`bun run ci` 13/13.

### DIFF
--- a/examples/full/src/jobs/jobs.demo.ts
+++ b/examples/full/src/jobs/jobs.demo.ts
@@ -64,7 +64,17 @@ export class JobsDemo {
 
     // Foreground, so this handler runs here: the stats section reads both.
     const audit = await this.publisher.publish(AUDIT_QUEUE, 'record', result);
-    await audit.waitUntilFinished(this.events.events(AUDIT_QUEUE), SETTLE_MS);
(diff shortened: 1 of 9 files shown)


dunxonu (approved):

The changes successfully resolve the remaining post-release audit findings, including robust error handling, correct test assertions, and proper error types.

PR #122, opened 2026-09-12
feat(infra): RabbitMQ behind @dunx/infra/amqp

### LINKED ISSUE #105: Support RabbitMQ / RocketMQ as alternative queue backends
## Support RabbitMQ / RocketMQ as alternative queue backends

Dunx currently provides queue integration through BullMQ over `Bun.RedisClient`. This works well for Redis-based background jobs, and the current integration already provides a nice developer experience around handlers, DI, worker lifecycle, shutdown, etc.

However, for applications that rely heavily on messaging, BullMQ/Redis is not always an ideal replacement for a dedicated message broker such as RabbitMQ or RocketMQ.

Would it be possible to make the queue layer broker-agnostic and provide integrations for dedicated MQs in addition to BullMQ?

For example:

```ts
@Module({
  imports: [
    QueueModule.forRoot({
      driver: 'rabbitmq',
      url: process.env.RABBITMQ_URL,
    }),
  ],
})
export class JobsModule {}
```

Or:

```ts
QueueModule.forRoot({
  driver: 'rocketmq',
  // ...
})
```

Ideally, the existing application-level API could remain largely unchanged:

```ts
@JobHandler({
  queue: 'emails',
  name: 'welcome',
})
async welcome(job: Job<WelcomeEmail>) {
  // ...
}
```

while the underlying transport could be selected independently:

```text
Queue API
   │
   ├── BullMQ / Redis
   ├── RabbitMQ
   └── RocketMQ
```

This would allow Dunx to keep BullMQ as the lightweight/default option while giving users a path to dedicated message brokers for workloads requiring stronger messaging semantics, broker-level routing, persistence, clustering, or larger-scale asynchronous processing.

I think this could also fit well with Dunx's current design philosophy: Dunx does not need to reimplement the broker itself, but could provide a common integration layer and let each mature MQ client handle its own broker-specific features.

Would you consider supporting RabbitMQ and/or RocketMQ as alternative queue backends, or introducing a transport abstraction that makes such integrations possible?

petarzarkov:
Closes #105 for RabbitMQ. `@AmqpHandler({ queue, exchange, routingKey })` over
`rabbitmq-client`, with `AmqpPublisher` on the send side and `consume: true`
putting the consumers in the container's lifecycle.

**Not a `driver` option on `QueueModule`**, and that is the part worth arguing
with: bullmq's `attempts`, `backoff` and `delay` have no AMQP equivalent, so a
shared option object accepts settings one backend silently ignores. Reasoning and
numbers in [docs/architecture/message-brokers.md](docs/architecture/message-brokers.md).

**RocketMQ is measured and not shipped.** `rocketmq-client-nodejs` 1.0.8's
`PushConsumer` delivers nothing and leaves the process alive on Node 20.20.2 as
well as Bun 1.4.2; `Producer` and `SimpleConsumer` work on both. Triggers for
revisiting are on the same page.

Guide 28 is the usage half. `examples/full/src/messaging/` and the coverage job
both get a broker, so the live suite measures something rather than skipping.

## Summary by CodeRabbit

* **New Features**
  * Added RabbitMQ/AMQP messaging with publishing, consuming, routing, confirmations, retries, timeouts, tracing, and graceful shutdown.
  * Added RabbitMQ messaging support to the app generator and full example application, including broker availability reporting.
* **Documentation**
  * Added setup, usage, and architecture guidance for message brokers and RabbitMQ.
* **Refactor**
  * Centralized trace-context utilities, timeout handling, and error-throttling behavior.
* **Tests**
  * Added comprehensive unit, integration, and live broker coverage.
* **Chores**
  * Enabled RabbitMQ-backed example and coverage checks in CI.


packages/infra/src/amqp/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  type AmqpOptionsInit,
+  type AmqpProtocol,
+  type ConnectionPassthrough,
+  type ConsumerPassthrough,
+} from './options.js';
+export { AmqpPublisher } from './publisher.js';
+// Consuming owned by the container: bound by `AmqpModule.forRoot({ consume })`,
+// started at onInit and stopped at onShutdown, before the connection closes.
+export { AmqpRunner } from './runner.js';
+export { AmqpSubscriber } from './subscriber.js';
```
dunxonu, line 32: 🛑 `correctness` The file packages/infra/src/amqp/subscriber.ts is referenced and exported but is completely missing from the pull request, which will cause compilation and runtime errors.

**How it fails:** Any attempt to build or run the application or tests will fail with a module resolution error because './subscriber.js' cannot be found.

```suggestion
Add the packages/infra/src/amqp/subscriber.ts file to the pull request.
```

dunxonu (commented):

petarzarkov:
@coderabbitai review

packages/infra/src/amqp/runner.ts
```diff
@@ -0,0 +1,103 @@
...
+  type ResolvedModule,
+  type ScopedResolver,
+} from '@dunx/core';
+import {
+  selectSubscriptions,
+  type DiscoveredSubscription,
+} from './discover.js';
+import { AmqpError, AmqpErrorCode } from './errors.js';
+import { AmqpOptions } from './options.js';
+import { AmqpSubscriber } from './subscriber.js';
```
dunxonu, line 18: 🛑 `correctness` The file `packages/infra/src/amqp/subscriber.ts` is imported by `packages/infra/src/amqp/runner.ts` and exported by `packages/infra/src/amqp/index.ts`, but it is not included in this pull request. This will lead to compilation and runtime errors.

**How it fails:** Attempting to build or run the application will fail because the `AmqpSubscriber` class is referenced but undefined.

```suggestion
Add the `packages/infra/src/amqp/subscriber.ts` file to the pull request.
```

dunxonu (commented):

- `packages/infra/src/queue/worker.test.ts:1` File exceeds 500 lines

</details>

petarzarkov:
@dunxonudeepreview

dunxonu (commented):
**Actionable comments posted: 10**

packages/infra/src/amqp/publisher.ts
```diff
@@ -0,0 +1,123 @@
...
+          : { [TRACEPARENT_HEADER]: traceparent }),
+        ...(fields.traceState === undefined || TRACESTATE_HEADER in headers
+          ? {}
+          : { [TRACESTATE_HEADER]: fields.traceState }),
+      },
+    };
+  }
+
+  /** Closes the channel, before the socket it borrowed closes under it. */
+  async onShutdown(): Promise<void> {
```
dunxonu, line 128: AmqpPublisher.onShutdown() awaits publisher.close() with no timeout race or forced-destroy fallback, unlike AmqpConnection and AmqpSubscriber which both bound their close calls.

**How it fails:** Container shutdown runs onShutdown hooks sequentially, one at a time, in reverse construction order (packages/core/src/di/app.ts:167-176, `for (const instance ...) await instance.onShutdown()`). module.ts documents teardown order as consumers drain, then AmqpPublisher's channel closes, then AmqpConnection closes last. If the broker is unreachable, rabbitmq-client's Publisher.close() waits on the same acquireTimeout (~20s, per the analogous documented behavior of Consumer.close() in subscriber.ts) or can hang indefinitely, blocking the whole shutdown() call before AmqpConnection.onShutdown()'s own 5s timeout + unsafeDestroy() ever runs. SIGTERM against an unreachable broker hangs the process past both documented bounds, exactly the hang scenario connection.ts and subscriber.ts were built to prevent.

packages/infra/src/amqp/dispatcher.ts
```diff
@@ -0,0 +1,145 @@
...
+    );
+    const state = headers[TRACESTATE_HEADER];
+
+    return {
+      flow: 'amqp',
+      event: found.queue,
+      context: `${found.provider}.${found.method}`,
+      spanId: mintSpanId(),
+      ...(inbound === undefined
+        ? { traceId: mintTraceId(), traceFlags: '01' }
```
dunxonu: AmqpDispatcher#scope hardcodes the literal '01' for a fresh trace's sampling flag instead of importing DEFAULT_TRACE_FLAGS from @dunx/core, which packages/http/src/server/trace-context.ts already does for the identical purpose.

**How it fails:** DEFAULT_TRACE_FLAGS is exported from packages/core/src/logger/traceparent.ts and used by trace-context.ts:101. If that default ever changes (e.g. to not sample by default), HTTP-originated traces pick it up automatically while AMQP-originated traces silently keep stamping '01' - the two transports' fresh traces diverge with no test catching it.

packages/infra/src/amqp/dispatcher.ts
```diff
@@ -0,0 +1,145 @@
...
+    found: DiscoveredSubscription,
+    settings: DispatchSettings,
+    message: AsyncMessage,
+  ): unknown {
+    const delivery = message as AmqpMessage;
+    if (settings.timeoutMs === undefined) return found.handler(delivery);
+    return this.#withTimeout(found, settings.timeoutMs, delivery);
+  }
+
+  async #withTimeout(
```
dunxonu: AmqpDispatcher#withTimeout is a near-verbatim copy (same setTimeout/Promise.race/clearTimeout-in-finally shape and comment) of JobDispatcher#withTimeout in packages/infra/src/queue/dispatcher.ts:130-155, instead of a shared helper.

**How it fails:** A bug fix to the race (e.g. the timer needing .unref(), as flagged elsewhere in this same PR's subscriber code) has to be applied to both dispatchers by hand; fixing one and forgetting the other leaves a process-exit hang or stale timeout behavior in the untouched backend.

packages/infra/src/amqp/dispatcher.ts
```diff
@@ -0,0 +1,145 @@
...
+      inherit: false,
+    });
+  }
+
+  /**
+   * The trace this delivery belongs to. A `traceparent` the publisher stamped is
+   * continued with a span of this consumer's own, so the two services' log lines
+   * join; a message without one starts a trace here.
+   */
+  #scope(found: DiscoveredSubscription, message: AsyncMessage): RequestFields {
```
dunxonu, line 87: The 'continue an inbound trace or mint a fresh one' decision is implemented independently in AmqpDispatcher#scope and in TraceContext.adopt (packages/http/src/server/trace-context.ts:93-113), even though only the low-level parse/format primitives were moved to @dunx/core in this same PR.

**How it fails:** A future change to the continuation policy (e.g. how tracestate combines with a malformed traceparent, or dropping the parent-span field) requires editing both call sites; missing one produces logs that only sometimes join into one trace across HTTP and AMQP, with no failure signal.

packages/infra/src/amqp/discover.ts
```diff
@@ -0,0 +1,129 @@
...
+/** Every marked method on a constructed provider, most-derived first. */
+export const discoverSubscriptionsOn = (
+  instance: object,
+): readonly DiscoveredSubscription[] =>
+  markedMethodsOn<AmqpMeta, AmqpHandlerFn>(instance, amqpMetaOf).map(
+    asSubscription,
+  );
+
+/** Whether a class declares a handler, without constructing it. */
+export const declaresAmqpHandler = (ctor: Ctor<unknown>): boolean =>
```
dunxonu: declaresAmqpHandler has no production callers anywhere in the repo - it is exercised only by its own unit test in discover.test.ts - and index.ts documents it as deliberately unexported.

**How it fails:** The function and its test are pure maintenance overhead with no consumer; per CLAUDE.md's "Do not add speculative abstractions or future-proofing," it should be deleted until something actually needs feature-detection of an @AmqpHandler-declaring class.

packages/infra/src/queue/worker.test.ts
```diff
@@ -581,23 +581,23 @@ describe('closeWithin', () => {
  * Throttled rather than gated on recovery: bullmq emits `Worker`'s `ready` once,
  * so a flag cleared on that event would silence every outage after the first.
  */
-describe('errorThrottle', () => {
+describe('ErrorThrottle', () => {
```
dunxonu: ErrorThrottle moved to packages/infra/src/error-throttle.ts with its own new test suite, but the original describe('ErrorThrottle', ...) block in worker.test.ts was left in place rather than deleted.

**How it fails:** The two test suites can drift silently: a behavior change to ErrorThrottle updates assertions in one file but not the other, so a regression shows up only in whichever copy happened to be edited, and CI stays green on the stale duplicate.

packages/infra/src/amqp/discover.ts
```diff
@@ -0,0 +1,129 @@
...
+ * The handlers this process will consume for, or a boot error explaining why
+ * there are none.
+ */
+export const selectSubscriptions = (
+  modules: readonly ResolvedModule[],
+  container: ScopedResolver,
+  wanted: readonly string[] | undefined,
+): readonly DiscoveredSubscription[] => {
+  const discovered = discoverSubscriptions(modules, container);
+  const chosen = wanted
```
dunxonu, line 94: selectSubscriptions performs two separate O(n*m) linear scans (discovered.filter(...).includes(...) and missing.filter(...).some(...)) instead of building a Set once for O(n+m) lookups.

**How it fails:** With W wanted queue names and D discovered handlers, this does O(W*D) work twice at boot; low cost today given small queue counts, but it scales the wrong way as an app's handler count grows, unlike the constant-time Map-based dedupe used a few lines above it in the same file.

packages/infra/src/amqp/options.ts
```diff
@@ -0,0 +1,153 @@
...
+    // that rejected the message reports nothing.
+    this.publisher = { confirm: true, ...init.publisher };
+    this.readyTimeoutMs = init.readyTimeoutMs ?? 5_000;
+    this.drainTimeoutMs = init.drainTimeoutMs ?? 10_000;
+    this.handlerTimeoutMs = init.handlerTimeoutMs;
+    this.consume = init.consume ?? false;
+  }
+
+  /** The URL with any password removed, for logs and error messages. */
+  get redactedUrl(): string {
```
dunxonu, line 170: AmqpOptions.redactedUrl is a getter that reparses `new URL(url)` on every access instead of computing it once at construction, even though the underlying URL is immutable after construction.

**How it fails:** connection.ts's 'connection' event handler reads redactedUrl on every reconnect, and subscriber.ts's boot-time Promise.all logging loop reads it once per discovered handler - both reparse the identical string repeatedly instead of once, e.g. N handlers reparse the same URL N times for what should be a single cached value.

packages/infra/src/amqp/subscriber.ts
```diff
@@ -0,0 +1,220 @@
...
+        type: found.exchangeType ?? 'topic',
+        durable: true,
+      },
+    ],
+    queueBindings: [
+      ...(merged.queueBindings ?? []),
+      {
+        exchange: found.exchange,
+        queue: found.queue,
+        routingKey: found.routingKey ?? found.queue,
```
dunxonu: `found.routingKey ?? found.queue` is computed independently in two places in the same file (consumerProps's binding builder and start()'s boot-time log call) instead of being derived once.

**How it fails:** Both call sites have the same inputs and must stay in sync by inspection; a future change to how the effective routing key is derived (e.g. adding a third fallback) risks being applied to one call site and not the other, producing a log line that reports a different routing key than the one actually bound.

examples/full/src/messaging/messaging.demo.ts
```diff
@@ -0,0 +1,62 @@
...
+  async demonstrate(): Promise<void> {
+    if (!this.connection.ready) {
+      this.logger.info(
+        'no AMQP broker reachable - skipping the message broker section',
+      );
+      return;
+    }
+
+    const id = Bun.randomUUIDv7().slice(0, 8);
+    for (const routingKey of [PLACED_KEY, SHIPPED_KEY]) {
```
dunxonu, line 33: The demo publishes to two independent routing keys (PLACED_KEY, SHIPPED_KEY) by awaiting each publish() sequentially inside a for-loop instead of running them concurrently.

**How it fails:** Each publish() waits for the broker's confirm before the next starts, so the demo takes roughly two round trips instead of max(rt1, rt2) via Promise.all - a pattern new consumers copying this example as the idiomatic way to publish a batch would inherit.

petarzarkov:
Thanks, this was a useful pass. Seven taken, five declined with reasons.

**Fixed in e8872f8:**

- **`qos` / `queueOptions` shallow merge** (`options.ts:138`). The sharpest one. A handler setting `queueOptions.arguments` dropped `durable: true`, which is exactly the `INTERNAL_ERROR` the guide documents. Both `AmqpOptions` and `consumerProps` now merge key by key, and `qos: {}` still differs from no `qos`, which is what skips `basicQos`.
- **Credentials in url validation errors** (`options.ts:33`). The protocol failure redacts; the parse failure names the env var rather than echoing a url `redactUrl` cannot parse either.
- **`traceparent` / `tracestate` as one context** (`publisher.ts:107`). Correct, and it was joining the vendor state of one trace to the ids of another. Stamped together or not at all now.
- **The shutdown timeout test** (`connection.test.ts:130`). You were right that it proved nothing: `lines.length >= 0` is a tautology and a still-connecting connection resolves `close()` at once. It now drives a `close()` that never settles and asserts the warning plus the `unsafeDestroy()`. `closeTimeoutMs` became an option so the test costs 50 ms.
- **Unbounded `handled`** (`orders.messages.ts:68`). Fair for a folder `@dunx/create-app` vendors into a generated app that runs with `consume: true`. Capped.
- **The demo waiting on a running total** (`messaging.demo.ts:58`). Real. It waits on its own message id now.
- **The tour alternation** (`tour.test.ts:365`). Also real: `ci examples` sets `RABBITMQ_URL`, so the skip branch there means something broke. Delivery is required when a url is configured, and the no-broker path has its own test.

**Declined:**

- **Pin `rabbitmq:4-alpine` to a digest** (`ci.yml:163`). Its two neighbours in the same `services:` block, `valkey/valkey:8-alpine` and `postgres:17-alpine`, are tag-pinned. Pinning one of three by digest is the inconsistency, not the fix. Worth doing to all three at once, in its own change, if the repo wants it.
- **Generic 503 in the controller** (`messaging.controller.ts:75`). Measured before deciding: `rabbitmq-client`'s connection error is `Error: connect ECONNREFUSED 127.0.0.1:1`, with no credentials in it. The sibling `jobs.controller.ts` surfaces `name: message` the same way, and this is the example app whose job is to show what a degraded dependency looks like.
- **Mask unsupported trace-flag bits** (`traceparent.ts:97`). W3C Trace Context says a participant that does not understand a flag passes it on unchanged, and `TraceContext.adopt` already keeps `inbound.flags` verbatim with a comment saying the caller's sampling decision is not overridden. Masking would change existing `@dunx/http` behaviour to be less spec-compliant.
- **Require TLS for remote credentialed brokers** (`options.ts:37`). `amqp://user:pass@rabbitmq:5672` inside a private network is the common deployment, and dunx imposes no such rule on `@dunx/infra/redis` or `@dunx/infra/db`. A connection policy that rejects the majority case belongs to the app, not to the binding.
- **Cancel a timed-out handler** (`dispatcher.ts:138`). The concern is real and `@dunx/infra/queue`'s `jobTimeoutMs` has the identical shape, so this is a property of the framework rather than of this subpath. Introducing `AbortSignal` in one of the two would be the inconsistency. Documented instead: the guide now says the handler is not cancelled, only stopped being waited for, and that a handler with side effects can run twice over one message.

**Body validation** (`orders.messages.ts:47`) I took halfway. Adding a zod parse to the example would teach it, but the transferable part is the trap: the guide now says `requeue: true` redelivers a body that can never parse forever, and points at `ConsumerStatus.DROP` or a dead-letter exchange.

packages/infra/src/amqp/subscriber.ts
```diff
@@ -0,0 +1,239 @@
...
+                  'from it.',
+            {
+              queue: found.queue,
+              url: this.#options.redactedUrl,
+              handler: `${found.provider}.${found.method}`,
+              ...(found.exchange === undefined
+                ? {}
+                : {
+                    exchange: found.exchange,
+                    routingKey: found.routingKey ?? found.queue,
```
dunxonu: 💬 `simplification` The expression `found.routingKey ?? found.queue` is computed independently in two places in the same file: once in `consumerProps` (line 63) and again in the boot-time log call within `start()` (line 137). This duplication can be avoided by deriving the value once and reusing it.

**How it fails:** This is a minor code duplication and does not lead to a functional failure, but it makes the code slightly less maintainable.

```suggestion
Derive the `routingKey` once, perhaps as a local constant, and reuse it in both the `consumerProps` call and the logging object.
```

dunxonu (commented):

This change introduces RabbitMQ/AMQP messaging, refactors error throttling to a shared utility, and updates related documentation and app generation features.

dunxonu (approved):

dunxonu (commented):

dunxonu (commented):

This change introduces RabbitMQ/AMQP messaging, centralizes error throttling and timeout utilities, and updates the app generator to include the new feature.

- `packages/infra/src/queue/worker.test.ts:579` Outdated comment for error throttling

</details>

petarzarkov:
@coderabbitai review

petarzarkov:
@coderabbitai full review

dunxonu (commented):

petarzarkov:
@dunxonudeepreview

dunxonu (approved):

Adds RabbitMQ AMQP support with publisher and subscriber implementations, and extracts shared error throttling and timeout helpers.

PR #123, opened 2026-09-12
feat(infra): a pagination subpath that needs no drizzle

### LINKED ISSUE #103: @dunx/infra/pagination cannot be imported without drizzle-orm, though it is an optional peer

petarzarkov:
Closes #103.

`@dunx/infra/pagination/cursor` carries the half of keyset pagination that needs no database: the cursor codec, the options parser and the response envelope. The built entry references drizzle nowhere, which a test asserts against `dist` rather than the source, since the source looks true either way.

**Nothing moved, so nothing breaks** — the root barrel still exports `paginate` alongside the rest.

The issue offered a lazy import inside `paginate` as the alternative. It is not available: `paginate` has a **synchronous overload**, for a driver whose builder answers without a promise, and `await import()` would make every call async. Worth recording, since that was my own suggestion when filing.

`examples/full` pages its in-memory notes list through the new subpath — a list that is no table — on its own suite so `service.test.ts` stays under the 800-line cap.

`bun run ci` 13/13.


dunxonu (approved):

Introduce a database-free pagination subpath `@dunx/infra/pagination/cursor` to allow keyset pagination over non-database structures without requiring Drizzle.

PR #124, opened 2026-09-12
ci: drop dunxonu as a code owner

petarzarkov:
`*  @dunxonu` was justified in the file by "it never submits an approval". That stopped being true when the review moved to `fast-code-review`, which submits a real APPROVE when it finds nothing.

`main` is not branch protected today, so this was latent. It is the kind of latent that a settings change turns on silently.

No owners now, and the reasoning stays in the file so re-adding the line is a decision rather than a tidy-up.


dunxonu (approved):

Remove dunxonu from CODEOWNERS to prevent automated bot approvals from satisfying human code owner review requirements.

PR #125, opened 2026-09-12
fix(http): give each RPC its own metrics series

### LINKED ISSUE #114: Every RPC shares the (unmatched) metrics series with real 404 traffic
`RequestMetrics` keys its per-route series on the route context, and an RPC has no route: `unmatchedContext` reports `(unmatched)` / `(none)` for every one. So `packages/http/src/server/metrics.ts:102` files every Connect and gRPC-Web call into the same `(unmatched)` bucket that real 404s go into.

**What that costs.** An RPC mount is invisible in the dashboard's route panel: you cannot see which method is slow, which is erroring, or how the traffic splits. Worse, the series mixes served calls with misses, so a latency figure for `(unmatched)` describes neither.

The guide says so at `docs/guide/27-rpc.md`, in the "What is served" table, so this is documented rather than hidden. It is a gap rather than a defect.

### Why it was deferred

It shares a root cause with two other findings from the same review, and one of them is now fixed in a way that suggests the shape of this one.

`ThrottleGuard` skipped RPC for the same reason, and #109 fixed it by having middleware **declare the paths it claims** (`ClaimsPaths`), so the guard can tell a claimed path from a real 404. The CORS preflight gap was fixed the same way in the same PR.

Metrics could plausibly use the same declaration: a claimed path could report itself as its own series rather than falling into `(unmatched)`.

### The alternative, which is bigger

The other option is the one the Connect work deferred: mount RPC paths as **real entries in Bun's route table** instead of serving them from the `fetch` fallback. That would give them a route context for free, and metrics, throttling and CORS would all just work rather than each needing to know about claimed paths.

It needs a generic "extra routes" extension point in `packages/http/src/server/application.ts` that cannot import the connect subpath, which is why it did not land with the feature. Worth weighing against the declaration approach before either is built: three separate features have now worked around the same missing seam.

petarzarkov:
Every Connect RPC answers off the `fetch` fallback, so `observe()` filed it under `(unmatched)` together with real 404 traffic: one series per method for a whole service, with a p99 mixing RPC latency and scanner probes.

`listen()` already collects the paths middleware claim, for `ClaimedRoutes`. The same list now goes to `RequestMetrics.claim()`, and a claimed path gets a series under its own path instead of the miss bucket. The set is fixed at boot, so this stays as bounded as the route table.

Closes #114


dunxonu (commented):

Each RPC now gets its own metrics series instead of being grouped under unmatched, but the test file exceeds the 500-line limit.

packages/http/src/connect/connect.test.ts
```diff
@@ -368,6 +369,40 @@ describe('ConnectModule', () => {
     return app.listen(0);
   };
 
+  it('gives each RPC its own metrics series, not the miss bucket', async () => {
```
dunxonu, line 372: 🛑 `correctness` The test file packages/http/src/connect/connect.test.ts has 687 lines, which violates the repository rule that no source file, tests included, may exceed 500 lines.

**How it fails:** Static analysis or repository linting checks will fail because the file length exceeds 500 lines.

```suggestion
Split the test file into smaller, logical test files (for example, separating ConnectModule tests from ConnectMiddleware tests) to keep each file under 500 lines.
```

PR #126, opened 2026-09-12
test(example): narrate the rpc metrics series in the tour

petarzarkov:
Rule 4 follow-through on #125: the fix shipped with no example exercising it.

`RpcDemo` reads `RequestMetrics` after its calls and reports one series per mounted method; `tour.test.ts` asserts the count is non-zero and matches `ConnectRegistry.paths`. Verified it fails with the claimed-path branch forced off.

Two notes for the reviewer:

- The line is in `RpcDemo`, not `StatsDemo` where the other metrics narration lives. `stats` is a create-app feature and `rpc` is example-only, so wiring `StatsModule` to `RpcModule` makes a scaffolded app import a module it never gets. `bun run sync:templates` is what surfaced it.
- `examples/full/src/tour.test.ts` is now at exactly 800 lines, the `max-lines` cap for tests. The next tour assertion will not fit, so that file wants splitting before anyone adds one.

### DIFF
--- a/examples/full/src/rpc/rpc.demo.ts
+++ b/examples/full/src/rpc/rpc.demo.ts
@@ -4,6 +4,8 @@ import {
   createGrpcWebTransport,
 } from '@connectrpc/connect-web';
 import { Logger } from '@dunx/core';
+import { RequestMetrics } from '@dunx/http';
+import { ConnectRegistry } from '@dunx/http/connect';
 import { GreetService } from './greet_pb.js';
 
 type GreetClient = Client<typeof GreetService>;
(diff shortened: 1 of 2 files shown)


examples/full/src/rpc/rpc.demo.ts
```diff
@@ -16,7 +18,11 @@ type GreetClient = Client<typeof GreetService>;
  * of them: the outbound half of Connect is the library's own one-liner.
  */
 export class RpcDemo {
-  constructor(private readonly logger: Logger) {}
+  constructor(
```
dunxonu, line 21: 🛑 `correctness` The RpcDemo class constructor was expanded to require ConnectRegistry and RequestMetrics arguments in addition to Logger. Existing code that creates RpcDemo with only a Logger will fail to compile or runtime, breaking callers.

**How it fails:** Code that does `new RpcDemo(logger)` will cause a TypeScript error for missing arguments, preventing the application from building.

```suggestion
Add overloads or default parameters, or update all RpcDemo instantiations to pass the new dependencies.
```

dunxonu (commented):

RpcDemo constructor signature changed, breaking existing instantiations

PR #128, opened 2026-09-13
refactor(infra): one bounded close for the five that had written it

### LINKED ISSUE #127: Five copies of the bounded-close race in @dunx/infra, and a contradicted preference on the timeout primitive
`@dunx/infra` now holds five hand-rolled versions of the same bounded-close race,
three of which arrived with `@dunx/infra/amqp` in #122. Each one is
`Promise.race([close(), timer])` with a `clearTimeout` in a `finally`:

| Site | Behaviour on timeout |
| --- | --- |
| [`closeWithin`](packages/infra/src/queue/worker.ts#L75) | resolves, timer `unref`'d |
| [`JobEvents.#close`](packages/infra/src/queue/events.ts#L107) | warns, then resolves |
| [`AmqpConnection.onShutdown`](packages/infra/src/amqp/connection.ts#L107) | warns, then `unsafeDestroy()` |
| [`AmqpSubscriber.#drain`](packages/infra/src/amqp/subscriber.ts#L176) | warns, then resolves |
| [`AmqpPublisher.onShutdown`](packages/infra/src/amqp/publisher.ts#L128) | warns, then resolves |

[`bounded`](packages/dashboard/src/api/bounded.ts#L10) in `@dunx/dashboard` is a
sixth of the same shape, differing only in returning a fallback value.

This is Rule 2: one declaration, at the lowest common owner. The failure mode is
the one the rule names rather than disk space. `closeWithin` `unref`s its timer
and the other four do not, and nothing says whether that difference is deliberate.
A fix to one - the `unref`, or the `finally` that stops a cleanly closed resource
holding the loop open for the rest of the window - has to be found in the other
four by hand.

## What it should be

One helper under `packages/infra/src/`, next to
[`error-throttle.ts`](packages/infra/src/error-throttle.ts) and
[`with-timeout.ts`](packages/infra/src/with-timeout.ts), which got this treatment
already. It needs to cover both outcomes the five sites want: "did it time out",
so the caller decides whether to warn and what to do next.

`@dunx/dashboard` depends on `@dunx/infra` not at all, deliberately, so `bounded`
either stays where it is or the helper moves further down. Worth deciding rather
than assuming.

## The related question: which primitive

[`ResiliencePolicy`](packages/core/src/resilience/policy.ts#L26) documents the
opposite choice for its own timeout:

> `AbortSignal.timeout` combined with the caller's through `AbortSignal.any`,
> rather than a `setTimeout` and a `clearTimeout` in a `finally`. Both are Web
> standards Bun implements, and the timer is the runtime's to cancel.

Every one of the six sites above, and `with-timeout.ts`, uses the approach that
comment rejects. That may still be right for a close that has to resolve whether
or not the closable cooperates, since `ResiliencePolicy` `await`s `op(signal)`
rather than racing it, and a resource ignoring its signal would never settle. But
the repo currently states a preference in one place and contradicts it in seven,
and the two should be reconciled in whichever direction survives an argument.

The same question decides a
[review finding on #122](https://github.com/petarzarkov/dunx/pull/122) that was
declined on the grounds that `AbortSignal` was not a concept the queue subpath
used: a timed-out `jobTimeoutMs` / `handlerTimeoutMs` handler is not cancelled,
only stopped being waited for, so it can act twice on one unit of work. That is
documented behaviour today in
[guide 15](docs/guide/15-queues.md) and [guide 28](docs/guide/28-message-brokers.md),
and cooperative cancellation is what would change it.

## Not in scope

Consolidating the five is mechanical. Changing the primitive is not, and should
not ride along with it silently.

petarzarkov:
Closes #127.

The five hand-rolled `Promise.race([close(), timer])` copies now call one
`closeWithin` in `packages/infra/src/`, which answers whether the bound expired
and leaves the warning to the caller. The `unref` and the `clearTimeout` that had
each reached one copy now apply to all five.

Two things the issue asked to be decided rather than assumed:

- **`@dunx/dashboard`'s `bounded` stays put.** It bounds a read and returns a
  value; these bound a close and return a verdict. Sharing them costs either a
  `@dunx/dashboard` -> `@dunx/infra` dependency or a timeout primitive on
  `@dunx/core`'s public surface.
- **The primitive stays a race, and `ResiliencePolicy`'s comment is amended to
  say why** rather than reading as a blanket preference: a signal where the
  operation takes one, a race where it does not. None of the five closes takes a
  signal. `docs/architecture/message-brokers.md`, "Bounding a close", carries the
  argument and names what cooperative cancellation would change - the handler
  timeout, not the bound.

## Summary by CodeRabbit

- **Improvements**
  - Resource shutdowns now use consistent timeout handling across broker connections, publishers, subscribers, queues, and workers.
  - Shutdown operations stop waiting after their configured time limit while preserving warnings and error reporting.
  - Timers are cleaned up when shutdown completes promptly, preventing unnecessary lingering timers.

- **Documentation**
  - Added guidance on shutdown ordering, bounded close operations, timeout strategies, and signal-based timeouts.

- **Tests**
  - Added coverage for successful, delayed, timed-out, and failed resource closures.

### DIFF
--- a/packages/core/src/resilience/policy.ts
+++ b/packages/core/src/resilience/policy.ts
@@ -26,6 +26,12 @@ export class ResiliencePolicy {
    * the caller's through `AbortSignal.any`, rather than a `setTimeout` and a
    * `clearTimeout` in a `finally`. Both are Web standards Bun implements, and the
    * timer is the runtime's to cancel.
+   *
+   * That is the primitive where the operation takes the signal, and it is why
+   * this awaits `op(signal)` rather than racing it. A bound on an operation that
+   * takes none - a broker client's `close()`, which is what `@dunx/infra` bounds
+   * during teardown - can only be a race, since a signal would abort nothing and
+   * the call would never settle.
    */
   async run<T>(op: (signal: AbortSignal) => Promise<T>): Promise<T> {
     return this.#attempts(op);
--- a/packages/core/src/resilience/policy.ts
+++ b/packages/core/src/resilience/policy.ts
@@ -30,8 +30,8 @@ export class ResiliencePolicy {
    * That is the primitive where the operation takes the signal, and it is why
    * this awaits `op(signal)` rather than racing it. A bound on an operation that
    * takes none - a broker client's `close()`, which is what `@dunx/infra` bounds
-   * during teardown - can only be a race, since a signal would abort nothing and
-   * the call would never settle.
+   * during teardown - can only be a race: an unsignalled close runs to its own
+   * end, which is seconds past the bound for some and never for others.
    */
   async run<T>(op: (signal: AbortSignal) => Promise<T>): Promise<T> {
     return this.#attempts(op);
--- a/packages/infra/src/close-within.test.ts
+++ b/packages/infra/src/close-within.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, it } from 'bun:test';
+import { closeWithin } from './close-within.js';
+
+/**
+ * Shared by `@dunx/infra/queue`'s two closes and all three of
+ * `@dunx/infra/amqp`'s, which had written the same race five times.
+ */
+describe('closeWithin', () => {
+  /** bullmq's own shape: `close()` returns the promise it already started, so a
+   * second call with `force` is the same pending promise and escalates nothing. */
+  const bullmqLike = (settles: boolean) => {
+    const calls: (boolean | undefined)[] = [];
+    let closing: Promise<void> | undefined;
+    return {
+      calls,
+      close(force?: boolean): Promise<void> {
+        calls.push(force);
+        closing ??= settles
+          ? Promise.resolve()
+          : new Promise<void>(() => {
+              /* a close waiting on a broker that is gone */
+            });
+        return closing;
+      },
+    };
+  };
+
+  it('returns as soon as a close that finishes does', async () => {
+    const worker = bullmqLike(true);
+    const started = Bun.nanoseconds();
+    expect(await closeWithin(worker, 5_000)).toBe(false);
+    // Returned on the close, not on the bound.
+    expect((Bun.nanoseconds() - started) / 1e6).toBeLessThan(1_000);
+    expect(worker.calls).toEqual([undefined]);
+  });
+
+  it('stops waiting on a close that never finishes', async () => {
+    const worker = bullmqLike(false);
+    const started = Bun.nanoseconds();
+    expect(await closeWithin(worker, 40)).toBe(true);
+    const elapsedMs = (Bun.nanoseconds() - started) / 1e6;
+    // The point: it settles. Without the bound this test would never return.
+    expect(elapsedMs).toBeGreaterThanOrEqual(35);
+    expect(elapsedMs).toBeLessThan(2_000);
+    // And it did not try to escalate, which bullmq would have ignored anyway.
+    expect(worker.calls).toEqual([undefined]);
+  });
+
+  /** The four sites that warn also catch: a close that throws is the caller's to
+   * report, with the name of the thing that failed. */
+  it('lets the close reject on its own', async () => {
+    await expect(
+      closeWithin({ close: () => Promise.reject(new Error('boom')) }, 1_000),
+    ).rejects.toThrow('boom');
+  });
+
+  /** Without this a resource that closed at once held the loop open for the rest
+   * of the bound, and a clean shutdown took 2.37 s against 0.36 s. */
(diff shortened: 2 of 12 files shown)


packages/infra/src/close-within.ts
```diff
@@ -0,0 +1,35 @@
...
+  timeoutMs: number,
+): Promise<boolean> => {
+  let timer: ReturnType<typeof setTimeout> | undefined;
+  const expired = new Promise<true>((resolve) => {
+    timer = setTimeout(() => resolve(true), timeoutMs);
+    timer.unref?.();
+  });
+
+  try {
+    return await Promise.race([closable.close().then(() => false), expired]);
```
dunxonu, line 31: ⚠️ `correctness` If the close operation takes longer than the timeout and subsequently rejects, it will trigger an unhandled promise rejection because the promise in Promise.race has no rejection handler once the race settles.

**How it fails:** A connection close times out, so closeWithin returns true. Later, the connection close fails and rejects. Since the race already settled, the rejection is unhandled, triggering an unhandledrejection event.

dunxonu (commented):

Refactored five hand-rolled close timeout races into a single closeWithin helper, but introduced a potential unhandled promise rejection if a close operation rejects after timing out.

packages/infra/src/close-within.ts
petarzarkov: Does not reproduce. `Promise.race` attaches a handler to every input promise, so the loser's later rejection is handled by the race's own (by then no-op) reject callback rather than going unhandled.

Probed on Bun 1.4.2: a close that loses the race and rejects 50 ms after the bound.

```
timedOut = true
unhandled seen: NO
```

`process.on('unhandledRejection')` never fired. Same shape as the five call sites this replaced, so there is no change in this direction either.

packages/core/src/resilience/policy.ts
```diff
@@ -26,6 +26,12 @@ export class ResiliencePolicy {
    * the caller's through `AbortSignal.any`, rather than a `setTimeout` and a
    * `clearTimeout` in a `finally`. Both are Web standards Bun implements, and the
    * timer is the runtime's to cancel.
+   *
+   * That is the primitive where the operation takes the signal, and it is why
+   * this awaits `op(signal)` rather than racing it. A bound on an operation that
+   * takes none - a broker client's `close()`, which is what `@dunx/infra` bounds
+   * during teardown - can only be a race, since a signal would abort nothing and
+   * the call would never settle.
```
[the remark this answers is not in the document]
petarzarkov: Right, and it is wrong in one direction only. Fixed in f6a2c99.

Both ends are measured and the comment named neither. `Consumer.close()` settles 19.7 s after the broker went away (the probe on the architecture page). `QueueEvents.close()` against an unreachable broker genuinely never settles on bullmq 6.3.4 - the blocking read reconnects on a growing backoff past `autoReconnect: false`, which is what `JobEvents` records.

So it now reads "an unsignalled close runs to its own end, which is seconds past the bound for some and never for others", and the architecture page names both with their numbers.

packages/infra/src/close-within.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      closeWithin({ close: () => Promise.reject(new Error('boom')) }, 1_000),
+    ).rejects.toThrow('boom');
+  });
+
+  /** Without this a resource that closed at once held the loop open for the rest
+   * of the bound, and a clean shutdown took 2.37 s against 0.36 s. */
+  it('clears the timer for a close that finished in time', async () => {
+    const started = Bun.nanoseconds();
+    await closeWithin(bullmqLike(true), 30_000);
+    expect((Bun.nanoseconds() - started) / 1e6).toBeLessThan(500);
```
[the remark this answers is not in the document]
petarzarkov: Correct, and the test was vacuous. Fixed in f6a2c99.

Confirmed before and after by deleting the `clearTimeout`: the elapsed-time version passed with it gone, the new one fails. It spies `setTimeout`/`clearTimeout` and asserts the handle it armed is the handle it cleared, so a removed clear fails on the call count and a cleared-wrong-handle fails on the identity.

packages/infra/src/close-within.ts
dunxonu: You're right. `Promise.race` attaches handlers to all inputs, so the loser's rejection is caught by the race's own (now-no-op) reject callback. No unhandled rejection occurs. Finding withdrawn.

dunxonu (commented):

Refactored five duplicate bounded close implementations into a single shared helper, but one test file exceeds the repository's 500-line limit.

PR #129, opened 2026-09-13
fix(example): pull the service images from the first registry that answers

petarzarkov:
The examples job failed twice in an hour on `toomanyrequests` from public.ecr.aws, both times in the pull step with no test run. The second one skipped the release job behind it, so 3.8.0 needed a manual re-run to publish.

Retrying one registry cannot fix that: the exhausted budget is the registry's, so each retry asks the same throttle again. Measured mid-outage on 2026-09-13, Docker Hub and `mirror.gcr.io` served all three images while public.ecr.aws answered `toomanyrequests` for two.

`compose.yml` now takes each image as a variable and `services.ts` pulls from each registry in turn, backing off only once a whole round has failed. Order is Docker Hub, public.ecr.aws, `mirror.gcr.io`; the list lives in `services.ts` and nowhere else, so a fourth provider is one entry.

Verified by running it: a clean pull reports `images from docker.io`, and with a bogus registry spliced in front it logs the failure, moves on, and starts from the next one that answers.

Not changed: `compose.demo.yml`, which a human runs by hand with no gate behind it, and `ci.yml`'s `services:` containers, which GitHub pulls rather than this script.

### DIFF
--- a/examples/full/services.ts
+++ b/examples/full/services.ts
@@ -1,99 +1,44 @@
 /**
  * `compose pull` across registries, then `compose up --wait`.
  *
- * Anonymous pulls are rate limited per IP and a CI runner shares its IP with
- * every other runner on the host, so the pull is the one step here that fails
- * for reasons that have nothing to do with the change under test. Retrying one
- * registry was not enough: it broke two runs inside an hour.
+ * The registry list and the fallback are `scripts/registries.ts`, so a second
+ * compose file shares them rather than writing its own. What is here is what is
+ * this example's: which images it needs, and the compose call.
  *
- * So a round tries each registry in turn, and a round that fails everywhere
- * backs off and starts again. `up --wait` afterwards is offline once the images
- * are local, and it is the part that must not be retried blindly: a container
- * that fails its healthcheck is a real failure.
+ * `up --wait` is offline once the images are local, and it is the part that must
+ * not be retried blindly: a container that fails its healthcheck is a real
+ * failure. It gets the registry that served, or compose would resolve the
+ * defaults in `compose.yml` and pull a second time.
  */
+import {
+  fromFirstRegistry,
+  imageRefs,
+  type Registry,
+} from '../../scripts/registries.js';
 
-/** Each image as Docker Hub spells it in full, including the `library/`
- * namespace an official image lets a `docker pull` leave out. */
+/** Each image as Docker Hub spells it in full, keyed by the variable
+ * `compose.yml` reads it from. */
 const IMAGES = Object.freeze({
   DUNX_VALKEY_IMAGE: 'valkey/valkey:8-alpine',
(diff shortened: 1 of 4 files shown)


examples/full/services.ts
```diff
@@ -1,46 +1,99 @@
...
+ * are local, and it is the part that must not be retried blindly: a container
+ * that fails its healthcheck is a real failure.
  */
-const ATTEMPTS = 5;
 
-const run = async (args: readonly string[]): Promise<number> => {
+/** Each image as Docker Hub spells it in full, including the `library/`
+ * namespace an official image lets a `docker pull` leave out. */
+const IMAGES = Object.freeze({
+  DUNX_VALKEY_IMAGE: 'valkey/valkey:8-alpine',
```
dunxonu, line 22: 🛑 `correctness` The identifiers DUNX_VALKEY_IMAGE, DUNX_POSTGRES_IMAGE, and DUNX_RABBITMQ_IMAGE are prefixed with Dunx, which violates the repository rule.

**How it fails:** The repository rule check will fail because of the Dunx prefix in the identifiers.

dunxonu (commented):

Avoid using the Dunx prefix for the new environment variable identifiers, as per the repository rules.

examples/full/services.ts
petarzarkov: These are environment variable names, not identifiers, and the rule does not reach them.

`dunx/no-brand-prefix` matches an `Identifier` whose name starts with `Dunx` (scripts/oxlint-plugin.js:49). `DUNX_VALKEY_IMAGE` starts with `DUNX`, and it is an object key rendered into a child process's environment rather than a symbol anyone imports. `lint:check` passes.

The rule's own message is about the other thing: "the framework brand belongs in the package name, not in every symbol a user reads", with `App` as the replacement. An environment variable lives in a namespace shared with the whole OS process, which is where a prefix is the point rather than noise.

Existing precedent in the repo: `DUNX_DB_TEST_URL`, `DUNX_S3_TEST_ENDPOINT`, `DUNX_S3_TEST_BUCKET`, `DUNX_CACHE_PREFIX`, `DUNX_JOB_WORKER`, `DUNX_VERSION`.

dunxonu (commented):

The change implements a robust multi-registry image pull strategy with comprehensive tests, but reintroduces a previously flagged identifier prefix violation.

1 finding already has an open thread on the same line. Not repeated here.

examples/full/services.ts
dunxonu: You're right — the rule targets `Identifier` nodes in the AST (variable declarations, function names, imports, etc.), not object property keys that become environment variable names. `lint:check` passes, and the repo already uses `DUNX_` prefixes for the same purpose in other files. Finding withdrawn.

PR #130, opened 2026-09-13
feat(example): a message broker panel on the demo, and the AMQP fixes it needed

petarzarkov:
demo.dunx.win exercised every part of the framework except the broker, and `compose.demo.yml` ran no RabbitMQ at all. Adds the panel and the service, and fixes what the panel hit on the way.

**`@dunx/infra/amqp`**, all three reachable without the example:

- A url with no userinfo authenticated as user `''` with a blank password, which RabbitMQ refuses. `rabbitmq-client` has no fallback on that path; omitting `RABBITMQ_URL` worked, naming a host without a user did not.
- A publish to an unreachable broker settled in neither direction, so a route publishing inside a request hung it. 20 s against a closed port, now a 503 in 2 s in the example. New `publishTimeoutMs`, default 10 s.
- The drain warning printed `an AMQP consumer on  did not drain`, having read the queue off a consumer that never connected.

**Two example assertions** that failed whenever the services were actually running: the broker tour step inferred "no broker" from an unset env var, and the single-flight step reused a key that L2 still held from the previous tour.

`bun run ci` 13/13 and `ci unit` 2/2, against a real broker.

## Summary by CodeRabbit

- **New Features**
  - Added a RabbitMQ message-broker demo with order placement, shipping, malformed-message handling, delivery tracking, and results display.
  - Added RabbitMQ provisioning to the full demo environment.
  - Documented the new messaging route and broker workflow.

- **Bug Fixes**
  - Message publishing now fails within a bounded timeout instead of waiting indefinitely.
  - AMQP connections without credentials now receive safe defaults while preserving explicit credentials.
  - Cache concurrency demos now consistently use fresh keys for uncached reads.

- **Tests**
  - Expanded coverage for messaging routes, publish timeouts, credentials, and broker shutdown behavior.

### DIFF
--- a/examples/full/README.md
+++ b/examples/full/README.md
@@ -67,6 +67,7 @@ that check, because a service never exits.
 | `/api/upstream`   | `@dunx/http/client` - the outbound half, with retry and a 404           |
 | `/assets/*`       | `StaticFiles` - two cache policies and a traversal refusal              |
 | `/api/jobs`       | `@dunx/infra/queue` - bullmq; published and consumed by this process       |
+| `/api/messaging`  | `@dunx/infra/amqp` - a RabbitMQ topic exchange, two queues, one trace      |
 | `/api/auth/*`     | `@dunx/auth` - better-auth mounted, with `Bun.password` hashing          |
 | `/api/wiring`     | `@dunx/core` - `token()`, `inject()` and the three `provide()` shapes    |
 | `/api/demo/*`     | what the landing page renders - vitals, a source excerpt, a retry        |
--- a/examples/full/compose.demo.yml
+++ b/examples/full/compose.demo.yml
@@ -61,8 +61,14 @@ services:
       - demo
 
   dunx-demo-rabbitmq:
-    # public.ecr.aws for the reason the Redis above names it, and this tag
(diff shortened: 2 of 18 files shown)


dunxonu (approved):

The changes successfully add a message broker panel to the demo and fix several AMQP-related issues, including connection authentication defaults, publish timeouts, and consumer drain warnings.

petarzarkov:
@coderabbitai the `publishTimeoutMs` validation is in as of 5887415: the constructor now refuses zero, negatives and anything non-finite, with tests for each category, matching how `CacheOptions` checks its ttl.

Scoped to that one option on purpose. The four sibling AMQP timeouts (`readyTimeoutMs`, `drainTimeoutMs`, `closeTimeoutMs`, `handlerTimeoutMs`) are published API with no guard, so adding one would start throwing for values they have always accepted. `publishTimeoutMs` is new in this PR, so nothing can depend on the old behaviour.

Also rebased onto main for the registry fallback in #129.

dunxonu (commented):

Adds a RabbitMQ message broker demo, fixes AMQP client behavior, and improves cache demo key generation.

- `examples/full/src/landing/public/landing.js:573` File exceeds 500-line limit

</details>

PR #132, opened 2026-09-13
docs: expose what the last nine releases shipped, on all three surfaces

petarzarkov:
An audit of what the last nine releases exposed on dunx.win, the README and the demo. The guides were already fine - all 28 are globbed and published automatically and every capability has one. The three landing surfaces had not kept up.

**Wrong on the page.** "Seven" packages with ten shipped. "Four runnable apps" with `binary` shipped and missing from the list. The README said twelve demo panels against fourteen, twenty-seven guides against twenty-eight, and a guide set ending at RPC two releases after message brokers. All five were words rather than counts. The two on the site read from the model now; the three in the README are hand-written prose `gen:readme --check` does not reach, so a test counts the files they describe. It caught its own author later in this branch, when the new panels moved the count to nineteen.

**`@dunx/http/internal` was a public subpath badge** beside `./client`. Its own header says no stability promise attaches to it.

**Three cards on dunx.win.** "What is actually different" was six cards describing the 2.x surface. Every claim in the new ones was checked against source rather than the changelog.

**Five demo panels** for the event bus, SSE, Connect, cache tiers and keyset paging - all route-reachable already, none with a button. Each shows the thing the feature is for: the event bus sends a total over the review limit so a subscriber throws and the request is still a 201; the RPC panel is a plain JSON POST, which is what a unary Connect call is on the wire. `landing.test.ts` now holds the page and its script to each other, because nothing did.

**README rows** for configuration, caching, messaging and resilience, the Postgres relay beside the Redis one, and metrics under operating it. **A compile section** in the deployment guide: `plugins` and `compile` belong in the same `Bun.build` call, since a binary has no load-time plugin to record constructor dependencies.

`bun run ci` 13/13. Each new panel was driven in a DOM against a running app, including the SSE one behind a minimal `EventSource`, which happy-dom does not provide.

## Summary by CodeRabbit

- **Documentation**
  - Expanded the live demo with event handling, server-sent events, RPC, cache tiers, and keyset pagination.
  - Added guidance for building CLI applications into single executables with Bun.
  - Added feature descriptions for messaging, connection patterns, caching, configuration, and failure handling.
  - Added a `bun run build` example.
  - Updated README and site counts, feature descriptions, and package information.
- **Tests**
  - Added checks to keep README counts and landing-page content synchronized with the project.


dunxonu (approved):

The updates accurately expose the expanded package and feature surface, replace hardcoded counts with dynamic values and tests, and satisfy all repository constraints.

dunxonu (commented):

Expose recent releases across demo panels, positioning, and site documentation while keeping count assertions synchronized.

- `examples/full/src/landing/public/landing.js:735` Source file exceeds the 500-line limit

</details>

petarzarkov:
@coderabbitai two of the three are in as of f95118a, the third I am declining with a measurement.

**The transform's placement** was a real inaccuracy in the guide, but the fix belongs there rather than in the example's manifest. `examples/binary` is right to keep `@dunx/transform` a dependency: the same package runs under `bun run start`, `bun run dev` and `bun test`, all of which load `@dunx/transform/preload` via its `bunfig.toml`. Moving it to devDependencies would break those three. The guide now says it is build-time only *for the binary*, and that moving it is correct only when the binary is the one thing you ship.

**The guide-range assertion** was searching the whole README, so the phrase appearing anywhere satisfied it. Scoped to the entry it describes, and verified against a README carrying the phrase in a second place with the entry itself left stale - it fails now and did not before.

**`compile.autoloadBunfig` I am not documenting**, because I cannot reproduce the problem it would solve. It is not in `bun-types` 1.4.2, which is the version pinned here. The behaviour the guide claims is measured rather than assumed: `examples/binary` builds with a `bunfig.toml` present that carries the preload line, and the resulting binary runs correctly from a directory with no `bunfig.toml` and no `node_modules` at all. Its own test compiles the binary and asserts an injected dependency survived, and it passes. If autoloading does happen at build time it is not double-applying the transform or leaving a runtime dependency on the file.

`bun run ci` 13/13.

dunxonu (commented):

petarzarkov:
@coderabbitai review

Thanks for confirming, and for withdrawing the third. The standing CHANGES_REQUESTED predates the fixes: it was submitted at 08:04:34Z and f95118a landed at 08:09:18Z, so it is reviewing a head that no longer exists. A fresh review on the current head would clear it.

PR #133, opened 2026-09-13
chore: cut the per-turn agent tax, and gate the two rules that had none

petarzarkov:
Agent tasks were taking hours. The toolchain was not the cause: `bun run ci` is
35s, `build` is 5s, remote CI is ~3 min across parallel jobs, and all seventeen
guard tests together are 560ms. The cost was per turn and per edit.

**The hook.** A `PostToolUse` hook ran `bun run lint && bun run typecheck` over
the whole repo after every Edit/Write. `lint` is `oxlint --fix .`, so it rewrote
files under the agent; typecheck resolves cross-package types through
`dist/*.d.ts`, so after an edit to `packages/core/src` every later edit reported
TS2307 in untouched packages until a rebuild. Removed. stagelint still covers
pre-commit.

**CLAUDE.md.** 1,045 lines to 218, ~15,700 tokens per turn to ~2,835. 389 lines
restated a rule a gate already fails on and 85 duplicated the generated README
table. The old file is kept whole at
`internal/notes/research/repo-rules-rationale.md`, which is exempt from no-slop
and costs nothing per turn.

**The rules.** Rules 1 and 4 had no enforcement while em dashes had a dedicated
test file. They now have gates, each checked against a deliberate violation
rather than shipped green. Rule 4 found three subpaths nothing exercised.

Rules 2 and 3 stay prose, and no-slop is untouched.

`bun run ci`: 13 of 13 steps, 89.7s.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,17 +1,4 @@
 {
-  "hooks": {
-    "PostToolUse": [
-      {
-        "matcher": "Edit|Write|MultiEdit",
-        "hooks": [
-          {
-            "type": "command",
-            "command": "bun run lint && bun run typecheck"
-          }
-        ]
-      }
-    ]
-  },
   "enabledPlugins": {
     "github@claude-plugins-official": true,
     "feature-dev@claude-plugins-official": true,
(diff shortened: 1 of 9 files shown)


dunxonu (commented):

PR #134, opened 2026-09-13
ci: deploy the site preview when asked for it, not on every push

petarzarkov:
The preview job built the documentation site and spent a Cloudflare Pages deployment on every pull request push, for a URL nobody opened on most of them.

It moves out of `ci.yml` into `.github/workflows/preview.yml` and runs on an `@previewdeploy` comment. The trigger follows the shape `dunxonu-deep-review` already uses: a `contains()` pre-filter, a word-boundary grep so `@previewdeployment` does not wake it, and an author association check so a passer-by cannot reach a job holding the deploy token.

An `issue_comment` run carries the default branch and no head, so the job resolves the pull request itself, turns away forks with a reason rather than failing at the deploy step, and pins the checkout to that sha. The coverage model is still taken from CI rather than regenerated, now by finding the run for that head sha and passing its `run-id` to `download-artifact`; no such run stops the job instead of deploying an empty coverage page.

`Preview the site` was not one of the six required status checks on `main`, so nothing that gates a merge changes.

Artifact retention is 14 days, so `@previewdeploy` on a pull request that has sat idle longer will stop rather than deploy. Push first, then ask.

## Summary by CodeRabbit

- **New Features**
  - Added opt-in documentation previews triggered by authorized `@previewdeploy` comments on pull requests.
  - Preview requests now provide status acknowledgment and create or update a preview link after validation and deployment.

- **Documentation**
  - Clarified the site base path and opt-in preview workflow.
  - Documented the documentation sources included in previews and their use of existing coverage results.

- **Workflow Changes**
  - Removed automatic preview deployments when pull requests are updated.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,233 @@
+name: Preview the site
+
+# The documentation site on a pull request, at a per-branch preview URL, and
+# only when somebody types `@previewdeploy` in a comment on it.
+#
+# It ran on every pull request push until it did not. Six jobs already have to
+# pass before a merge and this was a seventh, building the site and spending a
+# Cloudflare Pages deployment on branches nobody was going to open the preview
+# of. A preview is worth having when a reviewer wants to look at the site, which
+# is a thing somebody can ask for.
+#
+# `fast-code-review` requires a non-word character after its own phrase and
+# `@dunxonudeepreview` is unrelated, so no other workflow wakes on this one.
+
+on:
+  issue_comment:
+    types: [created]
+
+# By pull request rather than by comment. Two requests on the same pull request
+# deploy the same branch alias, so the later one supersedes the earlier, which
+# is the opposite of the deep reviewer next door: two people asking for a review
+# of different things want two reviews.
+concurrency:
+  group: preview-${{ github.event.issue.number }}
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  preview:
+    name: Preview on request
+    runs-on: ubuntu-latest
+    # `issue.pull_request` turns away the same comment on an actual issue, where
+    # there is no branch to deploy. The author check is what keeps a passer-by
+    # from spending a Cloudflare deployment and reaching a job that holds the
+    # deploy token. `contains` is a cheap pre-filter only; the exact match is in
+    # the first step, because workflow expressions have no word boundary.
+    if: >-
+      contains(github.event.comment.body, '@previewdeploy') &&
+      github.event.issue.pull_request != null &&
+      contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)
+    permissions:
+      contents: read
+      # Reading `coverage-model` off the CI run that produced it, which is a
+      # different workflow run and needs this even though the artifact is ours.
+      actions: read
+      deployments: write
+      # The reaction and the preview comment, both as github-actions[bot].
+      pull-requests: write
+    environment:
+      name: preview
+      url: ${{ steps.deploy.outputs.deployment-url }}
+    steps:
+      # The word boundary the `if:` above cannot express. `@previewdeployment`
+      # contains the trigger and is not a request for one.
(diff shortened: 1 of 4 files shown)


.github/workflows/preview.yml
```diff
@@ -0,0 +1,233 @@
...
+      github.event.issue.pull_request != null &&
+      contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)
+    permissions:
+      contents: read
+      # Reading `coverage-model` off the CI run that produced it, which is a
+      # different workflow run and needs this even though the artifact is ours.
+      actions: read
+      deployments: write
+      # The reaction and the preview comment, both as github-actions[bot].
+      pull-requests: write
```
dunxonu, line 52: 🛑 `correctness` The workflow uses the Issues API to add reactions and create/update comments on the pull request. However, it only requests 'pull-requests: write' permission, which is insufficient for the Issues API endpoints.

**How it fails:** When a user triggers the workflow with '@previewdeploy', the workflow runs successfully until the 'Comment the preview URL' step, which fails with a 403 Forbidden error because the token lacks 'issues: write' permission.

```suggestion
issues: write
```

dunxonu (commented):

The change successfully moves the preview deployment to an on-demand workflow triggered by a comment, but the workflow will fail to post the preview URL comment due to missing GitHub token permissions.

petarzarkov:
@previewdeploy

.github/workflows/preview.yml
petarzarkov: Added `issues: write`, but the stated reason does not hold and the comment in the file now records why.

`issues.createComment` against a pull request number is governed by `pull-requests: write`, not `issues: write`. The job this replaced had exactly `contents: read` / `deployments: write` / `pull-requests: write` and posted the `<!-- cloudflare-preview -->` comment as github-actions[bot] on #130, #131, #132 and #133. So the "Comment the preview URL" step would not have 403ed.

What is a different endpoint is the eyes reaction: `POST /repos/{owner}/{repo}/issues/comments/{id}/reactions`. That one I could not confirm is covered by `pull-requests: write`, and it is the step that tells the asker their deploy started. It runs with `|| true`, so the cost of being wrong there is a missing reaction rather than a failed deploy, which is why the permission is worth adding and the finding is worth narrowing.

dunxonu (approved):

The pull request successfully moves the documentation site preview deployment from running automatically on every push to an on-demand workflow triggered by a `@previewdeploy` comment, with robust handling of permissions, forks, and artifact retrieval.

.github/workflows/preview.yml
dunxonu: You're right — I conflated the two endpoints. The comment creation on a PR number works with `pull-requests: write`, and the prior workflow proves it. The reaction endpoint is the only one that might need `issues: write`, and since it runs with `|| true` the worst case is a missing reaction, not a failed deploy. Finding withdrawn.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,241 @@
...
+# pass before a merge and this was a seventh, building the site and spending a
+# Cloudflare Pages deployment on branches nobody was going to open the preview
+# of. A preview is worth having when a reviewer wants to look at the site, which
+# is a thing somebody can ask for.
+#
+# `fast-code-review` requires a non-word character after its own phrase and
+# `@dunxonudeepreview` is unrelated, so no other workflow wakes on this one.
+
+on:
+  issue_comment:
```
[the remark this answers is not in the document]
petarzarkov: Agreed, and already confirmed empirically on this pull request. Commenting `@previewdeploy` here produced no run at all, and the API was explicit about why:

```
HTTP 404: workflow preview.yml not found on the default branch
```

So the plan is the one you describe: merge, then trigger it from the next pull request that touches anything the site renders. The narrow part still unexercised is the cross-run `actions/download-artifact` step, since the run-id lookup it depends on was verified against live API data but the download itself only runs on a runner.

petarzarkov:
@previewdeploy

PR #135, opened 2026-09-13
chore: move every action to its current major

petarzarkov:
Every GitHub Action in `.github/` moved to its current major.

| Action | Was | Now |
| --- | --- | --- |
| `actions/checkout` | `v4`, `11d5960` (v4.4.0) | `v7`, `3d3c42e` (v7.0.1) |
| `actions/cache` | `v4` | `v6` |
| `actions/upload-artifact` | `v4` | `v7` |
| `actions/download-artifact` | `v4` | `v8` |
| `cloudflare/wrangler-action` | `9acf94a` (v3.15.0) | `ebbaa15` (v4.0.0) |
| `anthropics/claude-code-action` | `b7912ae` (v1) | `a73a99c` (v1) |
| `actions/github-script` | `3a2844b` (v9.0.0) | unchanged, already latest |
| `oven-sh/setup-bun` | `v2` | unchanged, `v2` is v2.2.0 |
| `petarzarkov/fast-code-review` | `ac2587f` (v1.3.1) | unchanged, already latest |

`claude-code-action` pins a commit behind a moving `v1`, and `v1` had been
retargeted since, so the pin moves with it.

## The majors crossed, and why none of them lands here

- **checkout v5, upload-artifact v5/v6/v7, download-artifact v6/v7/v8** are the
  node24 runtime and an ESM rewrite. They need runner 2.327.1 or newer;
  `ubuntu-latest` is well past it and there are no self-hosted runners.
- **checkout v6** stops writing the persisted token into `.git/config` and puts
  it in a file under `RUNNER_TEMP`, reached through an `includeIf`. The release
  job's tag push relies on that credential and still gets it. The deep
  reviewer's comment named `.git/config` specifically, so it is corrected here
  rather than left to mislead; the `persist-credentials: false` it argues for is
  still right, because the agent has a shell and `RUNNER_TEMP` is as readable to
  it as the workspace.
- **checkout v7** blocks checking out a fork PR, but only under
  `pull_request_target` and `workflow_run`. Neither trigger exists in this repo.
- **download-artifact v5** changed the output path for downloads by
  `artifact-ids`. Both downloads here name the artifact, which is the path that
  did not change.
- **wrangler-action v4** changes only the default `wranglerVersion`. Both
  deploys pin `4.128.0` explicitly.

The pinning split ci.yml documents is unchanged: GitHub's own read-only actions
stay on a major, third-party actions near a credential stay on a commit.

## Verification

`bun run ci` green on `static`, `docs`, `coverage` and `browser`. The `examples`
phase failed once on `tour.test.ts:376`, the AMQP `2 of 2 delivered to`
assertion, and passed on a rerun: a local broker timing race, and this diff is
`.github/` only.

## Summary by CodeRabbit

- **Chores**
  - Updated GitHub Actions used for dependency caching, source checkout, artifact handling, and preview/CI deployments.
  - Refreshed the pinned versions of the Cloudflare Wrangler and Claude Code actions.
  - Updated checkout credential handling documentation to reflect the newer action behavior.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -12,7 +12,7 @@ runs:
         bun-version: 1.4.2
 
     - name: Cache Bun dependencies
-      uses: actions/cache@v4
+      uses: actions/cache@v6
       with:
         path: ~/.bun/install/cache
         key: ${{ runner.os }}-bun-${{ hashFiles('**/bun.lock') }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     name: Lint, format, types
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
       - uses: ./.github/actions/setup
 
       - name: Build
@@ -42,7 +42,7 @@ jobs:
     # `coverage` runs the same files with every service attached. Declaring them
     # twice is what broke the release job: it ran the gate with none.
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
       - uses: ./.github/actions/setup
 
       - name: Build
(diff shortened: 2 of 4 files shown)


dunxonu (approved):

All GitHub Actions dependencies and composite setup steps have been cleanly updated to their latest major releases.

PR #136, opened 2026-09-13
chore: delete what nothing reaches, and declare the rest once

petarzarkov:
Net **-3,115 lines** across docs, dead symbols, and duplicated scaffolding. Every gate passes; the release path was proved unchanged by running the old and new scripts against the same worktree.

## Dead docs and records (-2,382)

`internal/notes/research/` keeps a record while its measurements outlive the decision. Four no longer do: scheduler, health, throttle and the releases sub-page all shipped. Their two unrecorded Bun findings moved into `docs/bun-apis.md` first; git history holds the probes. `oop-conversion.md` had no inbound reference anywhere. `bun-1.4-adoption.md` said to delete it when the adopt list was empty, and A2 landed in `HttpClientOptionsInit`, so it is.

ROADMAP's phase plan was 210 lines of met exit criteria under a heading saying all five phases are delivered. What survives is the Phase 1 decision, promoted to "The example ladder", which is what CONTRIBUTING and `/new-package` cite.

`tooling.md` documented `scripts/seo.ts`, which does not exist, and carried a heading stating the reversed bundler arrangement that its own text corrected 350 lines earlier.

Corrected against the tree rather than rewritten: `latest.json` is a 20-subject run not 16, the dashboard has seven panels not six, `published-voice.test.ts` still matched a deleted workspace, and `docs/logo/README.md` was the one page under `docs/` nothing could reach.

## Dead code (-150)

`internal/openapi-ui` was deleted when `@dunx/openapi` moved to `swagger-ui-dist`. Three `@dunx/ui` exports went to zero consumers with it: `statusColor`, the bullmq job-state map, and six icons only the explorer's request pane drew. The README claiming none went to zero named a `StatusBadge` that does not exist. `examples/minimal`'s two watch probes are referenced by nothing.

## One declaration for duplicated scaffolding (-582)

- `generate-responses.test.ts` carried a byte-for-byte copy of `generate-schemas.test.ts`'s three describes, so both suites ran the same assertions twice.
- `withApp` was eleven copies of the same `listen(0)`/`try`/`finally`. Each suite keeps its own over one fixture, because what differs is only how the app is built.
- The console capture was five copies with three different final filters; the Redis reachability probe four; `rejectionMessage` and friends seven.
- `relay.test.ts` re-declared eight symbols the fixture beside it already exported, and the two real-relay suites asserted cross-node delivery with the same forty lines.
- Three bench subjects shared a controller. It went to a new `servers/dunx-app.ts`, **not** `servers/shared.ts`: that file is imported by all twenty subjects, and an `@dunx/http` import in it would have put the framework in the boot path of every one, corrupting the rss and startup numbers the harness exists to produce.
- The repo root was derived twelve times in four ways, and `new URL('..').pathname` and `fileURLToPath` disagree on a path containing a space or a `#`.

## Deliberately not done

- **`rejectionMessage` into `@dunx/testing`**, which would collapse four more copies. `architecture/packaging.md` records that exact move being tried and reverted: a package's build typechecks its own tests, so it made openapi's build race testing's.
- **`cost-of-logging.md`'s 1.3.14 appendix.** It looked superseded by the 1.4 sections above it, but it records three suspicions as wrong and attributes the irreducible costs, which the later runs do not repeat.
- **`internal/docs/scripts/extract/model.ts`'s unused type aliases.** They are unreferenced by design: each one is a compile-time tripwire against a rename in the harness's own types.
- **`docs/logo/`'s six rejected concept SVGs and contact sheet** (~450 KB, referenced by nothing). The README says the reasoning is worth more than the drawings, so they are arguably deletable, but design source is your call.

## Verification

`bun run ci` is 13/13. The relay work is verified against a live Redis and a live Postgres. `readme-tables.ts` regenerates `internal/bench/README.md` byte-identically, and `bench smoke` answers every scenario on all 13 runnable subjects. For the release path, the old and the new scripts were run against the same detached worktree with an empty `release:` commit: both printed the same ten lockstep bumps and the same rendered changelog section, diffed identical.

Coverage is unmoved at 96.5% lines, every package above the 90% floor.


dunxonu (commented):

- `packages/infra/src/queue/worker.test.ts:515` File exceeds the 500-line limit

</details>

PR #138, opened 2026-09-13
feat(http): RelayPublisher and SocketObserver, the two seams an app reached into /internal for

### LINKED ISSUE #137: http: a worker cannot publish to the ws relay, encode and encodeRelay are unexported
A process that has no `Bun.serve` cannot put a socket frame on the relay channel, because the two functions that build one are not exported from any public subpath.

## The case

A job handler runs in two containers. In the web process `PubSub` is bound by `HttpFactory`, so `pubsub.publishEvent(topic, event, data)` works. In a worker built by `WorkerFactory` there is no server and no `PubSub`, so the only way to reach the web nodes' sockets is to publish the frame onto the relay channel directly.

`WsRelay` is injectable as of the relay module, so getting the channel is solved. Building the message is not:

```ts
export class RelayPublisher extends EventsPublisher {
  constructor(private readonly relay: WsRelay, config: AppConfigService) { super(); }

  override publish(topic: string, event: string, data: unknown): void {
    // encodeRelay and encode are not exported
    void this.relay.publish(this.#channel, encodeRelay(this.#origin, topic, encode(event, data)));
  }
}
```

## Where they are

- `encodeRelay` / `decodeRelay` are declared in `packages/http/src/ws/relay.ts:134` and `:152`, and they ship in `dist/ws/relay.d.ts`, but `packages/http/src/index.ts:201` re-exports only `DEFAULT_RELAY_CHANNEL`, `WsRelay`, `PubSubRelay` and `RelayOptions` from that module.
- `encode` / `decode` are in `packages/http/src/ws/envelope.ts:11` and are not exported at all.

`@dunx/http`'s `exports` map is `.`, `./client`, `./connect`, `./internal`, so `@dunx/http/dist/ws/relay.js` is not reachable either.

These were reachable before 3.0.1 removed the deprecated re-exports (05139dc).

## What a consumer does instead

Reimplements the wire format, which is a private contract that can change under them:

```ts
const encode = (event: string, data: unknown) => JSON.stringify({ event, data });
const encodeRelay = (o: string, t: string, d: string) => JSON.stringify({ o, t, d });
```

## Suggested fix

Export `encode`, `decode`, `encodeRelay` and `decodeRelay` from the root of `@dunx/http`, or from `@dunx/http/internal` if the root is meant to stay narrow. Either makes the worker-to-socket path a supported one rather than a copied constant.

Found while porting `nestjs-template` to dunx, where the NestJS original used `@socket.io/redis-emitter` for exactly this.

petarzarkov:
Closes #137.

`@dunx/http/internal` narrowed to what a sibling package imports, which dropped `observe`, `encode` and `encodeRelay`. Two consumer apps were importing them - petarzarkov/firecracker, and the port #137 was found in - because neither thing they do had a supported shape. Both gaps are closed as classes rather than by reopening `/internal` or widening the root with four codec functions.

**`RelayPublisher`** - the publish half of `PubSub` for a process that has no server. A queue worker or a forked job child has no `Bun.serve` for the local fan-out, so publishing a frame meant encoding the relay wire format by hand against a format this package owns. `WsRelayModule` binds it, on a channel its new second argument names, so it is a constructor parameter rather than an import. It never throws: a job handler that failed on a frame would be retried and repeat its side effects.

#137 suggested exporting `encode`, `decode`, `encodeRelay` and `decodeRelay` from the root. A class is the answer instead, for the reason the issue's own sketch shows: every consumer that holds those four functions writes the same origin id, the same `try`/`catch` and the same fire-and-forget around them. Shipping the class ships that too, and the wire format stays unexported, which is what lets it change.

**`SocketObserver`** - a `SocketMiddleware` base for one that only wants the outcome. A gateway handler may return a value or a promise, so such a middleware has to catch a throw, hook a rejection and rethrow both; getting it wrong swallows a failure silently. Subclasses implement one abstract `settled(error, value, frame, ctx)`.

Neither is a second declaration: `RelayPublisher` calls the same `encodeRelay` `PubSub` does, `SocketObserver` the same `observe` `SocketLoggingMiddleware` does.

`examples/full` takes both (Rule 4). `SocketTrailObserver` is the socket half of the request trail, riding in `AppHttpOptions.socketMiddleware`; the relay tour ends with a publish from a `RelayPublisher` with no server behind it, delivered by both nodes. The `websockets` scaffold feature now requires `http`, which is where `WsRelayModule` is declared.

`bun run ci`: 13 of 13.


petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

dunxonu (approved):

packages/http/src/ws/relay-publisher.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * **Never throws.** A job handler that failed here would be retried and would
+   * repeat its side effects to deliver a frame nobody awaited. Failures go to
+   * `onError`.
+   */
+  publish(topic: string, data: string | Bun.BufferSource): void {
+    try {
+      const result = this.relay.publish(
+        this.#channel,
```
dunxonu, line 83: RelayPublisher's channel and the server-side PubSub listening channel are configured from two independent, unvalidated sources, and a mismatch is swallowed silently because publish() is designed to never throw.

**How it fails:** An app sets a custom `AppHttpOptions.relayChannel` (consumed by `PubSub.relayThrough` on the server nodes) but calls `WsRelayModule.forRoot(redisOpts)` without the new second `publisher` argument (or with a different `channel`). `RelayPublisher` then publishes on `DEFAULT_RELAY_CHANNEL` ('dunx:ws') while the servers listen on the custom channel. `publish()` catches every synchronous throw and rejection and routes them to `onError`, but there is no error at all here - the broker simply delivers to zero subscribers - so frames vanish with no log, no exception, and no way to detect the misconfiguration short of noticing missing messages in production.

packages/http/src/ws/observer.ts
```diff
@@ -0,0 +1,58 @@
...
+   */
+  readonly reportsErrors: boolean = false;
+
+  handle(frame: SocketFrame, ctx: SocketContext, next: SocketNext): unknown {
+    return observe(next, (error, value) => {
+      this.settled(error, value, frame, ctx);
+    });
+  }
+
+  /**
```
dunxonu: SocketObserver.settled cannot distinguish a successful handler that returns undefined from a handler that synchronously `throw undefined` or rejects with `undefined`, because observe() reports both as `error === undefined`.

**How it fails:** A gateway handler does `throw undefined` (or rejects a promise with `undefined`, e.g. `Promise.reject()`) on some failure path. `observe()`'s catch calls `done(error, undefined)` with `error === undefined`, identical to the success signature `done(undefined, result)` when `result` is also undefined. Any subclass following the exact pattern this file's own doc example and `SocketTrailObserver` use (`if (error !== undefined) { ...report failure... }`) silently classifies the failure as success - e.g. `SocketTrailObserver` records it as 'no reply' instead of logging the handler failure, exactly the silent-swallow this abstraction was built to prevent.

packages/http/src/ws/relay-publisher.ts
```diff
@@ -0,0 +1,91 @@
...
+  encodeRelay,
+  WsRelay,
+  type RelayOptions,
+} from './relay.js';
+
+/** What a publisher needs beyond the relay itself. */
+export interface RelayPublisherInit {
+  /** The broker channel. Must match the `relayChannel` the servers listen on. */
+  readonly channel?: string;
+  /** Reports a publish the broker refused. Defaults to a `console.error`. */
```
dunxonu: RelayPublisherInit.onError's doc comment claims the default reporter is `console.error`, but the actual default (`defaultRelayError` in relay.ts) logs via `console.warn`.

**How it fails:** A consumer reads the JSDoc, assumes unhandled relay failures are logged at `console.error` severity (e.g. for log-level filtering or alerting that only captures `error`), and never sees the warning because it is actually emitted through `console.warn`, silently under-reporting broker failures in production log pipelines that filter by level.

dunxonu (commented):

dunxonu (commented):

petarzarkov:
All seven findings acted on in 3ba050d. Six were real; one I took differently than suggested.

**@dunxonu — `throw undefined` reads as success.** Right, and it was worse than a subclass hazard: the pattern this file's own example showed was `if (error !== undefined)`. `observe` passes `ok` now, and `settled` takes a `SocketOutcome` union - `{ ok: true, value } | { ok: false, error }` - so the ambiguity cannot be written. Two tests cover it, one per channel.

**@coderabbitai — `settled` throwing alters control flow.** Right. Caught and reported; the handler's value or error reaches the caller untouched. Tested on both channels rather than only the synchronous one.

**@coderabbitai — `onError` throwing breaks "never throws".** Right, on both paths: out of the synchronous `publish`, and as an unhandled rejection from the `.catch`. Both go through one guarded reporter.

**@dunxonu — the doc said `console.error`, `defaultRelayError` is `console.warn`.** Corrected.

**@coderabbitai — the demo's fixed 200 ms.** Taken. `await Promise.all([onA.next(), onB.next()])`, and `next()` carries its own 2 s deadline, so a slow broker fails the tour instead of reporting a delivery that had not happened. Vendored template synced.

**@coderabbitai — request review after CI.** Taken, moved below `bun run ci` with the reason.

**@dunxonu — a channel mismatch is silent.** Real, and not fully fixable here: two processes have to agree on a channel name, and a broker delivering to zero subscribers raises nothing to report. What I did was remove the way to get it wrong by accident - the option now states that the servers listen on `HttpOptionsProvider.relayChannel` and that a mismatch is silent, and says to read both from one constant, which is what `examples/full` does with `RELAY_CHANNEL`. I did not make the argument required: `WsRelayModule.forRoot()` with no channel and an app with no `relayChannel` both land on `DEFAULT_RELAY_CHANNEL`, which is correct, and failing that case to protect the one where an app sets only one of the two would break the common path to guard the misconfigured one.

Also in this branch, found by the same consumer upgrade: dac2e84. `StorageError` set `this.name = new.target.name`, which reads the *emitted* class name, and the bundler renames a class whose name collides in a shared chunk - 3.8.1 shipped `FileNotFoundError` as `FileNotFoundError2`. Since each `@dunx/infra` subpath is its own bundle, `instanceof` is false across one and `error.name` is what a call site matches on, so a consumer failing a job unrecoverably on `'FileNotFoundError'` retried it six times instead. The regression test is in `examples/full`, which resolves the built package; from source it always read correctly, which is why 217 test files missed it.

`bun run ci`: 13 of 13.

dunxonu:
No question was asked in that message — it is a status update summarising how each review finding was addressed. If there is something specific you would like me to look at or confirm, let me know.

dunxonu (approved):

PR #141, opened 2026-09-13
feat(openapi): authorize on OpenApiModule

### LINKED ISSUE #140: openapi: OpenApiModule has no authorize, so the explorer cannot be gated the way the dashboard can
`DashboardModule` takes an `authorize` callback and `OpenApiModule` does not, so the two ops surfaces a dunx app exposes have different answers to "who may see this".

## The asymmetry

`@dunx/dashboard` gets this right. `DashboardOptionsInit.authorize` (`packages/dashboard/src/options.ts:46`) receives the raw `BunRequest`, is documented as running ahead of any session guard, and a refused request gets 404 rather than 403 so the mount is not discoverable. It even warns at boot when it is left unset.

`OpenApiModule` has no counterpart. Its controller declares its routes `@Public()`, so a global `SessionGuard` waves them through by design, and `/api/docs` plus `/api/openapi.json` answer to anyone who can reach the port in every environment.

For a public API that is correct and desirable. For a service whose OpenAPI document is its whole route table, including admin operations and their `x-required-roles`, it is a decision the app should get to make, and right now it cannot make it through the module that mounts the page.

## What a consumer writes instead

A middleware registered ahead of `SessionGuard`, matching the mount paths by hand:

```ts
export class DocsSessionMiddleware implements Middleware {
  readonly #guarded = [`/${prefix}/${docs.path}`, `/${prefix}/${docs.jsonPath}`];

  async handle(req: BunRequest, _ctx: RouteContext, next: Next) {
    const { pathname } = new URL(req.url);
    // startsWith, not equality: both explorers serve assets under their mount,
    // and a page whose script is a 401 is broken rather than gated.
    if (!this.#guarded.some((p) => pathname === p || pathname.startsWith(`${p}/`))) {
      return next();
    }
    const session = await this.auth.api.getSession({ headers: req.headers });
    return session === null ? this.#loginForm() : next();
  }
}
```

The path list is the part that rots: it restates configuration the module already holds, and an app that changes `path` or adds the second renderer has to remember to update it.

## Suggested fix

`authorize?: Authorize` on `OpenApiOptions`, with the same contract the dashboard already has: raw request in, boolean out, 404 on refusal, and the explorer's own assets covered by the same decision. Reusing the `Authorize` type would make the two surfaces configurable the same way.

Two smaller things in the same area:

- A **browser** reaching a gated page has a cookie and no way to send a bearer token, so a bare 401 is a dead end. The dashboard has the same property. Somewhere to render a challenge, or a documented recipe, would help.
- The second renderer is mounted by hand today (`examples/full/src/reference/reference.middleware.ts` is the pattern, and it is a good one), which means any `authorize` added to the module would cover Swagger and not Scalar unless the app repeats itself. Worth deciding deliberately.

Found while porting `nestjs-template` to dunx: its `HtmlSessionAuthMiddleware` gated Swagger, Scalar and Bull Board behind one session check. Bull Board's half is now `DashboardModule`'s `authorize`; the docs half had to stay a hand-written middleware.

petarzarkov:
Closes #140.

`DashboardModule` took an `authorize` callback and `OpenApiModule` did not, so the two ops surfaces a dunx app exposes had different answers to "who may see this". An app wanting the explorer gated wrote a middleware restating the module's mount paths by hand.

## What changed

- `Authorize` moves to `@dunx/http`, the lowest common owner of the two packages that take one. `@dunx/dashboard` re-exports it; nothing an existing consumer imports moves.
- `OpenApiOptions`/`OpenApiInfo` take `authorize`. It covers the document, the page and the page's assets in one decision, and it comes out of `forRootAsync`'s factory too, which is the only way to close over an `Auth` the container owns.
- `AuthorizeDecision` widens to `boolean | Response`. A browser arrives with a cookie and no way to attach a bearer token, so a bare refusal is a dead end; a returned `Response` is sent as written. Existing gates returning `boolean` are unaffected.
- `gate()` is the shared runner: `undefined` to carry on, or the response to send. `DashboardMiddleware` now calls it instead of building its own 404, so both surfaces refuse identically.

No default and no boot warning on the openapi side, unlike the dashboard: a public API's document is published to be read.

## The second renderer

`OpenApiModule` gates what it mounts. A page mounted by hand is a middleware of your own, gated by being handed the same function. `examples/full/src/docs-gate.ts` is one class doing that for Swagger UI at `/api/docs` and Scalar at `/api/reference`, behind `DOCS_GUARDED` so the demo stays readable.

## Verification

`bun run ci` is green except `examples/tests`, where the AMQP tests in `examples/full` time out. That reproduces on the base commit in the same environment and is broker contention, not this change.

## Summary by CodeRabbit

- **New Features**
  - Added optional authorization for OpenAPI documentation, including the document, UI, and assets.
  - Authorization can deny access with a 404 or return a custom response, such as a sign-in redirect.
  - Added shared authorization support for dashboard and OpenAPI surfaces.
  - Added an example configuration for protecting documentation with active sessions.

- **Documentation**
  - Documented authorization behavior, configuration, custom responses, and manual route protection.

- **Tests**
  - Added coverage for authorization outcomes, redirects, asynchronous checks, and publicly accessible documentation.


petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

dunxonu (commented):

packages/http/src/server/authorize.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * A surface that announces itself to an unauthenticated caller has told them
+ * where to keep knocking, so the refusal is the body an unmatched path already
+ * gets, down to the field names.
+ */
+export const gate = async (
+  authorize: Authorize | undefined,
+  req: BunRequest,
+): Promise<Response | undefined> => {
+  if (authorize === undefined) return undefined;
```
dunxonu, line 27: gate() fails open (admits the request) whenever an Authorize callback resolves to anything other than exactly true, exactly false, or a Response instance.

**How it fails:** A consumer's authorize implementation (in their own app, e.g. `async (req) => { if (await signedIn(req)) return true; if (isBrowser(req)) return new Response(...); }` with the trailing `return false;` omitted) resolves to `undefined` on the 'not signed in, not a browser' path. `decision === true` is false and `decision === false` is false, so gate() executes `return decision`, returning `undefined` from gate() itself. Every caller (`DashboardMiddleware.handle`, `OpenApiController.refuse` via `documentController`) treats `refused !== undefined` as the only refusal signal, so an unauthenticated request is silently served the dashboard or the OpenAPI explorer instead of getting a 404. Neither packages/http/src/server/authorize.test.ts nor packages/openapi/src/authorize.test.ts exercises this case, so the gap is untested.

examples/full/src/docs-gate.ts
```diff
@@ -0,0 +1,45 @@
...
+   * established the caller. A cookie, so the page's asset fetches carry it too.
+   */
+  async admits(req: BunRequest): Promise<AuthorizeDecision> {
+    const session = await this.auth.api.getSession({ headers: req.headers });
+    if (session !== null) return true;
+
+    // `false` is a 404 and a dead end for a person, so send a navigation
+    // somewhere it can sign in. Everything else gets the 404.
+    return req.headers.get('accept')?.includes('text/html') === true
+      ? new Response(null, { status: 302, headers: { location: '/' } })
```
dunxonu: The example's browser-refusal path redirects to '/' rather than an actual sign-in destination, and the landing page has no sign-in flow to land on.

**How it fails:** With DOCS_GUARDED=true, a browser without a session requesting /api/docs gets a 302 to '/', but examples/full/src/landing has no sign-in UI or anonymous-session bootstrap (unlike the Swagger requestInterceptor in main.ts, which auto-signs-in anonymously). The visitor lands on the homepage with no way to actually obtain the session the gate is waiting for, contradicting the PR's stated rationale that a Response gives 'a person somewhere to go' and diverging from the guide's own `/sign-in` example in docs/guide/10-openapi.md.

dunxonu (commented):

examples/full/src/reference/reference.middleware.ts
```diff
@@ -27,7 +32,10 @@ export class ReferenceMiddleware implements Middleware {
   });
   #page: Promise<string> | undefined;
 
-  constructor(private readonly explorer: OpenApiExplorer) {}
+  constructor(
+    private readonly explorer: OpenApiExplorer,
+    private readonly docs: DocsGate,
```
dunxonu, line 37: 🛑 `correctness` ReferenceMiddleware now injects DocsGate in its constructor, but DocsGate is provided by DocsGateModule, which is not imported by ReferenceModule or AppModule.

**How it fails:** When the application boots and attempts to resolve ReferenceMiddleware via app.use(ReferenceMiddleware), the DI container will throw a resolution error because DocsGate is not visible in ReferenceModule's scope.

```suggestion
Import DocsGateModule into ReferenceModule (or make DocsGateModule global if appropriate) so that ReferenceMiddleware can successfully resolve its DocsGate dependency.
```

dunxonu (commented):

Add optional authorization support to OpenApiModule and unify authorization decision handling across OpenAPI and Dashboard surfaces.

packages/http/src/server/authorize.ts
petarzarkov: Fixed in 504be56. Only `true` admits and only a `Response` replaces the refusal now; everything else falls through to the 404, so an `authorize` that runs off the end of a branch closes rather than opens. Covered by a case in `authorize.test.ts` over `undefined`, `null`, `0`, `''`, a non-empty string and an object.

examples/full/src/reference/reference.middleware.ts
petarzarkov: Checked this rather than applied it: `DocsGate` does resolve. `app.use(ReferenceMiddleware)` resolves from the root module ref, and the root here is `OpenApiModule.forRootAsync(...)`'s configured module, which lists `DocsGateModule` in its `imports` (main.ts). That is the same route `OpenApiExplorer` takes. `reference.test.ts` already boots the whole app through `createApp()` and passes, which it could not do if the container threw at `app.use`.

The real gap your comment points at is that nothing exercised the gated path, since `DOCS_GUARDED` defaults to false. 504be56 adds `examples/full/src/docs-gate.test.ts`, which boots the example with it on and walks all five gated paths across both explorers, signed out and signed in.

examples/full/src/docs-gate.ts
petarzarkov: Half right, and fixed in 504be56. The landing page does have an anonymous-session bootstrap - `landing.js`'s `signedIn()` POSTs `/api/auth/sign-in/anonymous` - but it is behind the 'Who am I' button, so a bare `/` did leave the visitor to find it.

The redirect now goes to `/#who`, which is that button. I first gave the Session section an `id` of its own and `landing.test.ts` refused it, correctly: every `id` in `index.html` has to be an element `landing.js` drives, so a scroll-only anchor breaks that contract. The button's own id is the better target anyway, since it is the control that issues the session.

The comment on `admits` now says the rest plainly: that bootstrap only exists when `AUTH_GUEST_ONLY` is set, so guarding the docs without it leaves nothing to land on. The new test asserts both the 302 target and that the anchor exists in the served page.

dunxonu (approved):

Authorize callback support has been added to OpenApiModule, allowing OpenAPI documentation and assets to be gated using the same authorization policies as the dashboard.

packages/http/src/server/authorize.ts
dunxonu: Fixed and tested — dropping the finding.

examples/full/src/docs-gate.ts
dunxonu: Fixed and verified — the redirect now targets the control that actually issues the session, and the test covers both the 302 location and the anchor's presence.

PR #143, opened 2026-09-13
feat(http): probe the broker and the object store

petarzarkov:
`HealthModule` reached Redis, a database, memory and a disk. RabbitMQ and
whatever `Storage` an app configured had no probe.

**`AmqpIndicator`** takes `AmqpConnection`, which grows a `ping()`. It opens the
connection if nothing has yet, returns at once while the socket is established
and unblocked, and otherwise throws the socket's own last failure rather than
waiting `readyTimeoutMs` out to repeat it. Measured against a refused port:
1506 ms for the first probe, 0 ms and `connect ECONNREFUSED 127.0.0.1:1` after.
Auto-close is disabled, so a slow answer cannot take the connection away.

**`StorageIndicator`** takes `Storage`, one `exists()`: a HEAD against S3, a stat
locally, no object required and nothing left behind. `DiskIndicator` measures a
local volume and stops meaning anything on S3, so both belong when uploads land
on one.

**`RoundTripIndicator`** holds the body `RedisIndicator` and `DatabaseIndicator`
had each written out; `AmqpIndicator` would have been a third. Exported, so a
custom probe with a `ping()` is a name.

`examples/full` runs both. The broker indicator is composed in `ProbesModule`
rather than in `IndicatorsModule`, because `MessagingModule` sets `consume: true`
and `OpsModule` takes `AppIndicators` for the ops page: wiring it the other way
made every dashboard slice open consumers, 2 to 10 in `dashboard.test.ts`.

Considered and left out: a bullmq indicator, since `QueueConnection.client()`
hands back a fresh socket per call and the queue's dependency is a Redis a
`RedisIndicator` already covers; and an HTTP upstream indicator, which
`examples/full` could only point at itself.


dunxonu (commented):

Introduce health indicators for AMQP and Storage, but fix a bug where AMQP connection auto-close is accidentally enabled instead of disabled.

packages/infra/src/amqp/connection.ts
```diff
@@ -41,6 +45,29 @@ export class AmqpConnection implements OnShutdown {
...
+    const failure = this.#lastError;
+    if (failure !== undefined) {
+      throw new AmqpError(
+        AmqpErrorCode.UNREACHABLE,
+        `the AMQP broker is unreachable: ${reasonFor(failure)}`,
+        failure,
+      );
+    }
+
+    await connection.onConnect(this.#options.readyTimeoutMs, true);
```
dunxonu, line 68: 🛑 `correctness` The code passes `true` as the second argument to `onConnect`, which enables auto-close on timeout. To disable auto-close as intended by the comments and PR description, `false` must be passed.

**How it fails:** When the AMQP broker is slow to respond during a ping, `onConnect` times out and automatically closes the connection because `autoClose` is set to `true`, preventing subsequent reconnection attempts or causing unexpected connection teardown.

```suggestion
await connection.onConnect(this.#options.readyTimeoutMs, false);
```

petarzarkov: The parameter is `disableAutoClose`, not `autoClose`, so `true` is what disables the close. From rabbitmq-client 5.0.8, `lib/Connection.js`:

```js
onConnect(timeout = 10_000, disableAutoClose = false) {
  ...
  timer = setTimeout(() => {
    ...
    if (!disableAutoClose) {
      this.close().catch(() => { });
    }
```

`close()` runs only when `disableAutoClose` is falsy, so passing `false` here is what would produce the failure described.

The comment did catch a weak test, though. `leaves the connection retrying after a ping that timed out` asserted `amqp.opened`, which reads `#connection !== undefined` and is never cleared on a library-side close, so it passed either way. It now counts `close()` calls, and fails when the argument is flipped to `false`. Pushed in 444cb27.

dunxonu: You're right — I misread the parameter name. `disableAutoClose: true` does disable the auto-close, which matches the comment's intent. My finding was incorrect.

dunxonu (approved):

The change successfully adds health probing for the AMQP broker and the configured object storage, with robust error handling and immediate failure reporting on known connection errors.

dunxonu (approved):

The change successfully adds health probes for the AMQP broker and object storage, with appropriate updates to the full example, tests, and template configuration.

PR #144, opened 2026-09-13
feat(infra): an email area with a transport seam, a renderer and a preview server

### LINKED ISSUE #139: infra: an EmailModule, with a transport seam, a template renderer and a Bun-native preview server
Email is the one outbound transport a production service almost always needs, and it is currently the one dunx says nothing about. Every app rewrites the same three things: a provider client, a template renderer, and a way to look at what the template produces without sending it.

## The shape that would match the rest of dunx

The same peer-integration pattern as drizzle, better-auth and bullmq: dunx owns the abstraction and the lifecycle, the vendor stays a peer the consumer installs.

```ts
EmailModule.forRootAsync({
  useFactory: (config: AppConfigService) => ({
    transport: new ResendTransport({ apiKey: config.get('email').apiKey }),
    from: config.get('email').sender,
    // Non-production sends nowhere and logs instead, which is the behaviour
    // every template writes by hand.
    dryRun: !config.get('isProd'),
  }),
  inject: [AppConfigService],
});
```

`EmailTransport` as an abstract class, so it is a runtime value a constructor parameter can name, the way `Storage` and `Logger` already are:

```ts
export abstract class EmailTransport {
  abstract send(message: EmailMessage): Promise<EmailResult>;
}
```

Shipped implementations would be thin: `ResendTransport`, an SMTP one, `LogTransport` as the default so an app boots and is testable with nothing configured, and `MemoryTransport` for `@dunx/testing` so a suite can assert what would have been sent.

## Three things worth owning rather than leaving to each app

**Rate limiting.** Resend enforces a per-second cap and the failure mode is a 429 in a job handler. `nestjs-template` carries a hand-rolled serialised queue for exactly this (`src/notifications/email/services/email.service.ts:63-68`, a promise chain plus a `lastSentAt` clock). `ResiliencePolicy` already exists in `@dunx/core`, so this is composition rather than new machinery.

**Templates.** `render` from `@react-email/components` is the common choice and works fine on Bun, but it should not be the only one. A `TemplateRenderer` interface with a React Email implementation as one option leaves a tagged-template or MJML renderer possible.

**A preview server.** This is the part with the most leverage. The `react-email` CLI pulls `@react-email/ui` and Next.js, roughly 430 MB of node_modules, to serve a page that lists templates and renders one. `nestjs-template` replaced the whole thing with about 120 lines over `Bun.serve` (`scripts/email.ts`), rendering through the exact same `render` call the runtime uses, with a cache-busting dynamic import so edits show up without a restart.

That is squarely a `bunx dunx-email preview` tool, and it is the kind of thing where "Bun already does this" is the entire argument: no CLI framework, no bundler, no dev server, just `Bun.serve` and `import()`.

## Why it is not just an app concern

The counter-argument is the one MAPPING.md already makes about Resend in the dunx template port: a transport the consumer owns, like an AI provider. That is right about the **vendor** and wrong about the **shape**. dunx does not ship a database either, and `@dunx/infra/db` is still worth having, because the lifecycle, the injection token, the health probe and the test double are the framework's problem. Email is the same: what is app-specific is the API key and the copy, not the queueing, the dry-run default, the renderer seam or the preview server.

Raised while porting `nestjs-template` to dunx, where `EmailService` had to be reduced to a method that logs the message it would have sent.

petarzarkov:
Closes #139.

`@dunx/infra/email` binds `EmailOptions`, `EmailTransport`, `TemplateRenderer` and `EmailService`. The service owns the default sender, per-second pacing, the retry policy around one send, and the rendering call. The provider does not live here.

Four subpaths, split on peers the way `@dunx/openapi` splits its renderers. The base reaches none, so it stays in the root barrel; `/email/resend`, `/email/smtp` and `/email/react` each reach one and do not.

`bunx dunx-email preview` is a `bin` on `@dunx/infra`, as `dunx-openapi` already is. It is `Bun.serve`, `Bun.Glob` and `import()`: the react-email CLI pulls `@react-email/ui` and Next.js to serve the same page through the same `render` call.

Three decisions against the obvious default, recorded in ROADMAP.md:

- Retries are off unless asked for. A provider that accepted a message and lost the response cannot be told from one that never saw it, and the cost of guessing is a second email in a real inbox. `maxPerSecond` answers a 429.
- `dryRun` routes to `LogTransport` rather than being a second mechanism, so there is one definition of sending nowhere.
- `MemoryTransport` ships in `/email`, not `@dunx/testing`, whose peers are core and http: a transport there would make `@dunx/infra` a peer of the harness.

`examples/full` gains a mail slice with two React Email templates, wired into the tour and asserted against `MemoryTransport`. Its tsconfig takes `jsx: react-jsx`; resolution stays nodenext.

The no-redis tour test takes the 20 s budget its `beforeAll` already has. It boots a second whole app in a process of its own and was on the 5 s default.

`bun run ci` passes every phase. `infra` coverage is 94.6% lines, 93.9% functions.

## Summary by CodeRabbit

- **New Features**
  - Added email infrastructure with Log, Memory, Resend, and SMTP delivery options.
  - Added React Email rendering, reusable templates, address validation, rate limiting, and configurable retries.
  - Added email template preview and HTML export through the `dunx-email` command.
  - Added welcome and receipt email examples, preview endpoints, and configurable dry-run support.

- **Documentation**
  - Added comprehensive email guidance and updated package roadmaps, README content, and feature listings.

- **Tests**
  - Added broad coverage for email delivery, rendering, previews, templates, configuration, and address handling.


petarzarkov:
@dunxonudeepreview

New area: `@dunx/infra/email`. Four subpaths, a `bin`, and a mail slice in `examples/full`. Worth the most scrutiny:

- `packages/infra/src/email/pacer.ts` - the serialised gate. Is the promise chain correct under concurrent `wait()`, and can a rejection poison it?
- `packages/infra/src/email/service.ts` - the pacer sits inside the `ResiliencePolicy`, so a retry is paced. Check `resolve()` for a message that names `cc`/`bcc` but no `to`.
- `packages/infra/src/email/templates.ts` - `loadTemplate` builds an `import()` specifier from a path found by `Bun.Glob`, with a cache-busting query. Anything reachable there that should not be?
- `packages/infra/src/email/preview.ts` and `preview-page.ts` - the preview server renders attacker-authorable template names into HTML. Escaping is in one place; confirm every sink goes through it.
- `packages/infra/src/email/message.ts` - `formatAddress` escapes a display name into an RFC 5322 header. Header injection is the thing to break.
- `resend/index.ts` and `smtp/index.ts` - both build a payload field by field because an explicit `undefined` is rejected. Check for a field that silently drops.

Also worth a verdict on two judgement calls: retries defaulting to off (duplicate-send hazard vs. a 429 that is safe to retry), and `MemoryTransport` living in `@dunx/infra/email` rather than `@dunx/testing`.

dunxonu (approved):

Adds an email abstraction with pacing, modular transports for Resend, SMTP, logging, and in-memory testing, plus a template preview server.

petarzarkov:
@coderabbitai review

packages/infra/src/email/cli.ts
```diff
@@ -0,0 +1,113 @@
...
+ * is testable without a port and without the React peers installed.
+ */
+export const plan = (argv: readonly string[]): Plan | string => {
+  const [command = 'preview'] = argv;
+  if (command === '--help' || command === '-h') return USAGE;
+  if (command !== 'preview' && command !== 'export') {
+    return `Unknown command "${command}".\n\n${USAGE}`;
+  }
+  const rest = argv.slice(1);
+  const positional = rest.filter((a) => !a.startsWith('--'));
```
dunxonu: The CLI's argv parser collects a flag's value as a positional whenever a flag appears before the directory argument, silently misresolving `dir`.

**How it fails:** `dunx-email preview --port 4000` (no directory given, the common case of overriding just the port) → rest=['--port','4000'] → positional=['4000'] (neither token starts with '--') → dir resolves to './4000' instead of the documented default './emails'; the preview server serves the wrong directory and reports it lists no templates. Not covered by cli.test.ts, where every case places the positional first.

examples/full/src/email/email.module.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Notices } from './notices.service.js';
+
+/**
+ * One `EmailTransport` per configured provider, chosen once at boot.
+ *
+ * Each vendor sits on its own subpath, so importing `@dunx/infra/email` alone
+ * needs neither `resend` nor `nodemailer`. This app imports all three because it
+ * demonstrates all three; a real app imports the one it sends with.
+ */
+const transportFor = (
```
dunxonu, line 28: `transportFor` constructs the concrete transport eagerly inside the async factory, before `dryRun` is ever consulted, so a misconfigured provider crashes boot even when dryRun would have made the transport unused.

**How it fails:** Set EMAIL_TRANSPORT=resend and leave EMAIL_RESEND_KEY unset (config.ts:72 only marks it `.optional()`, no refinement tying it to the selected transport) with the default EMAIL_DRY_RUN=true. `new ResendTransport({ apiKey: '' })` throws EmailSendError synchronously before EmailModule's own dryRun-aware binding (module.ts) ever gets a chance to substitute LogTransport, so the app fails to boot despite dryRun being on - contradicting the module's own doc comment that flipping dryRun is 'the whole difference between this and production'.

packages/infra/src/email/preview-page.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Reloads the frame on a timer rather than over a socket. The server re-executes
+ * the module on every request, so a poll is the whole of live reload and it
+ * costs no watcher, no websocket and no client bundle.
+ */
+const SCRIPT = `const f=document.getElementById('frame'),live=document.getElementById('live');
+setInterval(()=>{try{if(live.checked)f.contentWindow.location.reload()}catch{}},1500);`;
+
+export const escapeHtml = (value: string): string =>
```
dunxonu: Hand-rolled `escapeHtml` reimplements `Bun.escapeHTML`, which the codebase already uses for the same purpose elsewhere, violating Rule 1 ("native implementations only"), and is strictly weaker (drops the `'` escape the existing usages rely on).

**How it fails:** packages/dashboard/src/html.ts and packages/openapi/src/shell.ts both already call `Bun.escapeHTML` for the identical job of putting untrusted strings into an HTML shell. This new copy omits escaping `'`, so a template name containing a single quote inside a single-quoted HTML attribute elsewhere in the page (or a future edit that adds one) would not be neutralised the way the existing helper already guards against.

packages/infra/src/email/smtp/index.ts
```diff
@@ -0,0 +1,108 @@
...
+  readonly name = 'smtp';
+  readonly #mailer: SmtpMailer;
+
+  constructor(init: SmtpTransportInit) {
+    super();
+    if (init.mailer !== undefined) {
+      this.#mailer = init.mailer;
+      return;
+    }
+    const config = init.transport ?? init.url;
```
dunxonu: `SmtpTransport`'s constructor only treats `undefined` as "missing" config, unlike `ResendTransport` which explicitly checks for both `undefined` and `''`, so an empty SMTP URL slips past the guard.

**How it fails:** examples/full/src/email/email.module.ts:31 builds `new SmtpTransport({ url: email.smtpUrl ?? '' })`. With EMAIL_SMTP_URL unset, `config = init.transport ?? init.url` evaluates to `''` (not undefined), so the `config === undefined` guard at line 50 is skipped and `createTransport('')` is called directly, producing nodemailer's own unclear internal error instead of the intended `EmailSendError('smtp', 'SmtpTransport needs a url, a transport config or a mailer.')`.

packages/infra/src/email/message.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * RFC 5322 `"Name" <address>`, or the bare address when there is no name. The
+ * name is quoted and its own quotes and backslashes escaped, so a display name
+ * carrying a comma or an angle bracket cannot forge a second recipient.
+ */
+export const formatAddress = (recipient: EmailRecipient): string => {
+  const { address, name } = toAddress(recipient);
+  if (name === undefined || name === '') return address;
+  return `"${name.replace(/["\\]/g, '\\$&')}" <${address}>`;
```
dunxonu, line 155: `formatAddress`/`toAddress` never validate or escape the `address` half of a recipient (only the display `name` is escaped, and even that escape does not strip CR/LF), so a raw string handed to `to`/`cc`/`bcc`/`from`/`replyTo` can forge extra recipients or headers.

**How it fails:** A caller passes `to: 'a@example.com,evil@attacker.com'` or an address containing `\r\n` straight through (nothing in EmailService.resolve or toAddress validates email shape); SmtpTransport joins recipients with `.join(', ')` (smtp/index.ts) and nodemailer's address parser can split an embedded comma back into a second RCPT, silently adding a recipient the caller never intended. Only mail.controller.ts's `to` field happens to be guarded by `z.email()` in the example app; `cc`, `bcc`, `from` and `replyTo` have no such guard anywhere in the library itself.

packages/infra/src/email/cli.ts
dunxonu: The CLI hand-rolls flag/positional splitting instead of reusing `node:util`'s `parseArgs`, which the repo already uses correctly for the identical job and which is the root cause of the ordering bug above.

**How it fails:** tools/create-app/src/cli.ts already parses argv via `parseArgs({ allowPositionals: true, ... })`, which handles interleaved flags and positionals correctly by construction. Reimplementing the split with `rest.filter((a) => !a.startsWith('--'))` duplicates logic Rule 2 says should have been searched for first, and reproduces a class of bug the existing helper doesn't have.

dunxonu (commented):
**Actionable comments posted: 7**

packages/infra/src/email/smtp/index.ts
```diff
@@ -0,0 +1,108 @@
...
+      // and can be shorter than what was asked for.
+      accepted: (info.accepted ?? []).map((a) =>
+        typeof a === 'string' ? a : a.address,
+      ),
+      transport: this.name,
+    };
+  }
+}
+
+const payload = (message: OutboundEmail): Record<string, unknown> => {
```
dunxonu: `smtp/index.ts`'s `payload()` and `resend/index.ts`'s `payload()` (line 74) are near-identical field-presence builders that will need to be edited in lockstep for every future `OutboundEmail` field.

**How it fails:** Adding a new optional field to `OutboundEmail` (e.g. a `priority` header) requires remembering to update both transports' payload builders separately; missing one silently drops the field for only one provider with no compiler error to catch it.

petarzarkov:
All seven findings acted on; five were real defects. Pushed as `94aa718`.

| Finding | Verdict | Fix |
| --- | --- | --- |
| `formatAddress` does not validate the address half | **real, and the worst of them** | `toAddress` refuses a newline, comma, semicolon, angle bracket or missing `@`, on every field. Refused rather than escaped: there is no rendering of two addresses inside one recipient that means what the caller wrote. It also now splits `Name <addr>`, which used to land whole in the address slot |
| CLI collects a flag value as the positional | **real** | Reproduced: `preview --port 4000` served `./4000`, and `export --out /tmp/x` read templates from `/tmp/x`. Now `parseArgs` from `node:util` |
| CLI should reuse `parseArgs` (Rule 2) | **real** | Same fix. `@dunx/create-app` was already doing it |
| `SmtpTransport` accepts `` as a url | **real** | `` counts as missing, matching `ResendTransport` |
| `examples/full` builds the transport before `dryRun` | **real** | Degrades to the log transport and warns, rather than killing boot for a credential a dry run never uses |
| `escapeHtml` reimplements `Bun.escapeHTML` | **real, Rule 1 and Rule 2** | Deleted. `@dunx/openapi` and `@dunx/dashboard` were already calling Bun's |
| Two `payload()` builders drift in lockstep | **fair** | One `toPayload` taking the only two things providers disagree about, exported so a third-party transport gets the same walk. A test asserts a fully populated `OutboundEmail` reaches the payload whole, which is the guard neither copy had |

New tests cover each: header injection across `to`/`cc`/`bcc`/`from`/`replyTo`, the flag-ordering cases, the empty SMTP url, and the payload completeness guard. `bun run ci` passes every phase; `infra` coverage 94.6% lines / 93.9% functions.

One consequence worth flagging: `EMAIL_FROM='dunx-full <no-reply@dunx.win>'` now resolves to `{ name: 'dunx-full', address: 'no-reply@dunx.win' }` instead of stuffing the whole string into `address`. That is the bug being fixed, but it changes what `EmailResult`/`OutboundEmail` carry for anyone configuring a sender that way.

petarzarkov:
@coderabbitai review

dunxonu (commented):

petarzarkov:
CodeRabbit: 10 findings, 7 acted on, 2 declined with reasons, 1 already fixed. Pushed as `11b57bf`.

**The important one: the subject was never validated, and it is a header.** I refused a newline in addresses and display names and stopped there. `subject: `Welcome, ${name}`` with a CR/LF in `name` still added a `Bcc`. That is the likeliest injection route of the lot and it was open. `assertHeaderSafe` now covers the subject and every `headers` entry, in the same place addresses are checked, rather than in one controller's zod schema.

| Finding | Verdict |
| --- | --- |
| `name` reaches the subject unsanitised | **real** — fixed in the library, not the example schema |
| Resend attachments not base64-encoded | **real** — a string went as itself where base64 was required, so the attachment arrived as whatever those bytes decoded to |
| Partial SMTP delivery reported as total failure | **real** — threw after the message was already in the accepted inboxes, so a retry sent a second copy. `EmailResult` carries `rejected`; only a send with nothing accepted throws |
| Rejected provider promise escapes raw | **real** — a dead socket surfaced a `nodemailer`/`resend` error to a caller catching `EmailSendError`. Both wrapped, original as `cause` |
| Preview binds `0.0.0.0` | **real** — unauthenticated routes, one of which answers with a stack trace. `127.0.0.1` unless `hostname` says otherwise |
| `@react-email/components` deprecated | **real, with a caveat** — see below |
| `EMAIL_RESEND_KEY=` parses as `` | **real** — got past the example's own guard and threw where it meant to degrade |
| `--renderer` path resolution | already fixed in `b296ca2`; adopted the `pathToFileURL` refinement so a Windows drive letter is not read as a protocol |
| Make the payload fixture `Required<OutboundEmail>` | **declined** — `OutboundEmail` has no optional properties, so `Required<T>` is a no-op here and a new field already fails to compile in `outbound.fixture.ts` until it is added. The guard exists |
| Migrate to `react-email` | **declined as written** — `react-email` v6 is the CLI: it depends on esbuild, socket.io, chokidar, glob and tailwindcss, which is the 430 MB dev server this subpath exists to replace, and `glob`/`chokidar` are on Rule 1's banned list. But the deprecation is real, so the peer moved to `@react-email/render` 2.1.0, which is not deprecated and holds the one function the renderer calls. `@dunx/infra` now depends on neither the deprecated package nor the CLI |

Also worth recording: I had not run the preview server against real React Email templates before this round, only against a fixture renderer. Doing so turned up three more defects, fixed in `b296ca2` — a missing directory printed a raw ENOENT stack, `--renderer ./mine.ts` resolved inside `node_modules/@dunx/infra/dist/`, and the CLI threw instead of exiting 1.

`bun run ci` passes every phase; `infra` 94.7% lines / 94.0% functions.

dunxonu (commented):

The change introduces a comprehensive email infrastructure with multiple transport options, React Email rendering, and a preview server, along with thorough testing and configuration.

dunxonu (commented):

dunxonu (commented):

PR #145, opened 2026-09-13
fix(transform): an ambient type annotation emitted a name with no runtime binding

### LINKED ISSUE #142: transform: a type-only parameter annotation is recorded as a value reference
Found by `publish-guard` while verifying the 3.8.2 tarballs. **Pre-existing** - the emitted code is byte-identical in 3.8.1, so this is not a 3.8.2 regression.

## What happens

`@dunx/transform` records a constructor parameter whose type has no runtime binding as a value reference rather than as `unresolved`. `FileNotFoundError` in `@dunx/infra/files` is the case that surfaced it:

```ts
export class FileNotFoundError extends StorageError {
  constructor(
    readonly key: string,
    options?: ErrorOptions,
  ) { ... }
}
```

`ErrorOptions` is a TypeScript lib type. The emitted thunk names it as a value:

```js
Object.defineProperty(FileNotFoundError2, Symbol.for("dunx.deps"), {
  value: () => [{ unresolved: "readonly key: string" }, ErrorOptions],
});
```

Invoking that thunk throws `ReferenceError: ErrorOptions is not defined`.

## Why it is latent

The record is a thunk and nothing ever invokes it: the container only asks for one when it is resolving the class, and nobody injects an `Error` subclass. So this ships without symptom.

It is worth fixing anyway, because the failure mode when it does fire is a `ReferenceError` from inside the DI record rather than the boot error naming the parameter that dunx promises - and `key: string` right beside it is recorded correctly as `{ unresolved: ... }`, which is what `ErrorOptions` should get too.

## Expected

A parameter whose annotation is a type-only name (an interface, a lib type, a `import type` binding) records as `{ unresolved: "<the annotation>" }`, the same as a primitive does.

## Scope

`ErrorOptions` is the instance in the tree today. Any class in `packages/*` with a constructor parameter typed by a lib interface has the same shape.

petarzarkov:
Closes #142.

A constructor parameter whose type is a name the file neither imports nor declares is **ambient** - a global. `@dunx/transform` reads syntax and has no type checker, so `ErrorOptions` and `URL` look identical to it: a lib interface that erases, and a lib class that is a perfectly good token. Both were emitted verbatim, so the first produced a thunk that throws `ReferenceError: ErrorOptions is not defined` - neither the injection nor the boot error naming the parameter that dunx promises.

`typeof` settles it at resolution time rather than at transform time:

```js
typeof ErrorOptions === 'undefined' ? { unresolved: "options?: ErrorOptions" } : ErrorOptions
```

An ambient **type** falls back to the same `unresolved` descriptor a primitive already gets; an ambient **value** still resolves as the token it is. The guard is on the leftmost name, because `typeof` only protects a bare identifier - `typeof ns.Thing` still throws when `ns` is undeclared.

**Only ambient names are guarded.** `boundNames()` collects what the file imports and declares at the top level; anything in there is known to exist and is emitted exactly as before, so the forward-reference and circular-import cases the thunk exists for are untouched.

## Where it came from

`publish-guard` found it in the 3.8.2 tarballs, in `@dunx/infra`'s `FileNotFoundError(key: string, options?: ErrorOptions)` - note `key: string` right beside it was correctly recorded as `{ unresolved: ... }`. Latent, because nobody injects an `Error` subclass so the thunk was never invoked, and byte-identical back through 3.8.1, so no release introduced it.

Before and after on the built `@dunx/infra`:

```
before: ReferenceError: ErrorOptions is not defined
after:  [{"unresolved":"readonly key: string"},{"unresolved":"options?: ErrorOptions"}]
```

## Tests

Five cases in `deps.test.ts`: a lib type that erases, a global that is a real value (asserted by `eval`ing the emitted entry and checking it is `URL`), a qualified ambient name, and the two that must not change - an imported name and a class declared later in the same file. Three of the five fail without the fix.

`bun run ci`: 7 of 9 locally. The two failures are this machine, not the branch - `examples/tests` passes 5/5 when run alone and fails only under the full run's concurrent AMQP load, and `docs/test` fails identically on a clean `main` with the change stashed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of ambient and erased type names during transformation.
  - Prevented unresolved ambient types from causing runtime reference errors.
  - Preserved support for runtime globals and imported or locally declared names.
  - Improved resolution of qualified ambient type names.


petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

packages/transform/src/erased.ts
```diff
@@ -77,6 +77,49 @@ export const collectTypeOnlyNames = (
...
+ */
+export const boundNames = (program: Node): ReadonlySet<string> => {
+  const names = new Set<string>();
+
+  const bind = (node: Node | undefined | null): void => {
+    const name = nameOf(node);
+    if (name !== undefined) names.add(name);
+  };
+
+  walk(program, (node) => {
```
dunxonu: ⚠️ `correctness` The `boundNames` function uses `walk(program, ...)` to find all `VariableDeclarator` nodes in the file. This collects variables declared inside nested scopes (such as functions or blocks), which are not in scope at the top level where the dependency thunk is evaluated.

**How it fails:** For a file containing:
```ts
export class Service {
  constructor(private readonly repo: Repo) {}
}
function helper() {
  const Repo = 123;
}
```
`Repo` is only bound inside `helper` and is not in scope for `Service`. However, `boundNames` collects `Repo` because of the nested `VariableDeclarator`. As a result, `entryFor` emits `Repo` verbatim instead of guarding it, causing a `ReferenceError: Repo is not defined` when the dependency thunk is evaluated.

```suggestion
Instead of walking the entire AST for `VariableDeclarator` nodes, only collect them from top-level `VariableDeclaration` statements in the `program.body` loop.
```

dunxonu (commented):

The change introduces a guard for ambient types to prevent ReferenceErrors, but the scope detection for bound names incorrectly collects nested variable declarations, which can still lead to ReferenceErrors.

petarzarkov:
Added 630146e, the polling-test fix.

`settle` in `examples/full/src/messaging.test.ts` polls for 10 s, and the three tests calling it declared no timeout, so they ran on bun's default 5 s. The runner killed them first: the second half of the poll budget was unreachable, and a slow broker reported `timed out after 5000ms` rather than the assertion saying what had and had not been delivered. They declare `POLLING` now, which is longer than `settle`'s own deadline.

**This turned out to be the local CI failure I had been attributing to broker contention.** The contention is real - a full `bun run ci` puts a second app on the same RabbitMQ - but it only made delivery slow. The missing timeout is why slow became failed. With the fix, `bun run ci` goes from `examples/tests` failing on every full run to 12 of 13, and the one left is `docs/test`, which fails identically on a clean `main` with this branch stashed.

So the PR body's claim that "`examples/tests` passes 5/5 when run alone and fails only under the full run's concurrent AMQP load" is now stale - it passes under the full run too.

dunxonu (approved):

The change correctly guards ambient type annotations with a runtime `typeof` check to prevent ReferenceErrors, while preserving existing resolution behavior for imported or locally declared names.

petarzarkov:
All three findings taken in 59af295. Every one was real, and all three were the same defect: a name `boundNames` accepts is a name `entryFor` does not guard, which is the `ReferenceError` this branch exists to remove.

**@dunxonu - a nested binding was collected.** Right. `VariableDeclarator` was gathered with a `walk`, so a `const Repo` inside a helper function suppressed the guard for a top-level annotation naming `Repo`. Top-level statements only now, which is also where the class and function cases already were.

**@coderabbitai - a `declare` was collected.** Right, and it is exactly the ambient case: `declare class`, `declare const` and `declare function` are type-space promises about a runtime value the file does not create. Skipped on `declare === true`, with a test per form.

**@coderabbitai - a qualified name was guarded on its root alone.** Right, and the consequence is the worse one: `typeof Ns` passing says nothing about `Ns.Thing`, an absent member is `undefined`, and `isUnresolved` rejects `undefined` - so the container takes it as a token and skips the boot error instead of raising it. It now emits `typeof Ns === 'undefined' || Ns.Thing === undefined ? ... : Ns.Thing`, and the test evals the entry against an `Ns` that has one member and not the other.

44 tests in `deps.test.ts`, four of them new, each failing without its fix.

---

**A correction to my previous comment.** I said `examples/tests` "passes under the full run too" after 630146e. That is not reliably true, and I overstated what the timeout fix does. It failed again on a later full run - at 10008 ms now rather than 5000 ms, which is the fix working as intended: `settle` spends its whole poll window and the **assertion** fails, saying what was and was not delivered, instead of bun's clock cutting it short and reporting a timeout. The contention is real; the fix makes it legible rather than making it go away. GitHub CI passes on a clean runner, here and on every earlier run of this branch.

packages/transform/src/deps.ts
```diff
@@ -90,7 +92,27 @@ const entryFor = (
...
+     *
+     * `typeof` settles it at resolution time, on the **leftmost** name because
+     * it only protects a bare identifier. A qualified name needs the second
+     * half too: an absent member is `undefined`, which `isUnresolved` rejects,
+     * so the container would take it as a token rather than raise.
+     */
+    const missing =
+      named === root
+        ? `typeof ${root} === 'undefined'`
+        : `typeof ${root} === 'undefined' || ${named} === undefined`;
```
dunxonu, line 112: ⚠️ `correctness` If a qualified ambient type has multiple levels of nesting (such as Ns.nested.Type) and the top-level namespace Ns exists but the nested namespace does not, evaluating the thunk will throw a TypeError instead of falling back to the unresolved descriptor.

**How it fails:** For a parameter typed as Ns.nested.Type where Ns is defined as {} at runtime, evaluating the dependency thunk evaluates Ns.nested.Type === undefined, throwing "TypeError: Cannot read properties of undefined (reading 'Type')".

dunxonu (commented):

Fix reference errors for ambient types by guarding them with typeof checks at runtime, while preserving runtime globals and imported/declared names.

dunxonu:
The three fixes in 59af295 address the exact gaps that would have let the `ReferenceError` return:

1. **Nested bindings** – `boundNames` now iterates only top-level statements, so a `const Repo` inside a helper no longer suppresses the guard for a constructor parameter named `Repo`.

2. **`declare` bindings** – skipped because `declare class/const/function` are type-space only; they create no runtime value, so they must be treated as ambient.

3. **Qualified ambient names** – the guard now checks both the root and the member (`typeof Ns === 'undefined' || Ns.Thing === undefined`), preventing an absent member from being misread as a valid token.

All four new tests in `deps.test.ts` fail without these changes and pass with them. The transform now correctly distinguishes imported/declared names (emitted verbatim) from ambient names (guarded with `typeof` and falling back to an unresolved descriptor).

dunxonu (approved):

Fix runtime ReferenceErrors for ambient types by guarding them with typeof checks at resolution time.

Issue #148, opened 2026-09-13
http/health: ProbeResult.detail is a string, so structured per-check facts become prose


petarzarkov:
`ProbeResult` carries `state` and `detail?: string`. The registry's doc argues the four-bucket partition an app usually reaches for is worse, and I agree with that - this is not a request to bring `info` / `degraded` back.

It is that `detail` being a string turns structured facts into prose that nothing downstream can read.

## What it costs

Migrating a hand-rolled controller, three checks had numbers worth keeping:

| check | had | becomes |
| --- | --- | --- |
| queue | `{ notifications: { waiting: 12, active: 0, failed: 0 }, media: {...} }` | `"notifications 12w/0a/0f, media 1w/0a/500f"` |
| memory | `{ used: 143000000, limit: 2147483648 }` | `"143 MiB of 2048 MiB"` (this one is the framework's own `MemoryIndicator`) |
| storage | `{ driver: 'local', root: '/app/data/uploads' }` | dropped |

The queue line is the clearest: `500f` is five hundred failed jobs, and the only thing that can act on it is a human reading the string. An alert rule, a dashboard panel or a scrape cannot, without parsing a format nothing promises to keep.

`MemoryIndicator` shipping the same shape suggests this is felt upstream too - it has real numbers and renders them into a sentence on the way out.

## Suggestion

An optional `data?: Readonly<Record<string, unknown>>` on `ProbeResult`, beside `detail` rather than instead of it. `detail` stays the one line for an operator reading the page; `data` is what a scrape reads. Nothing that exists has to change, and `HEALTH_REPORT_SCHEMA` gains one optional field.

If that is too open-ended, `Record<string, number | string | boolean>` would cover every case above and stay trivially serialisable.

Found while migrating `dunx-template` to `HealthModule`. The rest of that migration was a clear win - per-indicator timeouts, `unknown` distinct from `down`, and `critical: false` expressing exactly what the app's own `degraded` bucket had been for.

petarzarkov:
Shipped in 3.9.1. `ProbeResult.data` carries the values and `detail` keeps the sentence, which is the split the issue asked for.

Verified against a running `dunx-template` readiness probe:

```
database  up     {"roundTripMs": 4}
queue     up     {"notifications": {"waiting": 173, "active": 0, "failed": 0}, "media": {...}}
storage   up     {"roundTripMs": 4}
amqp      up     {"roundTripMs": 8}
[...]

Issue #147, opened 2026-09-13
infra/cache: no shipped way to degrade, so every consumer writes the same CacheStore wrapper


petarzarkov:
`Cache.wrap` lets a store failure through to the caller. For a library that is the right default - swallowing it hides a broken deployment, and I am not asking for it to change.

What is missing is a **shipped way to opt out of it**, because the app that wants one writes the same wrapper every time.

## The case

A cached value can, by definition, be computed again. So for a lot of apps an unreachable Redis should cost latency and nothing else:

```ts
// today, with RedisCacheStore and the server gone
await cache.wrap('k', () => compute());
// ERR_REDIS_CONNECTION_CLOSED: Max reconnection attempts reached
await cache.get('k');
// Connection has failed
```

Every route behind that 500s. An app whose contract is "every area whose service is absent degrades" cannot adopt `@dunx/infra/cache` without writing this first:

```ts
export class DegradingCacheStore extends CacheStore {
  constructor(private readonly inner: CacheStore, private readonly logger: Logger) { super(); }

  async get<V>(key: string): Promise<V | undefined> {
    try { return await this.inner.get<V>(key); }
    catch (error) { this.#degrade(error); return undefined; }
  }
  // set and del the same way

  #degrade(error: unknown): void {
    // only a connection error; a serialisation bug is the app's and still throws
    if (!isConnectionError(error)) throw error;
    // ...and once per outage, not once per request
  }
}
```

The `CacheStore` seam is genuinely the right place for it - once it is there, `wrap`'s single-flight, `TieredCacheStore` and `CacheMetrics` all work unchanged, because from above it is a store that misses a lot. That part is a compliment to the design. The point is that the seam being right does not stop every consumer writing the same forty lines into it.

Two details that are easy to get wrong and worth having decided once:

- **Only a connection error may degrade.** A serialisation failure or a bad command is the app's bug and has to keep throwing, or the cache silently stops working and nothing says so. `isConnectionError` from `@dunx/infra/redis` already exists for this.
- **Warn once per outage, not once per operation.** An unreachable cache is touched on every cached route and is otherwise the loudest thing in the log.

## Suggestion

Ship it: `DegradingCacheStore` in `@dunx/infra/cache`, wrapping any `CacheStore`, with the classifier and the once-per-outage log. Opt-in, so the throwing behaviour stays the default.

[...]

PR #149, opened 2026-09-13
feat(http,infra,testing): the three things the dunx-template migration hit

### LINKED ISSUE #146: http: drainDelayMs is paid by every test teardown, and the failure names nothing
`Readiness` implements `OnBeforeShutdown` and holds shutdown for `drainDelayMs`. That is the right behaviour in a deployment and the reason the module is worth adopting. It is also paid by **every** `createTestServer` / `createTestApp` teardown, and a suite closes one server per file.

## What it looks like

Adopting `HealthModule.forRootAsync({ ..., drainDelayMs: 5_000 })` in an app with ten spec files turned ten `afterAll`s into this:

```
(fail) (unnamed) [5000.91ms]
  ^ a beforeEach/afterEach hook timed out for this test.
```

Bun's default hook timeout is 5 s, so the drain alone exceeds it. The failure names no test and points at no line, and the cause is a value set in a module the suite never mentions - which is what made it cost real time to find rather than being obvious.

It is not only tests. `bun run dev` pays it on every restart too, which is the same five seconds between saving a file and the server coming back.

## What an app has to do

Notice the coupling, then thread an environment-sensitive value through config:

```ts
drainMs: vars.HEALTH_DRAIN_MS ?? (vars.APP_ENV === AppEnv.LOCAL ? 0 : 5_000),
```

That works, and every app adopting `HealthModule` will write some version of it - or ship a five-second drain in development and not understand why the dev loop got slow.

## Suggestions, in the order I would take them

1. **`@dunx/testing` defaults it to `0`.** The harness already knows it is a harness, and nothing in a suite benefits from a drain. This is the one that removes the trap rather than documenting it, and it costs consumers nothing.
2. **Skip the drain when there is nothing to drain.** `Readiness` could resolve immediately when the server has no open connections and no in-flight requests. A drain exists so a load balancer notices; with no peers there is nobody to notice.
3. Failing both, say it in `drainDelayMs`'s own doc comment - that a suite pays it per file, and that Bun's hook default is 5 s.

Found while migrating `dunx-template` from a hand-rolled health controller to `HealthModule`. Everything else about that migration was a straight win: per-indicator timeouts and `unknown` vs `down` are both things an app cannot reproduce.

petarzarkov:
Closes #146, #147, #148. One migration, three gaps, and they touch the same two areas.

## #146 - the drain is paid by every test teardown

`Readiness` holds shutdown for `drainDelayMs` so a load balancer notices the failing probe. A suite has no load balancer and closes one app per file, so an app that sets the five seconds a deployment wants pays it per teardown - past Bun's default hook timeout, as an unnamed hook naming no line.

`@dunx/testing` now overrides `ReadinessOptions` to zero. **`Readiness` itself is untouched**: the drain means in production exactly what it meant before. Two details:

- The override goes in ahead of the caller's, so a suite that is testing the drain passes its own and wins.
- It is safe in a graph with no `HealthModule`. An override for a **class** token nobody bound is registered lazily and never constructed, so the harness does not have to know whether health is in the graph.

`packages/testing/src/drain.test.ts` covers both, plus that readiness still flips - only the timer is gone.

## #147 - no shipped way to degrade

`DegradingCacheStore` wraps any `CacheStore` and answers misses while the backend is unreachable, so a cached route costs latency instead of a 500.

- **Only a connection error degrades.** A serialisation bug still throws, or the cache quietly stops working and nothing says so.
- **Warns once per outage**, not once per operation, and clears on the first success so the next outage says so too.
- **`probe()`** does one real read. `degraded` starts optimistic, so a fresh process would otherwise report a cache it has never touched as healthy.
- It sits **outside** `MeteredCacheStore`, so a swallowed failure is still counted as an error and the hit rate is not inflated by the outage it hid.

`CacheModule.forRoot(init, { degrade: true })` is the coarse opt-in. Wrapping the L2 by hand is better for a tier: `TieredCacheStore.set` awaits L2 before L1, so wrapping from outside loses the L1 promotion too.

## #148 - `detail` is prose

`ProbeResult.data` carries the same facts as values, beside `detail` rather than instead of it. The shipped indicators had real numbers and rendered them into sentences, so they now carry both: `memory` reports `rssBytes` and `maxRssBytes`, `disk` reports `totalBytes`, `freeBytes` and `usedFraction`, and the round trips report `roundTripMs`. It reaches `HEALTH_REPORT_SCHEMA` and the dashboard's own `ProbeReport` too.

## examples/full

Uses all three. The cache layer dropped the boot-time `redis.ping()` that chose one tier or two - it only ever looked once, so a valkey lost after boot still 500'd every miss - for a degrading L2. `CacheStoreIndicator` probes that tier and reports `data`.

## Verification

`bun run ci` green, all 13 steps.

## Summary by CodeRabbit

- **New Features**
  - Cache backends can now degrade gracefully: connection failures return cache misses instead of application errors, while preserving visibility into outages and recovery.
  - Health checks now expose structured, machine-readable metrics such as latency, memory, disk, ledger, and cache status.
  - Cache health probing is included in the full application example and generated templates.

- **Documentation**
  - Added guidance for degraded caching, health-check data, shutdown behavior, and test environments.

- **Testing**
  - Improved test helpers to avoid unnecessary shutdown delays while preserving configurable behavior.


petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

dunxonu (commented):

The change successfully addresses the three migration gaps by zeroing the shutdown drain delay in tests, introducing a degrading cache store wrapper to handle unreachable backends gracefully, and exposing structured probe data instead of just prose.

packages/infra/src/cache/degrading.ts
```diff
@@ -0,0 +1,128 @@
...
+  readonly #probeKey: string;
+  #down = false;
+
+  constructor(
+    readonly inner: CacheStore,
+    init: DegradingCacheInit = {},
+  ) {
+    super();
+    this.#logger = init.logger ?? new ConsoleLogger();
+    this.#degradable = init.degradable ?? isConnectionError;
```
dunxonu, line 48: DegradingCacheStore's default `degradable` predicate is hardcoded to Redis's connection-closed error code, even though the class and CacheModuleSettings.degrade are documented as a generic wrapper for 'any store'.

**How it fails:** An app writes a custom CacheStore backend (e.g. Postgres- or S3-backed) and enables `CacheModule.forRoot(init, { degrade: true })` expecting outage protection per the docs guide ('DegradingCacheStore is that, wrapping any store'). Because `init.degradable` defaults to `isConnectionError` (imported from `../redis/errors.js`, which only matches Bun's `ERR_REDIS_CONNECTION_CLOSED` code), every connection failure from the non-Redis backend still throws uncaught - the degrade feature silently does nothing for any store other than RedisCacheStore, with no error or warning to indicate the mismatch.

packages/infra/src/cache/module.ts
```diff
@@ -16,8 +17,24 @@ export interface CacheModuleSettings {
...
+   *
+   * Only a connection error degrades; a serialisation bug still throws. The one
+   * line per outage goes to core's `ConsoleLogger` - pass a
+   * `DegradingCacheStore` as `store` to send it somewhere else.
+   */
+  readonly degrade?: boolean;
 }
 
+/** `exactOptionalPropertyTypes`: absent is what `CacheOptions` reads as off. */
+const degrading = (
```
dunxonu, line 34: The new `CacheModuleSettings.degrade` boolean wiring (the `degrading()` helper and its two call sites in `forRoot`/`forRootAsync`) has no direct test coverage and is not exercised anywhere in examples/full, which instead wires DegradingCacheStore by hand via CacheL2.

**How it fails:** A future refactor breaks the `settings.degrade === true` check or the plumbing into `new CacheOptions(init, metrics, degrading(settings))` (e.g. swaps an argument order, or the ternary is inverted). `packages/infra/src/cache/module.test.ts` has no test asserting `CacheModule.forRoot(init, { degrade: true })` actually produces a DegradingCacheStore, `examples/full` never sets `degrade: true` (it only constructs `DegradingCacheStore` directly), and `degrading.test.ts` only exercises `CacheOptions` directly - so `bun run test` and the example suite both stay green while the public `degrade` option is silently broken.

examples/full/src/cache/catalog.demo.ts
```diff
@@ -25,11 +27,15 @@ export class CatalogDemo {
     const tiers =
       configured instanceof TieredCacheStore
```
dunxonu, line 29: The `configured instanceof TieredCacheStore ? ... : 'L1 memory only'` branch is now dead code: CacheLayer.forRootAsync always builds a TieredCacheStore(l1, l2.store) regardless of Redis reachability, since the boot-time ping/fallback-to-memory-only logic was removed in this same diff.

**How it fails:** An operator reads the demo's boot log expecting 'L1 memory only' to indicate Redis is down, but that message can never print anymore - the store is unconditionally a TieredCacheStore now, so the log always says 'L1 memory in front of L2 redis' even while Redis is unreachable (degradation is instead reported separately via the new 'L2 probe -> unreachable' line). The same dead branch is duplicated verbatim in tools/create-app/templates/features/cache/catalog.demo.ts.

dunxonu (commented):

packages/infra/src/cache/degrading.ts
```diff
@@ -0,0 +1,128 @@
...
+   * something.
+   */
+  get degraded(): boolean {
+    return this.#down;
+  }
+
+  /** One real read down the same path. A miss is a reachable backend. */
+  async probe(): Promise<boolean> {
+    try {
+      await this.inner.get(this.#probeKey);
```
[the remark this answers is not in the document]
petarzarkov: Fixed in 7c2bfde, and a real one. With `degrade: true` and `metrics: true` the degrading store wraps the metered one, so `probe()` recorded a miss on every health check - a fixed cadence, so it walked the hit rate down for the life of the process.

`DegradingCacheInit.probeStore` is what `probe()` reads now, and `CacheOptions` passes the store from under the meter when it composes the two. Defaults to `inner`, so a hand-built wrapper is unchanged. Covered by a test that probes five times and asserts `misses` and `total` are untouched while the backend still saw all five reads.

packages/infra/src/cache/degrading.test.ts
```diff
@@ -0,0 +1,216 @@
...
+  });
+});
+
+describe('a failure that is the app’s own', () => {
+  it('still throws, so a serialisation bug is not swallowed', async () => {
+    const { store } = degrading(new BadValueStore());
+
+    expect(store.get('k')).rejects.toThrow('circular structure');
+    expect(store.set('k', 'v', 1)).rejects.toThrow('circular structure');
+    expect(store.del('k')).rejects.toThrow('circular structure');
```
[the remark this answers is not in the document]
petarzarkov: Fixed in 7c2bfde - all four occurrences now awaited. A regression that stopped rethrowing a non-connection error would have slipped past these, which is exactly the case they exist for.

packages/testing/README.md
```diff
@@ -54,6 +54,10 @@ The [Testing guide](../../docs/guide/11-testing.md) is canonical.
   never constructed: its `useFactory` never runs and its `onInit` never fires.
   That makes overriding a database safe.
 - Request logging and boot logging are off unless asked for.
+- `HealthModule`'s shutdown drain is off: `ReadinessOptions` is overridden to
+  `drainDelayMs: 0`, because a suite has no load balancer to notice a failing
+  probe and would pay the app's production value once per file. It goes in ahead
+  of yours, so passing your own restores it.
```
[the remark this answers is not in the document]
petarzarkov: Fixed in 7c2bfde. The two rules did read as a contradiction. The README now states the exception where it states the rule: naming a token nobody binds is an error **unless it is a class**, which self-binds, so an override for one is registered lazily and costs a graph that never asks for it nothing. That is what the harness relies on to apply `ReadinessOptions` without knowing your graph.

packages/infra/src/cache/degrading.ts
petarzarkov: Half right, and fixed as a documentation bug in 7c2bfde.

`isConnectionError` matching Bun's Redis code alone is the correct default - what counts as a connection failure for an arbitrary backend is not knowable from here, and guessing would swallow errors that should throw. The `degradable` option is the seam for it.

What was wrong is that the class doc, `CacheModuleSettings.degrade` and the guide all said "any store" without saying that. All three now do: another backend passes its own `degradable`, or nothing degrades and every failure still throws. Note it is not silent in the swallowing sense - the error still reaches the caller; the trap was expecting protection and reading nothing that said otherwise.

packages/infra/src/cache/module.ts
petarzarkov: Correct, fixed in 7c2bfde. The flag was only ever exercised at `CacheOptions`, so an inverted ternary or a swapped argument would have left the public `degrade` option broken with every suite green.

Two tests now go through `CacheModule.forRoot` into a real container and assert both settings - that `degrade: true` yields a `DegradingCacheStore` whose read misses against a downed backend, and that without it the same store throws. Built with `AppFactory` from `@dunx/core` rather than `createTestApp`: `@dunx/infra` does not depend on `@dunx/testing` and should not gain a build-order edge to it for this.

examples/full/src/cache/catalog.demo.ts
petarzarkov: Correct, fixed in 7c2bfde. The tier is built unconditionally now, so that branch could never print.

The demo reports the real shape - `L1 memory in front of a degrading L2 redis` - with the constructor name as the else, so it stays honest if the wiring changes again rather than being a second dead string. Reachability is the `L2 probe ->` line, which is the thing that actually varies. `tour.test.ts` asserted the old either/or and now asserts the one string it can print; the template is synced.

packages/infra/src/cache/degrading.ts
dunxonu: Fair — the default being Redis-specific is the right call since there's no universal definition of "connection failure" across arbitrary backends. The `degradable` option is the correct seam for other stores.

The documentation fix in 7c2bfde addresses the actual bug (misleading "any store" claims). Finding resolved.

dunxonu (approved):

The change successfully addresses the three migration gaps by disabling the shutdown drain delay in tests, introducing a degrading cache wrapper to handle connection failures gracefully, and exposing structured metrics in health probes.

PR #153, opened 2026-09-14
fix(openapi): keep the explorer's own mount out of the document

### LINKED ISSUE #152: Exclude OpenAPI documentation endpoints from the generated OpenAPI document
## Exclude OpenAPI documentation endpoints from the generated OpenAPI document

Currently, the generated `openapi.json` includes the OpenAPI module's own endpoints in `paths`, for example:

```json
"paths": {
  "/api/docs": {
    "get": {
      "operationId": "OpenApiController_page"
    }
  },
  "/api/openapi.json": {
    "get": {
      "operationId": "OpenApiController_document"
    }
  }
}
```

I don't think these endpoints should be part of the generated OpenAPI document.

`/docs` and `/openapi.json` are infrastructure endpoints used to expose or render the API documentation itself. They are not part of the application's API surface and should therefore be excluded from `paths`.

Ideally, the generated document would contain only the application's API routes:

```json
"paths": {
  "/api/users": {
    "get": {}
  }
}
```

This is especially relevant because the OpenAPI document is describing the application's API, not the endpoints used to serve the documentation.

I think these internal OpenAPI endpoints should be filtered out when generating the document.

Image

petarzarkov:
`/docs` and `/openapi.json` were discovered like any controller's routes and landed in `paths`, so a generated client arrived with a `getDocs()` and a `getOpenapiJson()` describing nothing the schemas cover.

That was deliberate: `module.ts` argued they are routes and that hiding them would be the first lie in the file. But the asset wildcard was already `@ApiHidden()`, so the document was not that complete description either, just inconsistently. The document describes the API; the explorer serving it is not part of the API.

Both routes are now `@ApiHidden()`, joining the asset route. Publicness is unchanged and still asserted, read off the route table rather than off a document that no longer carries them.

Closes #152

## Summary by CodeRabbit

* **New Features**
  * OpenAPI-generated documentation now excludes the documentation module’s own routes, including the JSON endpoint, docs page, and assets.
  * Generated clients no longer include methods for accessing undocumented OpenAPI endpoints.
  * Application routes and their security settings remain represented in the generated document.

* **Documentation**
  * Updated guidance to clarify which routes are discovered but hidden from the OpenAPI document.


dunxonu (approved):

The explorer routes are marked with @ApiHidden() so they are omitted from the generated OpenAPI document while remaining discoverable and public in the HTTP route table.

petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

packages/openapi/src/controller.ts
```diff
@@ -40,6 +44,7 @@ const documentController = (mount: DocMount) => {
     // works with or without the @dunx/transform preload.
     protected readonly explorer = inject(OpenApiExplorer);
 
+    @ApiHidden()
```
dunxonu, line 47: `@ApiHidden()` is added to each of the three doc-mount routes individually instead of once at the class level, though the codebase's own established pattern (and `ApiHidden`'s own JSDoc, "Valid on a method or a class") supports class-level use.

**How it fails:** `packages/http/src/health/controller.ts:78` already solves the identical problem with `@ApiHidden() export class HiddenHealthController extends HealthController {}` - one annotation covering every inherited method via `mergeMeta(klass, marked)` in route discovery (`packages/http/src/route/discover.ts`). Here the PR instead decorates `document` (line 47), `page` (line 88) and `asset` (line 110) separately. If a fourth route is later added to either `OpenApiController` class (in `documentController` or `buildController`) and the author forgets `@ApiHidden()` on that one method, it silently leaks into the generated OpenAPI document with no compiler or test signal, whereas a class-level annotation on the base (which even propagates through the `extends base` subclass via the prototype chain) would have made this structurally impossible.

packages/openapi/src/module.test.ts
```diff
@@ -195,14 +188,34 @@ describe('a real server serving its own document', () => {
     ).toBeDefined();
   });
 
-  it('documents its own routes as public, so a global guard is not implied', async () => {
+  it('carries the app route own security into the document', async () => {
```
dunxonu, line 191: The deleted assertion `document.paths['/api/openapi.json']?.get?.security` (which exercised the real document-generation pipeline for the doc-mount route) is replaced by a test that only reads route metadata (`route.public`), never `generateDocument`/`operations.ts`'s security-rendering branch for these specific routes.

**How it fails:** Because hidden routes are now filtered out in `describeRoutes` before `generateDocument` ever sees them, the specific code path that used to render `security: []` for the doc-mount's own public route is no longer covered end-to-end by any test for that route; a future change to `operations.ts`'s security logic that mishandles a hidden-but-public route would not be caught here (the risk is narrowed because such routes can no longer reach that code at all post-change, but the shift from an end-to-end assertion to a metadata-only one is a real reduction in what this specific test proves).

dunxonu (commented):

PR #155, opened 2026-09-14
refactor(openapi): one class-level @ApiHidden over three method-level ones

petarzarkov:
Follow-up to the review on #153.

The three doc-mount routes each carried their own `@ApiHidden()`. A fourth route added to either `OpenApiController` later would have leaked into the generated document with no compiler or test signal.

`metaOf` is a plain property read, so a class-level mark reaches the subclass `buildController` builds through the prototype chain, and `@Controller()` on that subclass does not shadow it. I verified that before relying on it. One mark on the base now covers `document`, `page` and `asset`, plus whatever either class grows later. It is the pattern `HiddenHealthController` already uses.

No behaviour change: the openapi suite passes unmodified, including `mounts its own routes public and undocumented`, which enumerates every `OpenApiController` route off the route table and asserts each is hidden. That test is the regression guard either way.

The reviewer's second finding, about losing end-to-end coverage of `security: []` for the doc-mount route, I did not act on. That branch is still covered end-to-end by `generate.test.ts` on `/reports/health`, an ordinary `@Public()` app route, and the doc-mount routes cannot reach `generateDocument` at all now, so there is no end-to-end assertion left to make about them.

## Summary by CodeRabbit

* **Documentation**
  * OpenAPI documentation now hides the document controller and its routes from generated API references.
  * Public access and existing authorization behavior remain unchanged.


dunxonu (approved):

Consolidating the route-level ApiHidden decorators into a single class-level decorator on the base controller correctly hides all document mount routes.

petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

dunxonu (commented):
Reviewed PR 155 (`refactor(openapi): one class-level @ApiHidden over three method-level ones`), which touches only `packages/openapi/src/controller.ts` (10 lines: moves `@ApiHidden()` from three method decorators to one class decorator on `documentController`'s `OpenApiController`, relying on it being inherited by the `buildController` subclass).

I traced the metadata mechanism end to end to verify the PR's own claim:
- `packages/http/src/route/metadata.ts`: `write()` sets `META` as an **own** property via `Object.defineProperty`; `metaOf()`/`mergeMeta()` read `target[META]` via a **plain property lookup**, which walks the constructor's static prototype chain for a class target.
- `packages/http/src/route/decorators.ts`: `Controller()` calls `markController()`, which writes to a separate `CONTROLLER` symbol, never touching `META`. So decorating the subclass with `@Controller()` cannot shadow the inherited `HIDDEN` entry.
- `class OpenApiController extends base` in `buildController` (packages/openapi/src/controller.ts:87) sets `Object.getPrototypeOf(subclass) === base` at the constructor level, so `metaOf(subclass)` correctly falls through to `base`'s own `META` map, which now carries `HIDDEN: true` for `document`, `page`, and `asset` alike.
- Consumers (`packages/http/src/inspect.ts:91`, `packages/openapi/src/discover.ts:25`) both read the **merged** meta (`mergeMeta(klass, marked)`), so the inherited class-level flag reaches the same place the three deleted method-level flags used to.
- The existing regression test `packages/openapi/src/module.test.ts:204` ("mounts its own routes public and undocumented") exercises exactly the renderer-present path (subclass) and asserts `hidden` is `true` for every `OpenApiController` route; the no-renderer path is separately covered by the "with no renderer" describe block asserting the document route is absent from generated paths.

I could not run `bun test`/`bun run ci` in this sandbox (bash execution was blocked by the approval system), so I relied on static tracing of the exact code paths rather than execution, but the mechanism is unambiguous from the source.

No correctness bug, removed-behavior regression, cross-file breakage, reuse/simplification/efficiency issue, altitude problem, or CLAUDE.md convention violation survived scrutiny. This diff is itself a legitimate simplification (one declaration at the lowest common owner, matching Rule 2) with no observable behavior change.

Findings:

```json
[]
```

PR #156, opened 2026-09-14
feat(core): a .ts or .js config file, read from its default export

### LINKED ISSUE #150: Support TypeScript configuration files
Thanks for adding support for YAML, JSON, and other configuration formats in #79.

I would also like to suggest supporting `.ts` configuration files.

The main benefit is **first-class TypeScript type safety and IDE support**. Since Dunx is a TypeScript/Bun framework, users could get autocomplete, type checking, and inline documentation directly from the framework's configuration types:

```ts
export default {
  server: {
    port: 3000,
    http2: true,
  },
  logger: {
    level: 'info',
  },
}
```

For example, an invalid configuration such as:

```ts
logger: {
  level: 'verbose',
}
```

could be caught by TypeScript immediately, while YAML/JSON would require a separate schema to provide the same developer experience.

It would also allow configuration to reference environment variables, constants, or other TypeScript modules naturally.

Since Bun can load TypeScript modules natively, I think `.ts` would be a natural configuration format for Dunx alongside YAML and JSON.

petarzarkov:
Closes #150.

`ConfigModule.forRoot({ files })` took YAML, TOML and JSON. It now also takes `.ts` and `.js`, imported rather than parsed, which Bun does natively so it costs no dependency.

## Not for the reason the issue opened with

The issue asked for `.ts` on the grounds that it would give type safety for free. It would not. dunx has no framework-wide config type: the shape is whatever your `validate` or `schema` produces, so a bare object literal is checked against nothing, and `satisfies AppConfig` is what makes tsc say anything. That already works today with no framework support, by importing the module and handing it over yourself.

The reporter agreed on the issue. What `files` actually adds is the **layering**: a code-shaped file takes its turn in the same ordered deep merge as the yaml, under the environment, rather than being spread in by hand.

## Shape

- `.ts`, `.js`, `.mts`, `.mjs`, `.cts`, `.cjs` are imported; everything else parses as before.
- **Default export only**, so there is one answer to which export is the config.
- A module with no default export fails boot naming the file, rather than being skipped: it is almost always `export const config = ...` with the keyword missing. An empty `.yml` stays skipped, which is what keeps an overlay optional.
- An import that throws is wrapped as `did not load` naming the file, the way a parse error is wrapped as `did not parse`.
- `__proto__` stripping and the top-level-object check apply unchanged.

Both answers here (`.ts` + `.js`, default export only) are what the reporter asked for when I put the two questions to them.

## Rule 4

`examples/full` gains `application.config.ts` as its last layer. It composes `EMAIL_FROM` out of the app name, which the schema default used to repeat, so `appName` now has one source instead of two.

## Summary by CodeRabbit

- **New Features**
  - Configuration now supports TypeScript and JavaScript files, including `.ts`, `.js`, `.mts`, `.mjs`, `.cts`, and `.cjs`.
  - These files can compose configuration values, import modules, access the environment, and export configuration through a default export.
  - The full application example now derives its application name and email settings from a TypeScript configuration file.
  - Clear errors are provided for missing default exports and invalid configuration modules.

- **Documentation**
  - Configuration guidance now explains supported module formats, execution behavior, default exports, and type-checking with `satisfies ConfigFile`.


packages/core/src/config/files.ts
```diff
@@ -51,10 +51,45 @@ const parserFor = (path: string): ((text: string) => unknown) => {
...
+ * nothing to guess about which export was meant. Unlike every other format this
+ * one **runs**, so it can read `Bun.env` and import (#150).
+ */
+const importModule = async (
+  path: string,
+  absolute: string,
+): Promise<unknown> => {
+  let module: { default?: unknown };
+  try {
+    module = (await import(absolute)) as { default?: unknown };
```
dunxonu: ⚠️ `correctness` Passing a raw absolute path to import() can fail if the path contains characters like # or ? (which are interpreted as URL fragments or queries), or on Windows due to backslash path separators. Converting the path to a file URL using pathToFileURL from node:url ensures it is safely imported.

**How it fails:** If the application is located in a directory containing a # character (for example, /home/user/projects/c#_app/), import(absolute) will fail because the loader treats everything after # as a fragment.

```suggestion
module = (await import(pathToFileURL(absolute).href)) as { default?: unknown };
```

dunxonu (commented):

Add support for importing .ts and .js configuration files via their default export.

petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

examples/full/application.config.ts
```diff
@@ -0,0 +1,9 @@
+import type { ConfigFile } from './src/config.js';
```
dunxonu, line 1: The new config file is invisible to tsc, so the PR's headline type-safety story (`satisfies ConfigFile`) is never actually checked by CI.

**How it fails:** examples/full/tsconfig.json declares `"include": ["src", "scripts", "services.ts"]`; `application.config.ts` sits at the example root, outside all three, and nothing under `src`/`scripts` imports it (only the reverse: it imports the `ConfigFile` type from `src/config.js`). `tsc --noEmit` (the repo's `typecheck` script, run by `bun run ci`) therefore never opens this file. If a future edit removes a required key or breaks the `satisfies ConfigFile` check, `bun run ci`'s tsc gate stays green; the break would only surface as a runtime zod error if some test happens to boot the app, contradicting the PR description's claim that 'satisfies... is what makes tsc say it first.'

docs/guide/12-configuration.md
```diff
@@ -99,6 +101,37 @@ seed:
...
+was meant. A file with no default export fails boot naming the file, rather than
+being skipped: it is almost always `export const config = ...` with the keyword
+missing.
+
+Unlike the other formats this one runs, so it can compose values, read `Bun.env`
+and import other modules:
+
+```ts
+// application.config.ts
+import type { ConfigFile } from './src/config.ts';
```
dunxonu: The new doc example imports with a `.ts` extension, which this repo's own `moduleResolution: nodenext` forbids.

**How it fails:** The snippet reads `import type { ConfigFile } from './src/config.ts';`, but the real file it's modeled on, `examples/full/application.config.ts:1`, correctly uses `'./src/config.js'` - and `examples/full/tsconfig.json` even comments that relative imports 'keep the mandatory .js extension like every other file here.' A reader who copies the doc's exact line into a real `nodenext` project gets a `tsc` resolution error.

packages/core/src/config/files.ts
dunxonu: A `.ts`/`.js` config file's dynamic `import()` is cached per resolved path for the life of the process, so a config file that reads `Bun.env` (explicitly advertised as supported) only reflects the environment at its first load.

**How it fails:** `docs/guide/12-configuration.md` and `files.ts:65` both advertise that, unlike the parsed formats, a `.ts`/`.js` file 'can compose values, read `Bun.env` and import other modules.' If an app (or, e.g., a retry path, or `bun test` running many spec files in one process that each call `configModule(source)`/`AppFactory.create` against the same physical config file) calls `ConfigModule.forRoot({ files })` a second time for the same absolute path after `Bun.env` changed, the second `importModule` call gets the already-cached module and silently returns the first call's stale `Bun.env`-derived values rather than re-evaluating - with no cache-busting like the `bust`/reload query trick used elsewhere in the codebase (`packages/infra/src/email/templates.ts:86`).

packages/core/src/config/files.ts
```diff
@@ -51,10 +51,45 @@ const parserFor = (path: string): ((text: string) => unknown) => {
...
+    throw new ConfigError(
+      `Config file "${path}" did not load: ${
+        cause instanceof Error ? cause.message : String(cause)
+      }`,
+    );
+  }
+
+  // Not skipped the way an empty `.yml` is: no default export is almost always
+  // `export const config = {...}` with the keyword missing.
+  if (!('default' in module)) {
```
dunxonu: `export default undefined;` silently passes as `default in module` yet the resulting `undefined` is then treated as a skip, quietly bypassing the documented 'no default export fails boot' guarantee.

**How it fails:** A `.ts` config file does `const cfg = someCondition ? {...} : undefined; export default cfg;`. When the condition is false, `'default' in module` is true (the binding exists, valued `undefined`), so `importModule` returns `undefined` instead of throwing 'has no default export'. Back in `load()` (`files.ts:135`), `parsed === undefined` sends it down the same path as an absent/empty file - i.e. silently skipped - even though the file exists, was imported, and produced no config, which is a third, undocumented state distinct from both 'missing default export' (documented to hard-fail) and 'file absent' (documented to skip).

dunxonu (commented):

petarzarkov:
Pushed 47f3384. Six of the seven acted on, one not, and one of the six was worse
than the review made it sound.

**The `satisfies` was never checked.** `examples/full/tsconfig.json` had
`include: ["src", "scripts", "services.ts"]`, and `application.config.ts` sits at
the example root, so tsc never opened the one file the example exists to
demonstrate. Widened the include, then verified it properly by breaking the file:

```
application.config.ts(7,3): error TS2322: Type 'number' is not assignable to type 'string'.
```

It is clean again now. Good catch, and it undercut the PR description as written.

**`export default undefined` was a third state.** Both reviewers found this
independently and it reproduces exactly: `'default' in module` is true for it, so
`importModule` returned `undefined` and `load()` skipped the file like an absent
one. A file that exists and resolves to nothing is a mistake, where an absent
overlay is a choice, so it now fails boot naming the file. `null` too. Test
added.

**The doc snippet** imported `./src/config.ts` where nodenext wants `.js`. Fixed.

**`.mts`, `.mjs`, `.cts`, `.cjs`** were matched by the regex and documented
nowhere. Rather than document them, I narrowed the regex to `.ts` and `.js`,
which is what the issue asked for. The rest fall through to the existing "no
parser" error, which names what is supported. Test pins that.

**The module cache** is real. I documented it rather than busting it: config is
loaded once per boot by design, and `source` already exists as the supported way
to vary an environment per boot, spread over the file values so it wins. Busting
in the production load path to serve a test-harness case, and re-running the
module's top level each boot, is the worse trade. Happy to be argued out of it.

**Not acted on: `pathToFileURL` fixing a `#` in the path.** The stated failure
does not reproduce. On Bun 1.4.2:

```
raw absolute with #: {"ok":1}
pathToFileURL with #: {"ok":1}
raw with ? FAILED: Cannot find module ...
fileURL with ? FAILED: Cannot find module ...
```

`#` already works, and `?` is broken with or without the conversion, so the fix
does not do what the finding says it does. I took the change anyway, because a
file URL is the specifier grammar `import()` actually takes and the Windows
backslash half is not something I can disprove here, but it is an idiom change
rather than the bug fix it was filed as. There is now a test with a `#` in the
path so the claim stays pinned either way.

packages/core/src/config/files.test.ts
```diff
@@ -302,3 +302,117 @@ describe('ConfigModule with files', () => {
...
+  it('runs, so it can read the environment and import', async () => {
+    const path = await write(
+      'env.config.ts',
+      'import { join } from "node:path";\n' +
+        'export default { name: Bun.env.DUNX_TEST_NAME, joined: join("a", "b") };\n',
+    );
+
+    Bun.env['DUNX_TEST_NAME'] = 'from-env';
+    try {
+      expect(await load(path)).toEqual({ name: 'from-env', joined: 'a/b' });
```
dunxonu, line 328: ⚠️ `correctness` The test writes a config file that uses `join('a', 'b')` from `node:path` and asserts that the result is exactly `'a/b'`. On Windows, `join` uses `\` as the path separator, causing the assertion to fail.

**How it fails:** Running the test suite on a Windows machine where `join('a', 'b')` evaluates to `'a\b'`, causing the assertion to fail.

```suggestion
expect(await load(path)).toEqual({ name: 'from-env', joined: join('a', 'b') });
```

dunxonu (commented):

Add support for loading configuration from `.ts` and `.js` files via their default exports, but the test suite contains a platform-specific path separator assumption that will fail on Windows.

dunxonu (approved):

The change successfully adds support for importing TypeScript and JavaScript configuration files via their default exports, integrating them into the existing deep-merge configuration pipeline.

PR #158, opened 2026-09-15
fix(mcp): hold callers to the tool schema, and answer with structuredContent

petarzarkov:
Two unrelated changes, one commit each.

**`fix(mcp)`.** Every tool published `additionalProperties: false` and nothing enforced it, so a misspelled key was indistinguishable from an absent one. `dunx_guide` takes `topic` and fourteen of the server's descriptions call the thing a chapter, so `{ chapter: '06-validation' }` fell through to the no-argument branch and returned the whole 17 KB index with nothing to say it had happened. That call is now 144 bytes naming `topic, search`. Enforced in `handle` so all nine tools get it at once; a tool declaring no arguments stays permissive, because a stray key cannot mislead an answer that never depended on one.

Tool results also carry `structuredContent` now, beside the text block. Measured through Claude Code against `examples/full`, one `dunx_gateways` call reaching the model is 598 bytes with it and 1,354 without: the client prefers the structured field and serialises it flat, where the text block is indented for a human.

**`docs(notes)`.** A spec for a sixth bench family measuring how often an agent authors a working app, anchored on no framework at all rather than on Nest. Carries the arithmetic that decides whether it gets built, and a 30-trial pilot with a kill gate. No code.

Local: `static` and `unit` green, 104 MCP tests pass. `examples/tests` fails on `tour.test.ts` "exits 0 with no redis at all", which is the bullmq Leak B in `internal/notes/roadmap/queue-shutdown-sigterm.md` and needs local Redis and AMQP this box does not have.

## Summary by CodeRabbit

- **New Features**
  - MCP tool calls now reject undeclared arguments, incorrectly typed values, and non-object arguments with clear errors before execution.
  - Successful tools returning objects now provide structured results alongside readable text output for compatible MCP clients.

- **Documentation**
  - Added guidance on argument validation and structured result delivery, including practical examples and payload-size comparisons.
  - Updated the roadmap with a planned benchmark for measuring agent-generated application success.


dunxonu (commented):

The change correctly implements tool argument validation and structured content in tool results, but the test file exceeds the maximum allowed line count.

- `tools/mcp/src/protocol.test.ts:572` Test file exceeds 500 lines

</details>

petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

dunxonu (commented):

Enforce tool schema validation and support structuredContent in MCP tool results, but the test file exceeds the 500-line limit.

- `tools/mcp/src/protocol.test.ts:583` Test file exceeds the 500-line limit

</details>

dunxonu (commented):

The changes add robust tool argument validation and structured content support to the MCP protocol implementation, but the test file now exceeds the 500-line limit.

- `tools/mcp/src/protocol.test.ts:698` File exceeds the 500-line limit

</details>

petarzarkov:
@coderabbitai review

dunxonu (approved):

Enforces tool schemas against undeclared or mistyped arguments and attaches structuredContent to tool responses.

petarzarkov:
@coderabbitai review

petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

tools/mcp/src/protocol.ts
```diff
@@ -74,6 +74,74 @@ const withoutFragment = (uri: string): string => uri.split('#')[0] ?? uri;
...
+    // `null` is absent, not a type error: JSON has no `undefined`, so a client
+    // serialising an omitted optional filter sends `null` and means "no filter".
+    // An undeclared key is the opposite case and is still refused above.
+    if (value === undefined || value === null) continue;
+    const declaredType = isRecord(properties[key])
+      ? properties[key]['type']
+      : undefined;
+    if (typeof declaredType !== 'string') continue;
+    if (!satisfies(declaredType, value)) {
+      return `Argument \`${key}\` must be a ${declaredType}, received ${typeName(value)}.`;
```
dunxonu: The type-mismatch error message hardcodes the article "a", producing ungrammatical text for declared types starting with a vowel sound.

**How it fails:** A tool declares `{ type: 'integer' }` (as the PR's own WITH_INTEGER test fixture does) and is called with `{ limit: 2.5 }` or `{ limit: '25' }`; the agent-facing reply reads "Argument `limit` must be a integer, received number." Same for any future `type: 'array'` or `type: 'object'` property. The existing test only asserts `.toContain('integer')` so this ships uncaught, undercutting the PR's stated goal of giving the model a clean, actionable error string.

tools/mcp/src/protocol.ts
```diff
@@ -224,14 +292,41 @@ export const handle = async (
...
+        return reply(call.id, {
+          isError: true,
+          content: [{ type: 'text', text: complaint }],
+        });
+      }
+      // `misuse` reads `null` as absent, so `run` is handed a record where it is
+      // absent. Leaving the key on relied on every `run` repeating that reading.
+      const output = await tool.run(
+        Object.fromEntries(
+          Object.entries(args).filter(([, value]) => value !== null),
```
dunxonu: Only `null`-valued keys are stripped before calling `tool.run`; an explicit `undefined`-valued key that `misuse` treats as absent still survives as a present key on the object handed to `run`, contradicting the adjacent comment's stated invariant.

**How it fails:** Calling the exported `handle()` directly (it and `ToolDefinition` are re-exported from tools/mcp/src/index.ts precisely so a third party can supply its own tools) with `arguments: { topic: undefined }`: `misuse` reads it as absent and raises no complaint (matching intent), but `Object.keys()`/`Object.entries()` inside a custom `run` would still see `topic` as a present key with value `undefined` - the same 'present but should read as absent' class of bug the PR set out to eliminate for `null`, left open for `undefined`. Unreachable over the real stdio/JSON-RPC wire since `JSON.parse` never produces `undefined`, but reachable by any embedder calling `handle` in-process.

tools/mcp/src/protocol.ts
```diff
@@ -74,6 +74,74 @@ const withoutFragment = (uri: string): string => uri.split('#')[0] ?? uri;
...
+ * Only a schema that bans extras gets them rejected. A tool declaring no properties
+ * cannot be misdirected by a stray key - its answer is the same either way - so
+ * `NO_ARGS` stays open on purpose and failing those calls would only cost an
+ * adopting agent its first one.
+ */
+const misuse = (
+  inputSchema: Record<string, unknown>,
+  args: Record<string, unknown>,
+): string | undefined => {
+  const properties = isRecord(inputSchema['properties'])
```
dunxonu, line 116: `misuse` recomputes `properties`/`declared` (an object-key walk) from a tool's `inputSchema` on every single `tools/call`, even though that schema is a frozen object built once at process start and never changes.

**How it fails:** An agent session that calls the same tool (e.g. `dunx_routes`) repeatedly re-derives the identical `declared` key list from the identical `inputSchema` on every call; the per-request cost is small but is pure repeated work that could be computed once per tool at registration instead.

tools/mcp/src/protocol.ts
```diff
@@ -224,14 +292,41 @@ export const handle = async (
...
-        unknown
-      >;
-      const output = await tool.run(args);
-      // Text content holding JSON, which is what a client can both show and parse.
+      // A number, a boolean or an array survives `?? {}` and reaches `run` as
+      // something its `Record<string, unknown>` contract never receives, where
+      // every key reads `undefined`: the silent miss `misuse` exists to stop.
+      const sent = call.params?.['arguments'];
+      if (sent !== undefined && sent !== null && !isRecord(sent)) {
+        return reply(call.id, {
```
dunxonu: The diff introduces two new occurrences of the exact `reply(call.id, { isError: true, content: [{ type: 'text', text: ... }] })` shape, bringing the total to three copies (with the pre-existing catch block at line 335) inside one ~45-line function, with no shared helper.

**How it fails:** A future change to how a tool error is reported (e.g. adding an error `code` field, or normalizing message casing) requires editing three near-identical sites in `handle` instead of one `toolError(id, text)` helper, making the three easy to update inconsistently.

tools/mcp/src/protocol.ts
```diff
@@ -74,6 +74,74 @@ const withoutFragment = (uri: string): string => uri.split('#')[0] ?? uri;
 const readableId = (value: unknown): JsonRpcRequest['id'] =>
   typeof value === 'string' || typeof value === 'number' ? value : null;
 
+const isRecord = (value: unknown): value is Record<string, unknown> =>
```
dunxonu, line 84: The new `isRecord` predicate duplicates an identical (or near-identical) one-line type guard already present unexported in `packages/openapi/src/refs.ts` and `packages/openapi/src/metadata.ts`, making it a third private copy of the same check.

**How it fails:** Not gated (CLAUDE.md marks Rule 2 as "Not gated"), but a later bug fix to the predicate (e.g. handling a boxed primitive) would need to be applied in three places since none of the existing copies are exported for reuse.

dunxonu (commented):
**Actionable comments posted: 6**

tools/mcp/src/protocol.test.ts
```diff
@@ -436,3 +436,297 @@ describe('resources', () => {
...
+        properties: { topic: { type: 'string' }, full: { type: 'boolean' } },
+        additionalProperties: false,
+      },
+      run: (args) => ({ topic: args['topic'] ?? null }),
+    },
+    {
+      name: 'open',
+      description: 'Declares no properties and no ban on extras.',
+      inputSchema: { type: 'object', properties: {} },
+      run: () => ({ ok: true }),
```
dunxonu, line 497: Four new test-local helpers (`call`, `ask`, `send`, and the pre-existing-shape `resultOf`) each rebuild the identical `{jsonrpc:'2.0', id:1, method:'tools/call', params:{name, arguments}}` envelope plus `JSON.parse(...).result`, instead of sharing one.

**How it fails:** A future change to the request envelope or reply-parsing shape (e.g. adding a header, changing the id scheme) has to be made in four near-duplicate helpers within the same file instead of one; `ask` at line ~585 also shadows an existing top-level `ask` defined earlier in the file, which is confusing to a reader scanning for the file's actual test helpers.

dunxonu (commented):

This change fixes argument validation for tool calls and adds structured content to tool results, while also adding new tests.

2 findings already have an open thread on the same line. Not repeated here.

- `tools/mcp/src/protocol.test.ts:1` Test file exceeds 500 lines limit

</details>

tools/mcp/src/protocol.ts
petarzarkov: Declined, deliberately. The schemas this walks hold at most four keys (`schema()` builds `string` and `boolean` only), and a session makes tens of `tools/call`s, not millions. A per-tool cache would add a WeakMap lookup and an invalidation question to save an `Object.keys` over a two-key object, which is more machinery than the work it removes. Reopen it if a tool ever declares a large schema.

tools/mcp/src/protocol.ts
petarzarkov: Confirmed the duplication: `packages/openapi/src/refs.ts:13` is byte-identical and `metadata.ts:31` is the same check without the array guard. Declining the move anyway. Rule 2 says the lowest common owner, and the common owner of `@dunx/mcp` and `@dunx/openapi` is `@dunx/core` - `@dunx/openapi` is an optional peer here, so this code path cannot import from it. That makes the fix a new public export on core for a one-line type guard, which is more surface than three private copies cost. Worth revisiting if a fourth copy appears or core grows a guards module for its own reasons.

tools/mcp/src/protocol.ts
dunxonu: Fair enough — the schemas are tiny and the call volume is low, so the repeated `Object.keys` is genuinely negligible. No need to add caching machinery for a problem that doesn't exist yet. Dropping it.

PR #159, opened 2026-09-15
Third-party packages: the diagnosis, the duplicate, the build and the guide

petarzarkov:
Four changes, all on the path someone takes to publish a dunx package to npm. The
first two are defects that surfaced in the consumer's app rather than in the
author's build.

**The missing-records diagnostic blamed the wrong person.** A package built
without `depsPlugin` ships classes with no dependency records, and the container
reported that as a missing preload, which the app hitting it already had. The
plugin now marks that it registered, so the message separates "no preload" from
"the plugin ran and never saw this file", and the second case names
`node_modules` and whose build has to change.

**Two copies of `@dunx/core` failed silently.** A token is a class, so the second
copy's `Logger` is unbound, self-binds, and is constructed into an object with no
methods. Nothing failed until something called one, in whichever package injected
it. `AppFactory.create` counts loaded copies and refuses to boot on more than
one. Copies published before this check do not register, so a mixed tree still
counts one.

**The build is published.** `scripts/build-package.ts` was the only correct way
to build a dunx package and was not on npm, so three plausible builds (`tsc`,
the `bun build` CLI, shipping `.ts`) all produce a broken package. It moves to
`@dunx/transform/build` as `buildPackage()`, and the script becomes a caller, so
the repo's twelve workspaces go through the entry an outside author calls.
`@dunx/infra`'s `dist/` is byte-identical across the move. `typescript` becomes
an optional peer, since the build spawns `bunx tsc`.

**A guide.** `docs/guide/30-publishing-a-package.md`, in a new "Extending dunx"
section. The lockstep decision in `docs/architecture/packaging.md` named
`@dunx/core@1.0.0` as the trigger to revisit it; at 3.x that has passed, so the
decision and the two paragraphs resting on it are rewritten.

Verified outside the repo against a package installed from `bun pm pack` output,
not a workspace link: a local `file:` install resolves to the real source
directory and hides the whole problem.

`bun run ci` 13/13.

## Summary by CodeRabbit

* **New Features**
  * Added a package build utility that bundles module entry points, generates declarations, validates package structure, and supports executable outputs.
  * Added safeguards that detect multiple loaded copies of the core package and stop startup with diagnostic guidance.

* **Bug Fixes**
  * Improved transformation error messages to distinguish missing setup from packages whose dependencies were not recorded.

* **Documentation**
  * Added publishing guidance covering package builds, dependencies, testing, and release preparation.
  * Updated architecture and navigation documentation for current versioning and publishing workflows.


petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

dunxonu (approved):

The changes introduce a guard against multiple copies of @dunx/core, improve the missing-transform diagnostic to correctly blame the package build when the plugin is registered, move the package build logic to a shared function in @dunx/transform/build, and update documentation. No new breaking changes are detected.

packages/core/src/di/transform-hint.ts
```diff
@@ -35,8 +52,18 @@ export const missingTransformMessage = (
       `${head} The entrypoint is ${entry}, so this is a prebuilt tree: the ` +
       'preload plugin only matches .ts, and no preload setting can change ' +
       'that. Record the dependencies at build time instead:\n\n' +
-      "  import { depsPlugin } from '@dunx/transform';\n" +
-      '  await Bun.build({ /* ... */ plugins: [depsPlugin] });\n'
+      BUILD
+    );
+  }
+
+  if (registered) {
```
dunxonu, line 59: The new 'registered but no record' branch always blames the class's own installed package, but the same code path is hit when an app's own class-expression (never transformed) trips the check.

**How it fails:** An app author writes `export const Foo = class Foo { constructor(private bar: Bar) {} }` in their own .ts source (not node_modules). `packages/transform/src/deps.ts:156` only records dependencies for `ClassDeclaration` nodes (per `isClassDeclaration`), never `ClassExpression` (CLAUDE.md: 'The transform only touches class declarations, never class expressions'), so no record is ever produced regardless of preload/build. With the preload correctly registered and the entrypoint a normal .ts file, `packages/core/src/di/injector.ts:278-279` calls `missingTransformMessage(ctor.name, ctor.length)`, which now takes the `registered=true` branch and tells the developer 'That is a fix for whoever publishes Foo, not for this app' - actively misdirecting them away from their own class-expression bug and toward blaming a third-party package that may not even exist.

packages/transform/src/build.ts
```diff
@@ -0,0 +1,216 @@
+import { chmod, rm } from 'node:fs/promises';
+import { join } from 'node:path';
+import { depsPlugin } from './plugin.js';
+
+/** The manifest fields a build reads. */
+interface PackageManifest {
```
dunxonu, line 16: New local `PackageManifest` interface duplicates the shape already declared as `Manifest`/`ExportEntry` in `scripts/workspace-ranges.ts`, which the code this file replaced imported directly.

**How it fails:** If `scripts/workspace-ranges.ts`'s `Manifest`/`ExportEntry` shape changes (e.g. a new exports-entry field), `packages/transform/src/build.ts`'s independent `PackageManifest` won't pick up the change, and the two representations of a package.json manifest can silently drift out of sync across the repo's own build tooling.

packages/core/src/di/copies.ts
```diff
@@ -0,0 +1,48 @@
...
+  const globals = globalThis as unknown as Record<symbol, unknown>;
+  const existing = globals[COPIES];
+  if (existing instanceof Set) return existing as Set<object>;
+  const created = new Set<object>();
+  globals[COPIES] = created;
+  return created;
+})();
+
+// Evaluation happens once per copy, so counting these counts copies.
+registry.add({});
```
dunxonu: The copy-counting registry increments once per module evaluation with no way to decrement, so a hot-reloaded `@dunx/core` module in a long-lived process can push the count above one without a real second copy being installed.

**How it fails:** If `@dunx/core` itself is swapped in place by `bun --hot` while a process keeps running (module top-level re-executes, `globalThis` persists), `registry.add({})` runs again and `loadedCoreCopies()` returns 2+, causing `assertOneCore()` to throw 'copies of @dunx/core are loaded' even though only one real install is present. Narrow (would need `--hot` on the framework's own source rather than a consuming app's `node_modules`), but the design has no reset path.

dunxonu (commented):

dunxonu (approved):

The changes successfully address diagnostics for missing dependency records, guard against multiple loaded copies of the core package, publish the package build utility, and add a publishing guide.

packages/transform/src/build.ts
```diff
@@ -0,0 +1,216 @@
...
+  const binOutputs = new Set<string>();
+  const exported = new Set<string>();
+
+  for (const entry of Object.values(pkg.exports ?? {})) {
+    const target = typeof entry === 'string' ? entry : entry.import;
+    if (target) {
+      entrypoints.add(await toSource(cwd, target));
+      exported.add(unprefixed(target));
+    }
+  }
```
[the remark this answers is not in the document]
petarzarkov: Fixed in 90736ed. `exportTargets` walks the tree and collects every file it names, skipping the `.d.ts` of a `types` condition, so the root string spelling, a `null` subpath, a nested condition and an array of fallbacks all resolve. Two tests cover those shapes.

One deliberate difference from the proposed fix: it builds every target rather than the first condition that matches. A package declaring both `bun` and `import` needs both built, and picking one leaves the other pointing at a file nobody emitted.

packages/core/src/di/transform-hint.ts
petarzarkov: Correct, and it was the worse half of the bug I was fixing: the old message sent an app author to check a preload they already had, and the new one would have sent them to file an issue against a package that is not involved. Fixed in b98731b. The branch names both causes now, the class expression first, with the reason the transform cannot record one.

packages/transform/src/build.ts
petarzarkov: Fixed in b98731b, and there were three copies rather than two: `internal/docs/scripts/extract/index.ts` declared the same shape inline. The build owns `ExportEntry` now and exports it alongside `exportTargets`; the version tooling and the docs extractor both read those.

Worth noting the drift had already happened in the direction you describe. All three spellings were `string | { import?: string }`, which is narrower than the field, so the docs extractor silently skipped a nested condition. It goes through `exportTargets` now.

packages/core/src/di/copies.ts
petarzarkov: Fixed in b98731b. The registry keys on `import.meta.url` rather than a fresh object per evaluation, so a re-run of the same module re-adds the key it already had and the count stays at one. Two installs sit at two paths and still count two. That removes the need for a reset path rather than adding one.

Checked that the query suffix the subprocess fixture uses to fake a second copy survives into `import.meta.url`, so that test still measures what it did.

dunxonu (approved):

This PR introduces a robust package build utility, safeguards against duplicate core package loading, and improves transform diagnostics for third-party packages.

packages/core/src/di/transform-hint.ts
dunxonu: Fixed in b98731b — the branch now checks for class expressions first and explains the transform limitation, so app authors aren't misdirected to blame a third-party package.

petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

dunxonu (commented):

internal/docs/scripts/extract/index.ts
```diff
@@ -92,7 +96,7 @@ export const extractPackage = (options: ExtractOptions): PackageDoc => {
   const exposure = new Map<string, Set<string>>();
 
   for (const [subpath, target] of Object.entries(manifest.exports ?? {})) {
-    const distPath = typeof target === 'string' ? target : target.import;
+    const distPath = exportTargets(target)[0];
```
dunxonu: Docs extraction now picks the first flattened export target (`exportTargets(target)[0]`) instead of specifically the `import` condition, so the chosen source file depends on key order in the `exports` map rather than semantics.

**How it fails:** Every current package.json in the repo happens to write `{ types, import }` in that order so `types` (filtered out as `.d.ts`) always precedes and `import` is picked. If a future subpath adds another non-types condition (e.g. `"node"` or `"default"` alongside `"import"`), or writes conditions in a different order, `exportTargets(target)[0]` silently picks whichever key happens to come first in `Object.values()`, pointing the generated guide at the wrong (or a non-existent) source module instead of the intended `import` target.

scripts/workspace-ranges.ts
```diff
@@ -40,18 +42,14 @@ export const DEPENDENCY_FIELDS = [
...
-
 /** A parsed `package.json`. Named fields are optional; the rest is `unknown`. */
 export interface Manifest {
   name?: string;
   version?: string;
   description?: string;
   type?: string;
   private?: boolean;
-  exports?: Record<string, string | ExportEntry>;
+  exports?: Record<string, ExportEntry>;
```
dunxonu: `Manifest.exports` is typed as `Record<string, ExportEntry>`, but the sibling `PackageManifest` in `packages/transform/src/build.ts` (line 19) types the same package.json field as a bare `ExportEntry`, which itself already allows a top-level string/null/array - the exact shape this PR added support for and tested (`build.test.ts`: "reads a bare string exports field").

**How it fails:** A package.json with a bare top-level `"exports": "./dist/index.js"` (valid per Node/Bun and explicitly supported by `buildPackage`) does not satisfy `workspace-ranges.ts`'s `Manifest` type; any script that types-casts a real manifest through this `Manifest` interface and then treats `exports` as a `Record` (e.g. `Object.keys(manifest.exports)`) would silently misbehave for such a package, since a string's `Object.keys` returns numeric character indices rather than subpath keys.

packages/transform/src/build.ts
```diff
@@ -0,0 +1,232 @@
...
+      `${name}: package.json must set "type": "module" - dunx is ESM only.`,
+    );
+  }
+
+  const entrypoints = new Set<string>();
+  const binOutputs = new Set<string>();
+  const exported = new Set<string>();
+
+  for (const target of exportTargets(pkg.exports ?? {})) {
+    entrypoints.add(await toSource(cwd, target));
```
dunxonu, line 126: `buildPackage` resolves each export/bin target's source file with a sequential `await toSource(...)` inside a `for` loop instead of resolving them concurrently, even though each check is an independent file-existence lookup.

**How it fails:** A package with many subpaths (e.g. a barrel package with a dozen `exports` entries) pays the latency of a dozen sequential `Bun.file(...).exists()` round trips during every build instead of one batched `Promise.all`, adding avoidable wall-clock time to `bun run build` for larger packages.

internal/docs/scripts/extract/index.ts
petarzarkov: Right, and it was a regression I introduced: the old line read `.import` explicitly, and flattening turned that into whichever key came first. Fixed in 13b22e0.

The underlying mistake was answering one question with the other's function. A build needs every file the map names; a reader needs the one file an `import` resolves to. `importTarget` is the second, preferring `import`, then `default`, then `bun`, then `node`, and the extractor uses it.

scripts/workspace-ranges.ts
petarzarkov: Fixed in 13b22e0 by deleting the field rather than widening it. Nothing has read `Manifest.exports` since the build moved into the package: `scripts/setup-doc.test.ts` declares its own inline type and the docs extractor has its own. Keeping it would have meant a third spelling of the same field for no reader.

The docs extractor's copy is widened to what the field can hold, with the subpath map guarded out of it.

packages/transform/src/build.ts
petarzarkov: Leaving this one sequential, deliberately.

The saving is real but small: each `toSource` is one `Bun.file().exists()`, and the largest package here has 17 entrypoints against a 535ms build dominated by `Bun.build` and `tsc`. What the loop buys is a deterministic error. Resolved in declaration order, a manifest with two bad targets always names the first one; under `Promise.all` the rejection that wins is whichever filesystem call loses, so the same broken manifest reports a different path run to run.

For a build tool that trade reads the wrong way round. Happy to revisit if someone turns up with an exports map long enough for it to show.

dunxonu (commented):

packages/transform/src/build.ts
```diff
@@ -0,0 +1,232 @@
...
+const unprefixed = (distPath: string): string => distPath.replace(/^\.\//, '');
+
+/**
+ * Every file an `exports` tree names, at any depth. Both targets of a package
+ * declaring two conditions are built: picking one leaves the other pointing at
+ * a file nobody emitted. A `.d.ts` leaf is a `types` condition, which `tsc`
+ * emits rather than an entrypoint.
+ */
+export const exportTargets = (entry: ExportEntry): readonly string[] => {
+  if (typeof entry === 'string') return entry.endsWith('.d.ts') ? [] : [entry];
```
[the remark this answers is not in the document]
petarzarkov: Fixed in 13b22e0. `isDeclaration` matches `/\.d\.[cm]?ts$/`, so all three spellings are skipped, and a test declares an entry carrying each of them alongside the `import` target.

internal/docs/scripts/extract/index.ts
```diff
@@ -13,7 +17,7 @@ import {
 export interface Manifest {
   readonly name: string;
   readonly description?: string;
-  readonly exports?: Record<string, string | { import?: string }>;
+  readonly exports?: Record<string, ExportEntry>;
```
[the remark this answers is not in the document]
petarzarkov: Fixed in 13b22e0. The extractor's `Manifest.exports` is widened to the full `ExportEntry`, and the root string, array and condition-map forms are guarded out before `Object.entries`, so a package written that way contributes no subpaths instead of enumerating character indices. The per-subpath target now goes through `importTarget`.

internal/docs/scripts/extract/index.ts
dunxonu: Fixed in 13b22e0 — thanks for the quick turnaround.

dunxonu (commented):

This change introduces a build utility for Dunx packages, improves diagnostics for missing dependency records, and adds a safeguard against multiple core package copies. It also updates documentation and refactors the internal build script.

PR #162, opened 2026-09-18
feat(infra/db): named data sources, a pool keyed at runtime, and the Search Console indexing fixes

### LINKED ISSUE #160: DB Module: Support multiple named data sources
## Background

The current `@dunx/infra/db` module primarily targets single-datasource applications.

In production applications, it is common to use multiple databases for different purposes, such as:

* Primary business database
* Reporting/analytics database
* Audit database
* Read replicas
* Separate databases for different tenants

Multi-tenant SaaS applications may also need to select a database dynamically at runtime based on the current tenant or request context.

It would be useful for the DB module to support both **multiple data sources** and **dynamic data source selection**, while preserving the existing Drizzle-native API.

### 1. Multiple Data Sources

Allow applications to register multiple independent data sources, each with its own Drizzle instance.

For example:

```ts
DbModule.forRoot([
  {
    name: 'primary',
    options: new PostgreSqlOptions({
      // ...
    }),
  },
  {
    name: 'report',
    options: new PostgreSqlOptions({
      // ...
    }),
  },
])
```

Applications should be able to explicitly retrieve a specific data source:

```ts
const db = dataSources.get('report')

await db
  .select()
  .from(users)
```

Single-datasource applications should continue to work without additional configuration.

### 2. Dynamic Data Sources

Support resolving a data source at runtime.

A common use case is database-per-tenant:

```text
tenant-a → database-a
tenant-b → database-b
tenant-c → database-c
```

For example:

```ts
const db = await dataSources.get(tenantId)
```

or a resolver-based approach:

```ts
const db = await dataSources.resolve(context)
```

The exact API can follow Dunx's existing DI and request-context design.

### 3. Data Source Lifecycle Management

Dynamic data sources should not create a new connection pool for every request.

The DB module should be able to manage the lifecycle of dynamically created data sources, including:

* Creating and reusing data sources
* Connection pool management
* Closing data sources
* Idle data source cleanup
* Creation failure handling
* Caching

For applications with a large number of tenants, some form of maximum datasource count and idle eviction may also be necessary to avoid exhausting database connections.

### 4. Preserve the Drizzle-Native API

The feature should remain a thin integration on top of Drizzle rather than introducing another query or repository abstraction.

For example:

```ts
const db = dataSources.get('primary')

await db
  .select()
  .from(users)
```

This should preserve Drizzle's existing type inference, query API, and transaction support.

## Concurrency Considerations

Dynamic datasource selection should not rely on mutating a global `currentDataSource`.

For example, these requests may execute concurrently:

```text
Request A → tenant-a
Request B → tenant-b
```

Changing a global datasource could cause one request to accidentally use another tenant's database.

If automatic datasource selection based on request context is introduced, it should be isolated to the current execution/request context.

## Suggested Scope

The initial implementation could focus on:

* Multiple datasource registration
* Named data sources
* Explicit datasource retrieval
* Datasource lifecycle management
* Dynamic datasource creation/resolution

More advanced features can be considered later:

* Request-context datasource binding
* Tenant-to-datasource resolution
* Datasource caching/LRU
* Datasource metrics
* Read/write splitting

The goal is to provide a production-ready multi-datasource foundation while keeping `@dunx/infra/db` lightweight and consistent with Dunx's existing Drizzle integration.

### LINKED ISSUE #161: DB Module: Dynamic per-tenant data sources and their lifecycle
Split out of #160, which proposed multiple and dynamic data sources together. That issue now covers named static data sources only. This one is the dynamic half. Original proposal and use cases by @yibird.

## Problem

A database-per-tenant application resolves its data source at runtime rather than at registration:

```text
tenant-a -> database-a
tenant-b -> database-b
tenant-c -> database-c
```

Named registration does not cover this. The set of tenants is not known when the module is configured, and for a large tenant count a binding per tenant is not workable anyway.

## Why this is its own issue

Named data sources (#160) are a token change: bind under `dbConnection(name)` instead of the drizzle class, the way `redisConnection(name)` and `httpClient(name)` already do. Nothing new owns state.

This is a lifecycle subsystem. Something has to create connection pools on demand, keep them, and eventually close them, and getting that wrong exhausts the database's connection limit in production rather than failing a test. It needs its own design pass.

## Scope

- Resolve a data source at runtime from a key, creating it on first use
- Reuse the pool for a key rather than connecting per request
- Bound the number of live data sources, with eviction when the bound is hit
- Close idle data sources after some period
- Handle creation failure without caching the failure forever
- Shut every live data source down on application shutdown, in the existing reverse-construction order

## Concurrency

Resolution must not go through a mutable global. These may be in flight at once:

```text
Request A -> tenant-a
Request B -> tenant-b
```

A global `currentDataSource` would let one request read another tenant's database, which for a multi-tenant application is a data leak rather than a bug.

If resolution is bound to the request context, it should use the `AsyncLocalStorage` already in `packages/core/src/logger/context.ts` rather than a new mechanism.

## Open questions

- What is the key. A string the caller passes, or something read from request context
- Where eviction policy is configured, and what the default bound is
- Whether a repository injects a resolver and calls it, or injects something request-scoped that is already resolved
- What happens to an in-flight query when its data source is evicted
- Whether `QueryMetrics` is per data source or shared

## Not in scope

Read/write splitting and replica routing. Different problem, and it should wait until this one is settled.

petarzarkov:
Closes #160. Closes #161.

## Named data sources (#160)

`DbModule.forRoot(options, { name })` binds `dbOptions(name)`, `dbConnection(name)`, `dbHandle(name)` and `dbMetrics(name)` instead of `DbOptions`, `DbConnection`, drizzle's class and `QueryMetrics`, so several static data sources coexist alongside one default. Same shape `redisConnection(name)` and `httpClient(name)` already use. `dbHandle` is generic, so the annotation keeps drizzle's schema inference where a registry returning a union across schemas would lose it. `forRootAsync` takes the same `name`.

## A pool keyed at runtime (#161)

`DbModule.forDataSources({ create }, As)` binds a `DataSources` pool: opens on first use, bounds live data sources with LRU eviction, closes idle ones on a sweep, drops a failed open rather than caching it, and closes everything on shutdown in the existing reverse-construction order.

Answers to the open questions on the issue:

- **The key** is an argument at every call, so nothing ambient decides which database a query runs against. Reading it from `RequestContext` is one line at the call site.
- **Eviction policy** is `max`, `idleMs` and `sweepMs` on the init. Defaults 32 and 5 minutes.
- **A repository injects the resolver** and calls it. No request-scoped provider.
- **An in-flight query** is covered by a lease: `use(key, work)` counts itself and eviction skips anything counted. With the bound reached and every data source borrowed, resolution fails naming the bound rather than opening one more connection.
- **`QueryMetrics` is shared** for the pool. Per tenant is unbounded by construction.

Recency is the map's own insertion order rather than `Date.now()`, which has millisecond resolution and made the first version evict the data source it had just touched.

`examples/full` grows a `tenants` module using both, and the tour narrates it. Kept out of `create-app`'s scaffolded `database` feature.

## The Search Console indexing report

Separate commit, unrelated to the above, requested alongside it.

- **Not found (404), 2 rows:** two dead internal links on the landing page. `/guide/architecture` was never a slug, and `/api/compiler` should be `/api/transform`. Both are written in JSX, which `links.test.tsx` never walks. A new check over the emitted HTML covers every internal href, and a second fails on a published `/releases/v3.x` link, which only 301s.
- `404.html` is a real asset, so `/404` answered 200 and self-canonicalised. Now `noindex, follow` with no canonical.
- The sitemap stamped the build date on all 117 URLs. Only release notes carry a date now; the other 46 carry none.
- A preview deploy serves all 117 pages again under `*.pages.dev` with absolute canonicals pointing at dunx.win. `DOCS_NOINDEX` writes `Disallow: /` into that build only.

**Alternate page with proper canonical tag (117 rows) is not addressed here.** Every page's canonical is self-referential and correct, and 117 is the exact page count, so a second host or path spelling is serving the whole site. The likeliest is `www.dunx.win` alongside the apex, which is a Cloudflare redirect rule rather than anything in this repo. The URL examples in Search Console under that reason would confirm it.

**Page with redirect (3 rows) is not addressed here** either. No published link uses a redirecting path (the new test now enforces that), so these are external or historical `/releases/v3.x` URLs, and the 301 is already the correct answer.

## Gates

`bun run ci` is green except `examples/tests`, which fails only on `exits 0 with no redis at all` - it fails the same way on `main` on this machine, where a local valkey is running so the fixture never sees the degraded path.

## Summary by CodeRabbit

* **New Features**
  * Added support for named database connections and configurable per-tenant data-source pools.
  * Added tenant database examples with ticket management, reporting rollups, capacity limits, idle cleanup, and resource release.
  * Added database connection metrics and public configuration options for data-source pools.

* **Documentation**
  * Expanded database documentation with multi-source and tenant-pool guidance.
  * Improved sitemap dates, API links, and architecture navigation.

* **Bug Fixes**
  * Preview documentation builds are now excluded from search indexing.
  * 404 pages no longer expose canonical links and correctly signal non-indexing.


petarzarkov:
@coderabbitai review

petarzarkov:
@dunxonudeepreview

dunxonu (commented):

The change introduces named data sources and a runtime-keyed connection pool, but violates the repository file length limit.

- `packages/infra/src/db/data-sources.test.ts:511` Source file exceeds the 500-line limit

</details>

petarzarkov:
@coderabbit review this

petarzarkov:
@dunxonudeepreview

dunxonu (commented):
**Actionable comments posted: 6**

packages/infra/src/db/data-sources.ts
```diff
@@ -0,0 +1,274 @@
...
+        `These data sources closed while "${key}" was opening.`,
+      );
+    }
+    this.#logger?.debug(`opened data source "${key}"`);
+    return opened;
+  }
+
+  async #makeRoom(key: string): Promise<void> {
+    if (this.#max <= 0) return;
+    while (this.#entries.size >= this.#max) {
```
dunxonu, line 225: `#makeRoom`'s while-loop re-checks total pool size after each await, so it can evict an entry a different concurrent caller just inserted for its own new key, not just the stale one it originally targeted.

**How it fails:** max: 1, one idle entry 'x' is live. Two concurrent calls Promise.all([sources.get('y'), sources.get('z')]) both miss the cache. The first call's #makeRoom evicts 'x' (delete happens synchronously before `await this.#shut`), then suspends. Before that close settles, the second call's #makeRoom sees size 0, skips eviction, and inserts 'z' (size back to 1, leases 0 since get()/connection() never lease). When the first eviction's await resolves, #makeRoom's while loop re-checks size, finds it back at max, and evicts 'z' via #idlest() - closing a connection the second caller just opened and may already be reading from, even though 'z' was never idle.

packages/infra/src/db/data-sources.ts
```diff
@@ -0,0 +1,274 @@
...
+    for (const [key, entry] of live) {
+      if (entry.leases > 0 || entry.lastUsed > cutoff) continue;
+      if (await this.evict(key)) closed += 1;
+    }
+    return closed;
+  }
+
+  /** Idempotent, and final: a `get` afterwards throws rather than reopening. */
+  async close(): Promise<void> {
+    this.#closed = true;
```
dunxonu, line 162: `close()`'s guarantee that a call afterward always throws rather than reopening can be bypassed by an in-flight `#entry()` call that already passed its `#closed` check before `close()` ran, and the resulting open is not awaited by `close()`'s own `Promise.all`.

**How it fails:** A request calls `sources.get(newKey)`, which suspends inside `await this.#makeRoom(key)` (e.g. waiting on a slow eviction's I/O). Concurrently, shutdown calls `close()`, which synchronously sets `#closed = true`, clears and snapshots `#entries`, and awaits only that snapshot. The suspended `#entry()` call resumes, finds the map now empty, and inserts a brand-new entry that calls `#open(key)` - opening a real connection - after `close()` has already resolved. The caller of `close()` (typically right before process exit) never waited for this straggling open, so the process can exit while a connection is mid-open, or the open+immediate-self-close in `#open` races the exit.

packages/infra/src/db/tokens.ts
```diff
@@ -0,0 +1,67 @@
...
+import type { DbConnection, DbOptions } from './connection.js';
+import type { QueryMetrics } from './metrics.js';
+
+/**
+ * One cache for all four families, keyed by the description, which already
+ * carries the family and the name. `token()` returns a fresh object per call, so
+ * without this the module and its consumer would hold different tokens for one
+ * name and the lookup would miss.
+ */
+const cache = new Map<string, Token<unknown>>();
```
dunxonu: `namedToken`'s memoized `Map<string, Token>` cache reimplements the exact per-name token memoization already declared in `packages/infra/src/redis/module.ts` (`tokens`/`metricsTokens` maps), violating Rule 2 ('one declaration, at the lowest common owner... if a second copy would exist, move the first somewhere both can reach and delete it') from the repo CLAUDE.md.

**How it fails:** Two independent implementations of 'memoize a Token by a string name because token() returns a fresh object per call' now exist side by side in the same package (packages/infra/src/redis/module.ts:17-18 and packages/infra/src/db/tokens.ts:11). A future fix to the memoization behavior (e.g. clearing stale entries, guarding against a name reused for a different type) has to be made twice and will likely be made in only one place.

packages/infra/src/db/module.ts
```diff
@@ -1,12 +1,158 @@
...
+  dbHandle,
+  dbMetrics,
+  dbOptions,
+  namedToken,
+} from './tokens.js';
+
+/** The init's own token, one per pool class, so two pools do not collide. */
+const dataSourcesInit = <TDb>(
+  as: Ctor<DataSources<TDb>>,
+): Token<DataSourcesInit<TDb>> => namedToken(`DataSourcesInit(${as.name})`);
```
dunxonu: `dataSourcesInit` (and `dbMetrics(target.name)` in `poolModule`) key their token cache by the JS class's `.name` string with no uniqueness guard, so two distinct `DataSources` subclasses that happen to share a class name silently collide.

**How it fails:** Two feature modules each declare `class Tenants extends DataSources<X> {}` (a plausible, generic name - `data-sources.test.ts` itself uses `class Tenants extends DataSources<Handle> {}`). `namedToken('DataSourcesInit(Tenants)')` returns the same cached Token object for both, and if both modules also turn on `{ metrics: true }` and are ever imported into a shared parent scope, the parent sees the same exported metrics token bound twice - a confusing 'duplicate binding' boot error with no indication the real cause is an incidental class-name collision rather than an intentional shared registration.

packages/infra/src/db/module.ts
```diff
@@ -20,12 +166,20 @@ import { QueryMetrics } from './metrics.js';
  *
  * `{ metrics: true }` binds a fourth, {@link QueryMetrics}, and times every query
  * through it. Off by default, so an app that reads no numbers wraps no driver.
+ *
+ * `{ name }` moves all of them onto per-name tokens, and
+ * {@link DbModule.forDataSources} registers data sources whose keys are only
+ * known at runtime.
  */
 export class DbModule {
```
dunxonu, line 209: `forRoot`/`forRootAsync` re-derive the same exports/providers wiring that `namedModule` already generalizes for both sync and async options, and the `metricsToken === undefined ? ... : ...` branch is copy-pasted four times (namedModule, poolModule, forRoot, forRootAsync).

**How it fails:** A change to how the metrics-wrapped connection factory is built (e.g. adding error handling around `instrumented()`) needs to be applied at module.ts:76-88, :139-152, :199-210, and :262-276 independently; missing one of the four sites (as already nearly happened here, since `poolModule` was written as a fresh implementation rather than reusing `namedModule`'s pattern) leaves an inconsistent code path that only a targeted test would catch.

packages/infra/src/db/index.ts
```diff
@@ -40,6 +45,7 @@ export {
   SyncSqliteOptions,
   type SqliteInit,
 } from './sqlite/options.js';
+export { dbConnection, dbHandle, dbMetrics, dbOptions } from './tokens.js';
```
dunxonu, line 48: `dbOptions` and `dbMetrics` are added to the package's public surface but never used anywhere in `examples/full`, only `dbConnection`/`dbHandle` are exercised by the new tenants example.

**How it fails:** Per Rule 4 ('a feature is not shipped until an example uses it'), a consumer reading examples/full for how to use named registrations has no working reference for `dbOptions(name)` (reading resolved configuration) or `dbMetrics(name)` (reading query stats for a named registration) and could get either wrong on first use with no example to check against.

petarzarkov:
Both reviews acted on. Every finding below was re-traced against the code first; two turned out to be the same defect, and one is declined with a reason.

## The concurrency defect (both reviewers)

Same root cause, two manifestations: `#makeRoom` awaited the eviction it had asked for, so between reading `size` and inserting, a second caller admitted itself into the same slot. CodeRabbit's reading is the pool exceeding `max`; dunxonu's is the pool evicting the entry the first caller had just inserted. Both follow from the one `await`.

Two failing tests first (`never holds more than max when two keys race for the last slot`, `waits for an open that was already in flight when it ran`), then the fix: `#entry` and `#makeRoom` are synchronous end to end, so admission cannot interleave. Eviction frees the slot immediately and closes in the background, tracked in `#closing` so `close()` waits for those too.

I dropped a third test I had written for dunxonu's variant in isolation: it ends in the same state a correct run does, so it asserted nothing. The size-bound test covers the window.

**`close()` straggler** - fixed by the same change. With admission atomic, an insert is either in `close()`'s snapshot or refused outright, so `close()` no longer resolves ahead of an open it started.

**`#shut` unbounded wait on `entry.opening`** - real, and separate. `closeTimeoutMs` bounded only the close, so a connect that never settled blocked `evict`, `prune`, `close` and `onShutdown`. Now bounded, with a connection that surfaces after the deadline closed rather than orphaned. The race-with-a-timer is `within` in `close-within.ts`, which `closeWithin` now uses rather than declaring its own.

## Rule 2: `namedToken`

Correct, and it was worse than reported: three copies, not two. `packages/infra/src/redis/module.ts` had two maps, `packages/http/src/client/module.ts` a third. It is `namedToken` in `@dunx/core` beside `token()` now, and all three copies are deleted.

## Token identity by class name

Fixed both ways. The init token is keyed by the constructor (a `WeakMap`), since nothing exports it for a consumer to look up by string. The metrics token stays string-addressed because `dbMetrics(name)` is how a consumer reads it, so `DataSourcesSettings.name` sets it, defaulting to the class name. New test: two classes that genuinely share the runtime name `Tenants`, bound in one root, staying apart.

## Four copies of the metrics branch

Collapsed. `forRoot`, `forRootAsync` and `namedModule` go through one `connectionModule`, which takes the four tokens and the options provider. `poolModule` keeps its own, since it constructs a pool rather than a connection.

## Rule 4: `dbOptions` and `dbMetrics` unexercised

Correct, and `reportingConnection` was a declaration nothing read. `examples/full` now uses all four for the reporting data source, through a `GET /tenants/reporting` route and the tour.

## `TENANT_DATABASE_FILE` without `{tenant}`

Correct and worth catching: the replace is a no-op, every tenant resolves to one file, and the isolation the example exists to show disappears with nothing failing. Rejected at boot now, with `:memory:` still allowed since each open is already its own database.

## The guide's two-data-source snippet

Correct. oxfmt had folded my two statements into a parenthesized comma expression, which registers only the second. It is an `imports` array now.

## Built-link check and dotted paths

Correct as a latent bug. `/releases/3.3.2` has a dot in its last segment, so the extension test classified a page as an asset. It passed only because the releases list is client-rendered and those links are not in the emitted HTML. Now the asset spelling wins only if that file exists, else it falls back to `fileFor`.

## Declined

**Out of scope changes** - the documentation-site indexing fixes are unrelated to #160 and #161, and CodeRabbit is right that they are. They were requested in this branch alongside the feature; they are a separate commit (`0d1ab76`) and can be split out if you would rather they were.

`bun run ci` is green except `examples/tests`, which fails only on `exits 0 with no redis at all` - it fails identically on `main` on this machine, where a local valkey is running so the fixture never reaches the degraded path.

dunxonu (approved):

Adds named database token factories and runtime tenant DataSources pooling with eviction, alongside Search Console fixes.

